PR #14, opened 2013-06-21
Changing dtests to use a custom logging - ability to specify custom class names -> log levels in log4j

azarutin:


ptnapoleon:
Closing this as old. If we feel we need this, can always re-open and rebase the original patch

PR #92, opened 2014-09-19
Add test for commitlog archiving

beobal:
see CASSANDRA-6904


ptnapoleon:
@beobal Is this good to merge?

beobal:
@ptnapoleon I would say so. Tests pass with it (against trunk at least), plus it actually exercises the functionality from 6904

PR #134, opened 2014-12-08
Add tests for TTL feature

aboudreault:
Add tests for TTL feature.

Unfortunately, the ttl tests take around 4.5 minutes to run. I tried to use smaller interval in my tests, but this produced unexpected failure due to a too fast or too slow execution.

### DIFF
--- a/ttl_test.py
+++ b/ttl_test.py
@@ -1,9 +1,17 @@
 import time
 from collections import OrderedDict
 from cassandra.util import sortedset
+from cassandra.query import SimpleStatement
+from cassandra import ConsistencyLevel
 from dtest import Tester
 from pytools import since
-from pyassertions import assert_all, assert_row_count
+from pyassertions import (
+    assert_all,
+    assert_none,
+    assert_row_count,
+    assert_almost_equal,
+    assert_unavailable
+)
 
 
 class TestTTL(Tester):
@@ -11,10 +19,10 @@ class TestTTL(Tester):
 
     def setUp(self):
         super(TestTTL, self).setUp()
-        self.cluster.populate(3).start()
-        [node1, node2, node3] = self.cluster.nodelist()
+        self.cluster.populate(1).start()
+        [node1] = self.cluster.nodelist()
         self.cursor1 = self.patient_cql_connection(node1)
-        self.create_ks(self.cursor1, 'ks', 3)
+        self.create_ks(self.cursor1, 'ks', 1)
 
     def prepare(self, default_time_to_live=None):
         self.cursor1.execute("DROP TABLE IF EXISTS ttl_table;")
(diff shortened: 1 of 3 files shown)


ttl_test.py
```diff
@@ -0,0 +1,473 @@
...
+            [[1, 1, None, None]],
+            cl=ConsistencyLevel.ALL
+        )
+        ttl_cursor1 = self.cursor1.execute('SELECT ttl(col1) FROM ttl_table;')
+        ttl_cursor2 = cursor2.execute('SELECT ttl(col1) FROM ttl_table;')
+        assert_almost_equal(ttl_cursor1[0][0], ttl_cursor2[0][0], error=0.05)
+
+        time.sleep(3.5)
+
+        assert_all(
```
ptnapoleon: Instead of calling assert_all to validate that nothing is returned, just use assert_none

ttl_test.py
```diff
@@ -0,0 +1,473 @@
...
+        # Check that the expired data has not been replicated
+        assert_row_count(cursor2, 'ttl_table', 1)
+        assert_all(
+            cursor2,
+            "SELECT * FROM ttl_table;",
+            [[2, 2, None, None]],
+            cl=ConsistencyLevel.ALL
+        )
+
+        # Check that the TTL on both server are the same
```
ptnapoleon: You need exclusive_cql_connection here. The driver is not only connecting to the node you pass into patient_cql_connection, but using it as a contact point. If you want to restrict queries to one node, exclusive_patient_cql_connection should be used.

ttl_test.py
```diff
@@ -0,0 +1,473 @@
...
+        assert_all(
+            self.cursor1,
+            "SELECT * FROM ttl_table;",
+            [[1, 1, None, None], [2, 2, None, None]]
+        )
+        time.sleep(3.5)
+        self.node1.stop()
+        cursor2 = self.patient_cql_connection(self.node2)
+        cursor2.execute("USE ks;")
+        assert_unavailable(cursor2.execute, "SELECT * FROM ttl_table;")
```
ptnapoleon: Why would this be unavailable? I must be missing something. You are only using CL ONE.

aboudreault: node1 is stopped. node2 doesn't have the data yet.

ptnapoleon: Why wouldn't node2 have the data yet? It was up when it was inserted. Unless you hashed the partition keys and can guarantee they both belong to node1.

ptnapoleon: Additionally, even if it did have an empty table, which it may, at CL.ONE it should still have the table, so it should still be able to respond to the query, just with an empty set.

aboudreault: ah, right. So, the data are only in node1 because there is only 2 row.... but that could change at test run. I will simply stop the node2 before inserting to be sure.

ptnapoleon: Well it would be deterministic which node those rows belong to (ignoring ccm changes to default tokens), but what it really seems you're checking for is that the data isn't there, not that the query returns unavailable.

ttl_test.py
ptnapoleon: If I explained this poorly, which I did, ping me privately and hopefully I can explain better.

ttl_test.py
ptnapoleon: Okay, I was wrong. This SHOULD return unavailable, but not because node2 lacks the data. There is a range for which all replicas are down (the range node1 owns in an RF=1 situation), thus a `select *` query will return Unavailable.

aboudreault:
Fixed comments and Rebased.

ptnapoleon:
+1!

PR #76, opened 2014-08-03
Rbac - Auth tests for role based access control

mike-tr-adamson:
This PR should be merged when https://issues.apache.org/jira/browse/CASSANDRA-7653 is resolved

PR #149, opened 2015-01-14
CASSANDRA-7653

ptnapoleon:
@beobal Are these ready to merge? Should I close #76 as out of date?


beobal:
yes, those are good to go and supercede #76. I also have the bones of an upgrade test (https://github.com/beobal/cassandra-dtest/commit/88f1dada81c340cbc8318ce145d93951c5a0c3c6). It doesn't pass atm, but I believe that's down to a driver issue.

PR #159, opened 2015-01-27
UDF, inc repair and 8613 coverage addtions

shawnkumar:


ptnapoleon:
sstable_repairedset_test is failing against 2.1. 
aggregate_udf_test is failing against trunk.

cql_tests.py
```diff
@@ -449,6 +449,25 @@ def tuple_query_mixed_order_columns_test9(self):
         assert rows_to_list(res) == [[0, 0, 0, 0, 0], [0, 0, 1, 1, 0], [0, 0, 1, 1, -1],
                                      [0, 0, 1, 0, 2], [0, -1, 2, 2, 2]], res
 
+    def simple_tuple_query_prepare(self, cursor):
```
ptnapoleon: Why create a separate prepare function if it's only being called in one test?

repair_compaction_test.py
```diff
@@ -33,60 +71,141 @@ def repair_compaction_fine_test(self):
...
+        self.assertNotIn('repairedAt: 0', finaloutput)
+
+        os.remove('initial.txt')
+        os.remove('final.txt')
+    
+    @since('2.1')
+    def compaction_test(self):
+        cluster = self.cluster
+        cluster.populate(3).start()
+        [node1,node2,node3] = cluster.nodelist()
```
ptnapoleon: Better to use `node1, node2, node3 = cluster.nodelist()`

ptnapoleon:
Have you considered renaming repair_compaction_test?

ptnapoleon:
+1

PR #160, opened 2015-01-28
Add some tests for to commitlogs

aboudreault:
Work In Progress. Do not merge, this is only partial work, more will be added.

Current tests:

* Test default commitlog segment size
* Test default compressed commitlog segment size
* Test small commitlog segment size 
* Test small compressed commitlog segment size
* Test IGNORE commit_failure_policy 
* Test DIE commit_failure_policy 
* Test STOP commit_failure_policy 
* Test STOP_COMMIT commit_failure_policy 

We only test some of failure policies in Cassandra-2.0 due to the lack of memtable options, which produce inpredictable flushing of the commitlogs.

### DIFF
--- a/assertions.py
+++ b/assertions.py
@@ -53,7 +53,7 @@ def assert_almost_equal(*args, **kwargs):
 
     vmax = max(args)
     vmin = min(args)
-    assert vmin > vmax * (1.0 - error), "values not within %.2f%% of the max: %s" % (error * 100, args)
+    assert vmin > vmax * (1.0 - error) or vmin == vmax, "values not within %.2f%% of the max: %s" % (error * 100, args)
 
 def assert_row_count(cursor, table_name, expected):
     """ Function to validate the row count expected in table_name """
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -1,0 +1,261 @@
+import os
+import stat
+import glob
+import time
+import platform
+import subprocess
+import ccmlib
+from tools import since
+from cassandra.cluster import NoHostAvailable, OperationTimedOut
+from dtest import Tester, debug
+from assertions import assert_one, assert_none, assert_almost_equal
+
+
+class TestCommitLog(Tester):
+    """ CommitLog Tests """
+
+    def __init__(self, *argv, **kwargs):
+        super(TestCommitLog, self).__init__(*argv, **kwargs)
+        self.allow_log_errors = True
+
+    def setUp(self):
+        super(TestCommitLog, self).setUp()
+        self.cluster.populate(1)
+        [self.node1] = self.cluster.nodelist()
+
+    def tearDown(self):
+        self._change_commitlog_perms(stat.S_IWRITE | stat.S_IREAD | stat.S_IEXEC)
+        super(TestCommitLog, self).tearDown()
+
+    def prepare(self, configuration={}, create_test_keyspace=True, **kwargs):
+        conf = {}
+        if self.cluster.version() >= "2.1":
+            conf['memtable_heap_space_in_mb'] = 512
+        conf.update(configuration)
+        self.cluster.set_configuration_options(values=conf, **kwargs)
+        self.cluster.start()
+        self.cursor1 = self.patient_cql_connection(self.node1)
+        if create_test_keyspace:
+            self.cursor1.execute("DROP KEYSPACE IF EXISTS ks;")
+            self.create_ks(self.cursor1, 'ks', 1)
+            self.cursor1.execute("DROP TABLE IF EXISTS test;")
+            query = """
+              CREATE TABLE test (
+                key int primary key,
+                col1 int
+              )
+            """
+            self.cursor1.execute(query)
+
+    def _change_commitlog_perms(self, mod):
+        path = self._get_commitlog_path()
+        os.chmod(path, mod)
+        commitlogs = glob.glob(path+'/*')
+        for commitlog in commitlogs:
+            os.chmod(commitlog, mod)
+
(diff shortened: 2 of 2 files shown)


commitlog_test.py
```diff
@@ -0,0 +1,76 @@
+import os
+import subprocess
+from dtest import Tester
+from assertions import assert_almost_equal
+
+
+class CommitLog(Tester):
```
ptnapoleon: Can you rename this TestCommitLog, to match the style of the other test cases?

aboudreault: oh, absolutely... that's a mistake.

commitlog_test.py
```diff
@@ -0,0 +1,76 @@
...
+        """ Returns the number of commitlog files in the directory """
+
+        path = self._get_commitlog_path()
+        return [os.path.join(path, p) for p in os.listdir(path)]
+
+    def _get_commitlog_size(self):
+        """ Returns the commitlog directory size in MB """
+
+        path = self._get_commitlog_path()
+        cmd_args = ['du', '-B', '1', path]
```
ptnapoleon: Is there an OS generic way to do this? As is, this won't work on WIndows.

aboudreault: Are we currently supporting running the dtests on Windows? Or it's for its future support? I haven't found the equivalent in python to get the *real* disk usage of a file, and not its apparent-size. I will take another look...

EnigmaCurry:
Does shutil.disk_usage return different results in some cases?

EnigmaCurry:
OK, I get what you're saying now, you do in fact need to calculate the size of each file and not rely on the dir stats - maybe http://stackoverflow.com/a/12480543/56560

aboudreault:
getsize is the function that was not working. It returns the *apparent-size*  (man du) of the file and not the disk usage of the file.

aboudreault:
psutil.disk_usage is really about the partition where a file is, not the file itself.

commitlog_test.py
ptnapoleon: We currently support dtest on windows.

assertions.py
```diff
@@ -31,7 +31,7 @@ def assert_one(cursor, query, expected, cl=ConsistencyLevel.ONE):
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = cursor.execute(simple_query)
     list_res = rows_to_list(res)
-    assert list_res == [expected], "Expected %s from %s, but got %s" % (expected, query, list_res)
+    assert list_res == [expected], "Expected %s from %s, but got %s" % ([expected], query, list_res)
```
ptnapoleon: Feel free to just make this change on master.

commitlog_test.py
```diff
@@ -0,0 +1,260 @@
...
+    def small_segment_size_test(self):
+        """ Test a small commitlog_segment_size_in_mb (5MB) """
+        segment_size_in_mb = 5
+        self.prepare(configuration={
+            'commitlog_segment_size_in_mb': segment_size_in_mb
+        }, create_test_keyspace=False)
+        self._commitlog_test(segment_size_in_mb, 60, 13, files_error=0.12)
+
+    @since('3.0')
+    def default_compressed_segment_size_test(self):
```
aboudreault: Will this be run on trunk when merged? If so, I guess I should add a require decorator.

commitlog_test.py
```diff
@@ -0,0 +1,260 @@
...
+        """ Test default compressed commitlog_segment_size_in_mb (32MB) """
+
+        self.prepare(configuration={
+            'commitlog_compression': [{'class_name': 'LZ4Compressor'}]
+        }, create_test_keyspace=False)
+        self._commitlog_test(32, 23, 2, compressed=True, files_error=0.5)
+
+    @since('3.0')
+    def small_compressed_segment_size_test(self):
+        """ Test a small compressed commitlog_segment_size_in_mb (5MB) """
```
aboudreault: same as the above comment..

aboudreault:
@ptnapoleon Can you take a final look on this? Thanks

assertions.py
```diff
@@ -31,7 +31,7 @@ def assert_one(cursor, query, expected, cl=ConsistencyLevel.ONE):
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = cursor.execute(simple_query)
     list_res = rows_to_list(res)
-    assert list_res == [expected], "Expected %s from %s, but got %s" % ([expected], query, list_res)
+    assert list_res == [expected], "Expected %s from %s, but got %s" % (expected, query, list_res)
```
ptnapoleon: You appear to be accidentally undoing e9f4f2ae5ab6585b24666862 here

commitlog_test.py
```diff
@@ -0,0 +1,260 @@
...
+                        num_commitlog_files, compressed=False,
+                        size_error=0.05, files_error=0):
+        """ Execute a basic commitlog test and validate the commitlog files """
+
+        if compressed:
+            segment_size_in_mb *= 0.7
+        segment_size = segment_size_in_mb * 1024 * 1024
+        self.node1.stress(['write', 'n=150000', '-rate', 'threads=25'])
+
+        if platform.system() != 'Windows':
```
ptnapoleon: There is a built in ccm function ccmlib.common.is_win() to use to check if it's windows

ptnapoleon:
Other than the 2 comments I just left, lgtm. +1

assertions.py
aboudreault: arf, good catch.

aboudreault:
Addressed comments and rebased. Will merge soon.

PR #162, opened 2015-01-28
Fix runtime and race condition.

ptnapoleon:
By flattening the lists to not include dicts, they become
hashable, meaning the comparisons are 5x faster, critical
with data sizes of 5k-50k rows. Race condition was due to
variable definition coming after callback.


ptnapoleon:
@knifewine to review.

knifewine:
+1

ptnapoleon:
This also fixes the problem on trunk, and I can confirm that flatten is not causing a loss of data.

PR #166, opened 2015-02-06
Improve upgrade tag discovery

knifewine:
give better info about tags/heads being used
allow finding 'tentative' tags
change trunk to 3.0
update version from strings to tuples for easier handling
query tags direct from apache git, or local if configured

### DIFF
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -1,5 +1,6 @@
 import bisect
 import os
+import pprint
 import random
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


aboudreault:
I am not familiar with that upgrade test yet, but this LGTM!

knifewine:
@ptnapoleon and/or @EnigmaCurry can you have a look at this? Changes don't alter the guts of the upgrade process used in tests, just make some alterations to how versions are specified and what tags are matched (so we can find 'tentative' tags).

upgrade_through_versions_test.py
```diff
@@ -115,7 +143,7 @@ def __init__(self, *args, **kwargs):
     @property
     def test_versions(self):
         # Murmur was not present until 1.2+
```
ptnapoleon: Do we still need this? I didn't think we were testing upgrades from 1.1.x anymore.

ptnapoleon:
+1.The code is verbose, but not too confusing. The list of improvements is quite nice.

EnigmaCurry:
TIL git ls-remote.

LGTM.

PR #167, opened 2015-02-06
quiet driver log noise on test cleanup

knifewine:
set log level to 'critical' to suppress (most) noise
there's perhaps a small risk of masking useful
information so I added an env var allowing
this behavior to be overridden by setting
SILENCE_DRIVER_ON_SHUTDOWN=false


knifewine:
@ptnapoleon mind giving this a look? Should lower the log noise on cassci/local testing. /cc @mshuler

mshuler:
lgtm

PR #169, opened 2015-02-17
Add sstablesplit test for CASSANDRA-8623

aboudreault:
Add sstablesplit test for CASSANDRA-8623

### DIFF
--- a/sstablesplit_test.py
+++ b/sstablesplit_test.py
@@ -1,7 +1,10 @@
 from dtest import Tester, debug
+from tools import since
 
 from os.path import getsize
 import time
+import subprocess
+import tempfile
 
 class TestSSTableSplit(Tester):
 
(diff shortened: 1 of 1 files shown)


sstablesplit_test.py
```diff
@@ -65,3 +68,30 @@ def _do_split(self, node, version):
...
+        Covers CASSANDRA-8623
+
+        Check that sstablesplit doesn't crash when splitting a single sstable at the time.
+        """
+        cluster = self.cluster
+        cluster.populate(1).start()
+        node = cluster.nodelist()[0]
+        version = cluster.version()
+
+        node.watch_log_for('thrift clients...')  # See comment in split_test about this
```
ptnapoleon: I found a better way to do this, and thought I had removed all instances of my bad hack. I'll patch split_test. Sorry about my bad example.

sstablesplit_test.py
```diff
@@ -65,3 +68,30 @@ def _do_split(self, node, version):
...
+        self._do_compaction(node)
+        with tempfile.TemporaryFile(mode='w+') as tmpfile:
+            node.run_sstablesplit(keyspace='keyspace1', size=2, no_snapshot=True,
+                                  stdout=tmpfile, stderr=subprocess.STDOUT)
+            tmpfile.seek(0)
+            output = tmpfile.read()
+
+        debug(output)
+        failure = output.find("java.lang.AssertionError: Data component is missing")
+        self.assertEqual(failure, -1, "Error during sstablesplit")
```
ptnapoleon: Won't the test fail if any errors show up in the log? Why search for a specific one?

sstablesplit_test.py
ptnapoleon: See 87ef6ee515a83f

sstablesplit_test.py
aboudreault: Note that the error I'm looking for is from the sstablesplit utility. It is not  in the node log. That's why I capture the sstablesplit stderr to check if things went wrong or not.

aboudreault: I guess I could also just find for "Error" in the log... but this error log I'm looking for is really related to the ticket issue. What do you prefer?

ptnapoleon: No, no you're right. I overlooked that you are searching the sstablesplit output, not in C*. Skimmed the code too hastily. Sorry.

ptnapoleon:
+1

aboudreault:
Great, thanks for the review!

PR #130, opened 2014-12-04
(WIP) Add test for CASSANDRA-8366: Did not get positive replies from all endpoints

aboudreault:
(Work in progress) Do not merge.

First version to test cassandra-8316. This PR requires https://github.com/pcmanus/ccm/pull/216

PR #171, opened 2015-02-20
Add test for CASSANDRA-8366

aboudreault:
Add test for CASSANDRA-8366

### DIFF
--- a/incremental_repair_test.py
+++ b/incremental_repair_test.py
@@ -6,7 +6,8 @@ from re import search, findall
 import unittest
 import time
 import os
-from assertions import assert_invalid, assert_one, assert_all, assert_none
+from assertions import assert_invalid, assert_one, assert_all, assert_none, assert_almost_equal
+from nose.plugins.attrib import attr
 
 class TestIncRepair(Tester):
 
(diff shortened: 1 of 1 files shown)


aboudreault:
@mshuler Since this test is long, can we change the default jenkins run command for: nosetests -a '!long' To not run long tests daily.

ptnapoleon:
Does this supercede #130?

aboudreault:
No, I will go back on 130 soon.... not sure if it is still worth adding it.

ptnapoleon:
I am +1 on the test. Your calculation for load size does duplicate the work that @mambocab did in #182, which he will be extracting and putting into ccm so you can just call node.data_size(). What you have is fine for now, as long as we make sure to change it once the CCM fix is in.

aboudreault:
+1, nice addition Jim. Merging this one. I have added an entry in my backlog to recheck 182 status and do the proper change in my code.

mshuler:
I couple test runs and the annotation looks fine to me - I'll add -a '!long' to the normal commit jobs. The time difference on just incremental_repair_test.py was 40 min vs 9 min w/o multiple_subsequent_repair_test

aboudreault:
Thanks @mshuler! It seems longer than locally though.. it was more 20-25 minutes during my development. Probably due to the machine load.

PR #175, opened 2015-02-26
makes since decorator work with classes

### LINKED ISSUE #152: Add ability to annotate entire class with since('version')

mambocab:
Changes the `@since` decorator to work with classes. It does so by decorating each test method on its argument if its argument is a class.

I've written a test for this decorator [here](https://github.com/mambocab/cassandra-dtest/blob/87d034e721deafbf0c2c803020480afa22961563/test_skip_class.py). Is there a place to put test rig tests?

closes #152

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -3,6 +3,7 @@ from decorator  import decorator
 from distutils.version import LooseVersion
 import re, os, sys, fileinput, time
 
+import nose
 from cassandra import ConsistencyLevel
 from cassandra.query import SimpleStatement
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Did you mean to merge this into master from the branch on riptano/cassandra-dtest?

ptnapoleon:
Is there no way to simply just skip the class, rather than decorating each method? We could also, if that is impossible, add a new decorator `sinceClass()` that is for classes only. Name TBD

mambocab:
> Did you mean to merge this into master from the branch on riptano/cassandra-dtest?

The source is correct, the destination isn't. I'll close and open a new one once we figure out implementation stuff (below).

> Is there no way to simply just skip the class, rather than decorating each method?

Should be doable [with `unittest.skip`](https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures). The limitations there:

- it'll break compatibility with pre-2.6. Do we care about that?
- it'll circumvent the [existing test-skipping facilities](https://github.com/riptano/cassandra-dtest/blob/master/dtest.py#L451) in `Tester`. Is there some reason to use that over `unittest.skip`?

ptnapoleon:
it'll break compatibility with pre-2.6. Do we care about that?
No

    it'll circumvent the existing test-skipping facilities in Tester. Is there some reason to use that over unittest.skip?
No

PR #176, opened 2015-02-27
@since decorator work with test classes

mambocab:
Addresses #152 -- this solves the same problem as #175, but the pull request is sent to the correct branch :fireworks:

The proposed solution in #175 worked by replacing every test function on the class with a wrapped version. This implementation works by replacing the class's `setUp` method, which is called before every test.

It'd be nice to just wrap the class itself, rather than one of its methods, but we can't find the C* version for a class until after `setUp` executes and the cluster gets initialized (so far as I can tell). I tried instantiating the class and getting the version inside `since`, but that messed up class creation somehow and removed the `runTest` method. It might be useful to add functionality to `Tester` that allows us to find the version without instantiating the test class.

### DIFF
--- a/auth_roles_test.py
+++ b/auth_roles_test.py
@@ -16,9 +16,9 @@ role2_role = ['role2', False, False]
 cassandra_role = ['cassandra', True, True]
 
 
+@since('3.0')
 class TestAuthRoles(Tester):
 
-    @since('3.0')
     def create_drop_role_test(self):
         self.prepare()
         cassandra = self.get_session(user='cassandra', password='cassandra')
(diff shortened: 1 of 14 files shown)


ptnapoleon:
Can you do the same for `@require`? And then take all of the classes that are decorated completely with since, and mark the entire class as since?

mambocab:
Yes. Would you like the `@require` changes in a separate PR?

ptnapoleon:
Yeah, probably best to put the changes to `@require` in a separate PR, but I see no compelling reason to isolate the implementation of the `since` change from this PR

mambocab:
I've moved the decorator to classes where possible. Ready for review.

ptnapoleon:
+1

PR #177, opened 2015-03-03
Additional tests

stef1927:
Additional tests for cassandra-7875 and cassandra 7816, one fix for test of cassandra-8516

### DIFF
--- a/prepared_statements_test.py
+++ b/prepared_statements_test.py
@@ -41,5 +41,5 @@ class TestPreparedStatements(Tester):
             session.execute(query_statement, (0,))
         except InvalidRequest as ir:
             print ir
-        except:
+        except Exception:
             raise
--- a/prepared_statements_test.py
+++ b/prepared_statements_test.py
@@ -1,0 +1,45 @@
+from dtest import Tester, debug
+from cassandra import InvalidRequest
+
+KEYSPACE = "foo"
+
+
+class TestPreparedStatements(Tester):
+    """
+    Tests for pushed native protocol notification from Cassandra.
+    """
+
+    def dropped_index_test(self):
+        """
+        Prepared statements using dropped indexes should be handled correctly
+        """
+
+        self.cluster.populate(1).start()
+        node = self.cluster.nodes.values()[0]
+
+        session = self.patient_cql_connection(node)
+        session.execute("""
+            CREATE KEYSPACE IF NOT EXISTS %s
+            WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': '1' }
+            """ % KEYSPACE)
+    
+        session.set_keyspace(KEYSPACE)
+        session.execute("CREATE TABLE IF NOT EXISTS mytable (a int PRIMARY KEY, b int)")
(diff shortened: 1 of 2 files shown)


pushed_notifications_test.py
```diff
@@ -77,8 +86,38 @@ def move_single_node_test(self):
...
-            notification = waiter.wait_for_notification(60.0)
+            notifications = waiter.wait_for_notifications(60.0)
+            self.assertEquals(1, len(notifications))
+            notification = notifications[0]
             change_type = notification["change_type"]
             address, port = notification["address"]
             self.assertEquals("MOVED_NODE", change_type)
             self.assertEquals(self.get_ip_from_node(node1), address)
+
+    @attr('selected')
```
ptnapoleon: What is the purpose of this attribute?

pushed_notifications_test.py
```diff
@@ -77,8 +86,38 @@ def move_single_node_test(self):
...
+    def restart_node_test(self):
+        """
+        Restarting a node should generate exactly one DOWN and one UP notification
+        """        
+
+        self.cluster.populate(2).start()
+
+        node1 = self.cluster.nodes.values()[0]   
+        node2 = self.cluster.nodes.values()[1]
```
ptnapoleon: These two lines could be condensed to `node1, node2 = self.cluster.nodelist()`

prepared_statements_test.py
```diff
@@ -0,0 +1,45 @@
...
+        print "Number of matching rows:", len(session.execute(query_statement, (0,)))
+
+        session.execute("DROP INDEX bindex")
+
+        try:    
+            print "Executing prepared statement with dropped index..."
+            session.execute(query_statement, (0,))
+        except InvalidRequest as ir:
+            print ir
+        except:
```
thobbs: In Python, you should almost always do ``except Exception:`` instead of an open except block.  The reason is that the current code will catch things like ``SystemExit`` and ``KeyboardInterrupt`` (see the Exception heirarchy for more details: https://docs.python.org/2/library/exceptions.html#exception-hierarchy).

ptnapoleon: Good catch. Sorry I missed that.

pushed_notifications_test.py
stef1927: It's just to run only the test with that attribute via the -a option.

{code}
nosetests -s -a 'selected' pushed_notifications_test.py
{code}

This will only run the tests with that attribute.

It can be removed, along with the import.

Is there a better way to only run one test?

pushed_notifications_test.py
stef1927: Got it, thanks!

prepared_statements_test.py
stef1927: Thanks

pushed_notifications_test.py
thobbs: Yes, you can do: ``nosetests module.py:ClassName.method_name``.  You can also leave off the method name to run the whole class.

stef1927: Thanks!

ptnapoleon:
+1

PR #179, opened 2015-03-04
Fixed bug for test of CASSANDRA-7816

stef1927:


aboudreault:
Assigned myself as tester in 7816, will take a look at this asap.

stef1927:
Great, thanks

pushed_notifications_test.py
```diff
@@ -97,12 +101,11 @@ def restart_node_test(self):
         """        
 
         self.cluster.populate(2).start()
-
         node1, node2 = self.cluster.nodelist()
 
         waiter = NotificationWaiter(self, node1, "STATUS_CHANGE", num_notifications=2)
 
-        for i in range(3):
+        for i in range(5):
```
mambocab: Any particular reason to change this?

mambocab:
I'm curious about https://github.com/riptano/cassandra-dtest/pull/179/files#r26330695, but +1

pushed_notifications_test.py
stef1927: Yes occasionally I could not reproduce the multiple UP notifications with only 3 attempts and so I increased it to 5.

mambocab: I figured it was something like that. Thanks!

Issue #182, opened 2015-03-05
bootstrap_test.TestBootstrap.simple_bootstrap_test


ptnapoleon:
https://issues.apache.org/jira/browse/CASSANDRA-8322

mambocab:
I've pulled out a `get_size` function that uses `nodetool info` to pull out the load. It's here:

https://github.com/mambocab/cassandra-dtest/blob/bootstrap-size/bootstrap_test.py#L9

I can move this to `tools.py` if it'd be useful elsewhere, but I'd want to spend a little time making it more robust. Right now it assumes all units are KB, and enforces that with an assert so it doesn't silently say that 2.1 GB is the same as 2.1 KB.

I've seen this fail under 2.1.2 and 2.1.3 with the relationship between the initial and final load on `node1` being wrong, though not by much:

```
FAIL: simple_bootstrap_test (bootstrap_test.TestBootstrap)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mambocab/cstar_src/cassandra-dtest/bootstrap_test.py", line 74, in simple_bootstrap_test
    assert_almost_equal(initial_size, 2 * size1)
  File "/home/mambocab/cstar_src/cassandra-dtest/assertions.py", line 56, in assert_almost_equal
    assert vmin > vmax * (1.0 - error) or vmin == vmax, "values not within %.2f%% of the max: %s" % (error * 100, args)
AssertionError: values not within 16.00% of the max: (466.78, 576.58)

[...]

dtest: DEBUG: initial_size: 466.78
dtest: DEBUG: size1: 288.29
dtest: DEBUG: size2: 280.68
```

There's also an error happening when calling `patient_cql_connection`, shown here:

https://gist.github.com/anonymous/37008ac2a106823c0721

I don't see this error in `counter_tests.py:TestCounters.simple_increment_test`, which also uses `patient_cql_connection`.

ptnapoleon:
That error means one of two things: The cluster didn't start, or the cluster took so long to start that patient timed out. Worth investigating the issue either way.

It's probably not worth moving this into `tools.py` yet, since no one else would be using it.

I'm okay with the test still failing, getting a better size calculation is the first step to stabilizing this test. We can figure out why it continues to fail once this correction is merged.

ptnapoleon:
Fixed by #209

PR #184, opened 2015-03-09
Bootstrap size

mambocab:
Part of addressing #182. Adds a function that uses `nodetool` to calculate the load on a node, and uses that function in `bootstrap_test:TestBootstrap.simple_bootstrap_test`. Caveats:

- As discussed in #182, `get_size` isn't robust enough to be used elsewhere - it's a one-off.
- The test fails -- the relationship between `node1`'s initial size and it's final size is off.
- `patient_cql_connection` throws an error, though it doesn't in other tests, as far as I've seen. See #182.
- Because of the failures mentioned above, I've gotten a little debug-statement happy. These should probably be removed, though maybe not until the problems are fixed.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -4,6 +4,18 @@ from tools import new_node, insert_c1c2, query_c1c2
 from assertions import assert_almost_equal
 from ccmlib.cluster import Cluster
 from cassandra import ConsistencyLevel
+from decimal import Decimal
+
+
+def get_size(node):
+    """Uses nodetool info to get load on node"""
+    info = node.nodetool('info', capture_output=True)[0]
+    load_line = filter(lambda s: s.startswith('Load'),
+                       info.split('\n'))[0].split()
+    load_num, load_units = load_line[2], load_line[3]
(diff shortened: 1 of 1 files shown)


mambocab:
Also: this is 5 separate commits. I'll squash them once I get a +1.

ptnapoleon:
The debug statements are fine. The test is already failing, so I'm not worried about it still failing after this. I'm +1, but want to see what @aboudreault or @knifewine think.

knifewine:
one small nit, might be nice to cast as something like Decimal instead of float, in case someone wants to do math and make decisions with the figures later it will give them a little less rope.

otherwise +1

mambocab:
I've switched out `float` for `Decimal` and squashed. I'll merge after one last sanity check.

mambocab:
Sorry, I wasn't clear -- can someone give me a sanity check on the changed PR?

knifewine:
lgtm

mambocab:
Perfect, thanks @knifewine :fireworks: :tada: :fireworks:

PR #186, opened 2015-03-09
removes unused not_implemented decorator

mambocab:
This decorator wasn't used anywhere. I think an unimplemented test is the usual use case for `unittest.skip`, so I don't see any point to keeping this decorator around. 

Objections?

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -150,13 +150,6 @@ def replace_in_file(filepath, search_replacements):
             line = re.sub(regex, replacement, line)
         sys.stdout.write(line)
 
-def not_implemented(f):
-    def wrapped(obj):
-        obj.skip("this test not implemented")
-        f(obj)
-    wrapped.__name__ = f.__name__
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I don't object. I'd rather them `raise NotImplementedError` anyway so it shows up in JUnit results

mambocab:
Maybe I'm misinterpreting the decorator name. What does `not_implemented` mean? Does it mean "this test is a stub", or "this tests functionality that hasn't been implemented"? I was assuming the former.

In any case, nobody calls it, so it should still be removed by my reckoning.

knifewine:
I think there probably wasn't a distinction. I'm all for just adding test stubs and raising NotImplementedError until we fill them in. For the most part I don't think we should skip tests unless they are actually breaking the jobs.

knifewine:
also, +1

mambocab:
Sweet, thanks :+1: Merging.

PR #189, opened 2015-03-09
increments minimum # of notifications to wait for

mambocab:
This PR fixes a problem with `pushed_notifications_test`. I know @aboudreault know's what's going on with these tests, so he may be the best reviewer.

I'm not sure this is the best solution. It might be better to have a `NotificationWaiter` initialized with `num_notifications=x` to wait for `x+1` notifications. The existing API is a little dangerous -- I think it encourages bugs like the one this PR fixes.

See the commit message, reproduced below:

****

This is a subtle bug in the test logic. In the initial implementation of the first test, we're waiting for 1 notification, so we used the default `num_notifications` value of 1. Because of the way NotificationWaiter works, though, that means it will stop waiting once it observes a single notification. That means that as long as the first notification meets the expectations of the test, the test will pass, no matter how  many more notifications would have been observed. The assertion "the NotificationWaiter must have only observed one notification" really means "the node received 1 *or more* notifications", since the NotificationWaiter stops waiting after the first.

Similarly, the second test's initial implementation will succeed when the first two notifications are 'DOWN' and 'UP', regardless of what notifications follow.

To avoid this, this commit increments the number of notifications to wait for in each of these tests.

It also removes the check in NotificationWaiter that makes sure it observed exactly num_notifications notifications -- that belongs in the test logic, not the observer.

### DIFF
--- a/pushed_notifications_test.py
+++ b/pushed_notifications_test.py
@@ -10,7 +10,7 @@ class NotificationWaiter(object):
     Cassandra over the native protocol.
     """
 
-    def __init__(self, tester, node, notification_type, num_notifications=1):
+    def __init__(self, tester, node, notification_type):
         """
         `address` should be a ccmlib.node.Node instance
         `notification_type` should either be "TOPOLOGY_CHANGE", "STATUS_CHANGE",
@@ -28,7 +28,6 @@ class NotificationWaiter(object):
 
         # the pushed notification
         self.notifications = []
-        self.num_notifications = num_notifications
 
         # register a callback for the notification type
         connection.register_watcher(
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Is this related to #179 at all?

mambocab:
It deals with the same tests, and I found the problem while looking at
#179, but otherwise, no.

mambocab:
The more I think about this, the more I think the fix really should just be at [line 56](https://github.com/mambocab/cassandra-dtest/blob/duplicate-event-test-bug/pushed_notifications_test.py#L56). Instead of:

```
if len(self.notifications) >= self.num_notifications:
    break
```

 it should be

```
if len(self.notifications) > self.num_notifications:
    break
```

mambocab:
New version pushed based on discussion with @aboudreault. Changes the semantics a bit, but I think for the better. The tests still fail in the way I expect, as described in https://issues.apache.org/jira/browse/CASSANDRA-7816.

pushed_notifications_test.py
```diff
@@ -98,12 +97,12 @@ def move_single_node_test(self):
     def restart_node_test(self):
         """
         Restarting a node should generate exactly one DOWN and one UP notification
-        """        
+        """
```
aboudreault: Same nitpick than: https://github.com/riptano/cassandra-dtest/pull/197/files#r26483507

aboudreault:
LGTM! just a minor nitpicking.

pushed_notifications_test.py
mambocab: My mistake -- thanks for pointing it out!

mambocab:
:+1: Thanks!

PR #192, opened 2015-03-10
removes commented-out tests

mambocab:
These tests have been commented out for over a year. They should either be updated or removed; the latter is ready to go with this PR.

If you want to play along at home, check out this branch:

https://github.com/mambocab/cassandra-dtest/tree/uncomment-cqlsh-tests

and run the tests with:

    nosetests -v cql_tests.py:TestCQL.keyspace_test cql_tests.py:TestCQL.table_test cql_tests.py:TestCQL.nan_infinity_test

Results from running on trunk here:

https://gist.github.com/mambocab/9f229774b82bbe3d6be3

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -1035,63 +1035,6 @@ class TestCQL(Tester):
         # Reserved keywords
         assert_invalid(cursor, "CREATE TABLE test1 (select int PRIMARY KEY, column int)", expected=SyntaxException)
 
-    #def keyspace_test(self):
-    #    cursor = self.prepare()
-
-    #    assert_invalid(cursor, "CREATE KEYSPACE test1")
-    #    cursor.execute("CREATE KEYSPACE test2 WITH replication = { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }")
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Who commented them out originally? What was the commit?

mambocab:
All commented out as part of

https://github.com/riptano/cassandra-dtest/commit/e1af52a4683165dfd449adeff0a93e5f4609766d

@pcmanus: thoughts?

pcmanus:
It's unfortunately too old for me to remember why I did that, sorry. This might well have gone in the commit by mistake since that sounds unrelated to the commit intent. The commented out tests don't look crucial but from eyeballing them, they don't look wrong either. Are they not passing when uncommented?

mambocab:
Yeah, they fail. Results from running on trunk here:

https://gist.github.com/mambocab/9f229774b82bbe3d6be3

pcmanus:
Ok, so the 2 first tests (keyspace_test and table_test) should be easy to fix. Both tests are throwing the error we expect them to throw, but it appears that `assert_invalid` doesn't catch those particular errors. I think we just need to add a `expected=SyntaxException` and `expected=AlreadyExists` in the `insert_invalid` calls.

For the last test, the {{nan_infinity_test}} one, seems like we just need to replace `nan` by `float('nan')` (and similar for `inf`).

mambocab:
Awesome, thanks @pcmanus! I can fix these up since I'm already messing with them. I'll let you know if I have more questions.

PR #193, opened 2015-03-11
uncomments and updates 3 commented-out tests

mambocab:
This solves the same problem as #192, but by welcoming the tests back into the fold rather than destroying them.


ptnapoleon:
+1

PR #194, opened 2015-03-12
Add test for local quorum queries while a node is bootstrapping

aboudreault:
Add test for local quorum queries while a node is bootstrapping

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -1,7 +1,11 @@
+import os
 import random
 import time
-
-from dtest import Tester
+import tempfile
+import subprocess
+import tempfile
+import re
+from dtest import Tester, debug
 from tools import new_node, query_c1c2, since, InterruptBootstrap
 from assertions import assert_almost_equal
 from ccmlib.node import NodeError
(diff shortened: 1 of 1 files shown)


ptnapoleon:
@aboudreault What's the status on this?

aboudreault:
I tried to finish that last week... but now getting issues related the FlushWriter in 2.0... should be good to work on this this week.

ptnapoleon:
No rush, just wanted to make sure we still wanted to keep this PR open. We have a lot built up

aboudreault:
ready for review

bootstrap_test.py
```diff
@@ -199,3 +203,69 @@ def manual_bootstrap_test(self):
...
+        node3 = new_node(cluster, data_center='dc2')
+        node3.start(no_wait=True)
+        time.sleep(3)
+
+        with tempfile.TemporaryFile(mode='w+') as tmpfile:
+            if version < "2.1":
+                node1.stress(['-o', 'insert', '-n', '500000', '-t', '5', '-e', 'LOCAL_QUORUM', '-K', '2'],
+                             stdout=tmpfile, stderr=subprocess.STDOUT)
+            else:
+                node1.stress(['user', 'profile='+stress_config.name, 'ops(read=1)',
```
ptnapoleon: It looks like we're reading here, but in the `if version < "2.1":` block, we're inserting. Which is correct?

aboudreault: no, fixing

aboudreault: fixed

ptnapoleon:
LGTM, +1

PR #196, opened 2015-03-13
Slowdown multiple subsequent repairs test to not use 200 threads

aboudreault:
Slowdown multiple subsequent repairs test to not use 200 threads

@mshuler feel free to merge if this is enough to fix your yesterday's issue.


mshuler:
lgtm - this is the sort of minor change that you can just commit and doesn't need to PR

PR #199, opened 2015-03-18
tests for CASSANDRA-8838

yukim:
[CASSANDRA-8838](https://issues.apache.org/jira/browse/CASSANDRA-8838) introduced new feature around bootstrappping/replace node.
Tests are added to verify its functionality.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -1,8 +1,9 @@
 import random, time
 from dtest import debug, Tester
-from tools import new_node, insert_c1c2, query_c1c2
+from tools import new_node, insert_c1c2, query_c1c2, since, InterruptBootstrap
 from assertions import assert_almost_equal
 from ccmlib.cluster import Cluster
+from ccmlib.node import NodeError
 from cassandra import ConsistencyLevel
 
 
@@ -15,6 +16,9 @@ class TestBootstrap(Tester):
             # node that hasn't started yet, and when it does, it gets
             # replayed and everything is fine.
             r'Can\'t send migration request: node.*is down',
+            # ignore streaming error during bootstrap
+            r'Exception encountered during startup',
+            r'Streaming error occurred'
         ]
         Tester.__init__(self, *args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


ptnapoleon:
I'm +1, these are sophisticated tests, nice job. If I could make one nit, it's that the comments on `InterruptBootstrap` calls should explicitly state that the node being killed is doing outbound streaming. It would make it much clearer.

ptnapoleon:
Looking for a second review, @aboudreault or @EnigmaCurry ?

bootstrap_test.py
```diff
@@ -76,3 +80,80 @@ def read_from_bootstrapped_node_test(self):
...
+        t = InterruptBootstrap(node1)
+        t.start()
+
+        # start bootstrapping node3 and wait for streaming
+        node3 = new_node(cluster)
+        node3.set_configuration_options(values={'stream_throughput_outbound_megabits_per_sec': 1})
+        try:
+            node3.start()
+        except NodeError:
+            pass # node doesn't start as expected
```
aboudreault: Can the node3 start for some reason? If so, is it wrong? Assuming so, could we ensure the expected exception with assertRaises ?

bootstrap_test.py
```diff
@@ -76,3 +80,80 @@ def read_from_bootstrapped_node_test(self):
...
+        t.join()
```
aboudreault: same comment than above...

bootstrap_test.py
ptnapoleon: I think the issue is the node fails to start due to the failed bootstrap, right @yukim? `NodeError` is a ccm problem. The logs are still parsed at the end of the test for unexpected errors, and Yuki updated the expected log errors at the top of the test.

aboudreault:
Just a question about a try/except in the code. Otherwiset, LGTM. (y)

yukim:
ccm throws NodeError at the end of ```node#start()``` if the node is not running. (see https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L508)

I am intentionally killing the node to let bootstrap fail, so the node throws NodeError from ```node#start()``` and I need to catch them to keep test running.

ptnapoleon:
Are good to merge then, @aboudreault ?

aboudreault:
Yes!

PR #201, opened 2015-03-20
Add test for CASSANDRA-8933

ptnapoleon:
@pcmanus This test passes on both 2.1-head and trunk. Is there something wrong with the test logic, that isn't exposing the error?


ptnapoleon:
As mentioned in the ticket, this functionality IS working, thus the test should pass. Looking for review from @aboudreault

aboudreault:
LGTM!

PR #203, opened 2015-03-24
Add manual boostrap test for Cassandra-9022

aboudreault:
This test is not supposed to fail when using vnodes, but I haven't restricted it since it could catch a future bug that might happen with/without vnodes. The tests is running 3 times since sometime, it doesn't happen.


ptnapoleon:
+1

PR #205, opened 2015-03-24
Add utility class for interacting with JMX

thobbs:
This is a utility class that's needed for testing CASSANDRA-8993, but should come in handy elsewhere.  It makes it quite easy to use JMX from within python.


aboudreault:
oh, that's cool. I was currently doing  some work with metrics and was thinking about something like that. However, I was more thinking about the ability to tell CCM to record these metrics automatically. So maybe we could add that feature in ccm too based on this work. (not necessary now... just thinking out of loud)

aboudreault:
@thobbs I don't know really these Java libs.... but is there any reason to use Jolokia instead of mx4j ? (just curious)

thobbs:
There are a couple of reasons I went with Jolokia over mx4j.  The first is that with mx4j, we'd need to get the jar into the classpath before starting Cassandra; that's a bit more complicated than simply attaching to a running process with Jolokia.  Second, the mx4j api is xml-centric, whereas Jolokia uses a pretty simple JSON api.

By the way, I tried integrating this into ccm before I opted for dtests.  The main problem that I ran into was around shipping the jar with the ccm package.  If ccm gets installed as an egg (preventable in some cases, but, for example, not preventable when you do ``python setup.py install``, as far as I could tell), we'd have to handle reading from the zip and writing out to a separate file somewhere.  That was looking like a serious pain, and by comparison, using it from the dtests was much simpler.

aboudreault:
I see. Thanks for the info. I'll review the PR later today.

aboudreault:
LGTM! Thanks!

PR #209, opened 2015-03-25
Fix bootstrap size differences

thobbs:
To minimize spurious size differences, I made the following changes:
* Disable compression for the table
* Compact the sstables (probably not needed, but safer)
* Account for the size of the node before loading data

I also added conversions to floats to avoid doing math between floats and Decimals (which results in an error).


ptnapoleon:
+1. The type error was just introduced last week, thanks for fixing it.

PR #210, opened 2015-03-25
CASSANDRA-9007

ptnapoleon:
I already excluded the new file from jenkins. Does just basic data loss validation. A good prototype to discuss what we want this harness to look like long term. 

As is, there is going to be far far too much log parsing for errors. If the harness is going to be in Python, then changes need to be made in stress so the harness can get simple error codes returned.

### DIFF
--- a/validation.py
+++ b/validation.py
@@ -1,0 +1,77 @@
+from dtest import Tester, debug
+from tools import since
+import subprocess, tempfile, os, shutil
+
+@since('3.0')
+class TestValidation(Tester):
+
+    def prepare(self):
+        cluster = self.cluster
+
+        cluster.populate(3).start()
+
+        return cluster
+
+    def stress_write(self, node):
+        node.stress(['write', 'n=1M', '-mode', 'native', 'cql3', '-rate', 'threads=10', '-pop', 'dist=UNIFORM(1..1M)'])
+
+    def stress_read(self, node):
+        # Verify the data
+        tmpfile = tempfile.mktemp()
+        with open(tmpfile, 'w+') as tmp:
+            node.stress(['read', 'n=1M', '-mode', 'native', 'cql3', '-rate', 'threads=10', '-pop', 'dist=UNIFORM(1..1M)'],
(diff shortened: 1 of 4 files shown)


ptnapoleon:
Looking for review @aboudreault

mambocab:
The intent of `reverse_test` wasn't clear to me at first -- that's making sure that failed reads are recognized when a node is down, right? That depends on an implicit RF=1, so I think that should be documented. And if I'm totally off-base, then add a docstring explaining the whole test.

Works for me locally on `trunk`. If that's the intended scope, then +1 after documenting `reverse_test`.

knifewine:
I was similarly confused by base_test (esp since base has a bunch of
different connotations). Docstrings would be nice esp on the tests. Even if
the code is crystal clear they form a statement about purpose that might
help keep people in the future from perverting your test intentions and
confusing things.

mambocab:
> that's making sure that failed reads are recognized when a node is down

I wasn't clear there -- I meant "reads that failed because a node was down are recognized as failures".

validation.py
```diff
@@ -0,0 +1,109 @@
...
+        # Verify the data
+        tmpfile = tempfile.mktemp()
+        with open(tmpfile, 'w+') as tmp:
+            node.stress(['read', 'n=1M', '-mode', 'native', 'cql3', '-rate', 'threads=10', '-pop', 'dist=UNIFORM(1..1M)'],
+                stdout=tmp, stderr=subprocess.STDOUT)
+        return tmpfile
+
+    def validate_stress_output(self, outfile, expect_failure=False, expect_errors=False):
+        """
+        Validates if data was loss, or if stress encountered errors.
```
mambocab: Validates if data was lost, and determine if stress encountered errors.

mambocab:
Clean up the one sentence I commented on:

https://github.com/riptano/cassandra-dtest/pull/210#discussion_r27245145

And then LGTM :tada:

validation.py
ptnapoleon: See I typed `data loss`, then injected the `was` without re-reading

PR #212, opened 2015-03-27
Test for CASSANDRA-8979

yukim:
[CASSANDRA-8979](https://issues.apache.org/jira/browse/CASSANDRA-8979) improved repair not to stream unnecessary gc-able data.
This PR adds test for that.


ptnapoleon:
+1

PR #214, opened 2015-03-31
Correct consistency checks for CL.ONE writes and reads

thobbs:
When writing and reading at ConsistencyLevel.ONE, there's no guarantee
that reads will see all writes.


ptnapoleon:
+1. Thanks for fixing that.

PR #215, opened 2015-03-31
Check sstableloader exit status, error if non-zero

thobbs:
This doesn't fix the failing `snapshot_test.TestSnapshot.test_basic_snapshot_and_restore` test, but exposes the problem more clearly.

### DIFF
--- a/snapshot_test.py
+++ b/snapshot_test.py
@@ -1,9 +1,17 @@
+import glob
+import os
+import shutil
+import subprocess
+import tempfile
+import time
+
 from dtest import Tester, debug
 from tools import replace_in_file, since
-import tempfile, shutil, glob, os, time
 import distutils.dir_util
 
+
 class SnapshotTester(Tester):
+
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


snapshot_test.py
```diff
@@ -39,7 +47,17 @@ def restore_snapshot(self, snapshot_dir, node, ks, cf):
         node_dir = node.get_path()
         snapshot_dir = os.path.join(snapshot_dir, ks, cf)
         ip = node.address()
-        os.system('{node_dir}/bin/sstableloader -d {ip} {snapshot_dir}'.format(**locals()))
+
+        executable = '%s/bin/sstableloader' % (node_dir,)
```
ptnapoleon: Why the `(node_dir,)` and not just `node_dir`?

thobbs: It's just a safer way to do the same thing, especially in terms of handling future changes to the code.  Not really a big deal, just my personal preference.

The alternative ``'{node_dir}'.format(node_dir=node_dir) is also fine, if you prefer that.

ptnapoleon:
+1

PR #221, opened 2015-04-02
repair_test should test both sequential and parallel repair

yukim:
`repair_test.py` should test both sequential and parallel repair.
3.0(trunk) changes default to parallel, and in fact, I found a bug only happens in parallel repair.
Also, 3.0 defaults to incremental repair but we now have `inceremental_repair_test.py`, `repair_test.py` should stick to test only 'full' repairs.


ptnapoleon:
+1

PR #222, opened 2015-04-02
Add test for CASSANDRA-8670

ptnapoleon:
Patch by @aweisberg. Looking for review from @mambocab.

Ariel, this test fails currently. Does it pass with your proposed patch for CASSANDRA-8670?

### DIFF
--- a/largecolumn_test.py
+++ b/largecolumn_test.py
@@ -2,6 +2,10 @@ from dtest import Tester, debug
 from tools import new_node, since
 import time
 
+"""
+Check that inserting and reading large columns to the database doesn't cause off heap memory usage
+that is proportional to the size of the memory read/written.
+"""
 @since('3.0')
 class TestLargeColumn(Tester):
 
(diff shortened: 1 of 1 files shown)


largecolumn_test.py
```diff
@@ -0,0 +1,33 @@
...
+@since('3.0')
+class TestLargeColumn(Tester):
+
+    def stress(self, cluster, node, size):
+        size = str(size);
+        node.stress(['write', 'n=5', "no-warmup", "cl=ALL", "-pop", "seq=1...5", "-schema", "replication(factor=2)", "-col", "n=fixed(1)", "size=fixed(" + size + ")", "-rate", "threads=1"])
+        node.stress(['read', 'n=5', "no-warmup", "cl=ALL", "-pop", "seq=1...5", "-schema", "replication(factor=2)", "-col", "n=fixed(1)", "size=fixed(" + size + ")", "-rate", "threads=1"])
+
+    def directbytes(self, node):
+        return node.nodetool( "gcstats", capture_output=True)[0].split("\n")[1].split()[6]
```
mambocab: This is fragile to changes in `nodetool gcstats`. Maybe check that the first line of the output of `node.nodetool(...)` is what you expect.

largecolumn_test.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def cleanup_test(self):
+        cluster = self.cluster
+        cluster.set_configuration_options( { 'commitlog_segment_size_in_mb' : 128, 'internode_compression' : 'none' })
+        cluster.populate(2).start(jvm_args=[" -Dcassandra.netty_use_heap_allocator=true "])
+        node1, node2 = cluster.nodelist()
+
+        session = self.patient_cql_connection(node1)
+        print "Before stress {0}".format(self.directbytes(node1))
```
mambocab: Should these use `dtest.debug`?

ptnapoleon: Yes, they should

mambocab:
The intent of this test isn't clear. `stress` should be renamed to something like `stress_with_col_size`. I think it'd be helpful to add comments or docstrings explaining:

- the significance of using the different heap allocator, and if that allocator's behavior is what's being tested.
- the significance of setting `commitlog_segment_size_in_mb` to 128, using `1024 * 1024 * 63` as the column size for stress, and testing against `1024 * 1024 * 2`.

mambocab:
Also, this test passes on trunk for me.

ptnapoleon:
@mambocab Think this is good to merge now? It's not as well commented as you wanted, but I'll pass that to @aweisberg

aweisberg:
I'll update the test to address comments.

mambocab:
@aweisberg can you ping me on here when you do? I'll re-review then.

mambocab:
:+1: :fireworks:

PR #223, opened 2015-04-03
Port global_row_key_cache_test to CQL3

thobbs:
Done as part of CASSANDRA-8996


ptnapoleon:
Most amazing patch ever. +1

PR #226, opened 2015-04-07
Test pre-2.1 index sstables are read after upgrade CASSANDRA-9116

beobal:


ptnapoleon:
The way you restrict the versioning assumes that the bug will only re-occur with a regression in 2.1, not in 2.0, since the 2.0 version is frozen. Is this assumption correct, i.e., code changes in 2.0 without breaking anything in 2.1 couldn't cause this?

beobal:
yes, that's right. The problem is that the default location for sstables changed in 2.1, but where there are existing sstables in the old location we continue to use them. The code that implements that logic is broken for 2i sstables. So my assumption is that we're not going to change where the index sstables are written to in 2.0.15+ (of course, some other late regression in 2.0 could break it, but I wasn't trying to cover all possible bugs).

ptnapoleon:
Thanks, it seems like a safe assumption.

ptnapoleon:
This is failing where expected for me. I assume it's passing with your patch for 9116?

beobal:
It is

ptnapoleon:
+1

PR #230, opened 2015-04-08
Fix ttl tests about time precision on GCE

aboudreault:


ptnapoleon:
Do we want to test this on gce first? The changes seem reasonable, and I'm +1.

aboudreault:
If no objection, I don't think it is worth testing it on gce manually. I think failures will be fixed with these changes.

ptnapoleon:
No objection. I trust your judgment.

ptnapoleon:
I'm pulling this in so I can run tests with the changes on cassci

PR #231, opened 2015-04-10
Countupdatetest

shawnkumar:
just adds a test to check that select count(*) is returning the right count

### DIFF
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -210,6 +210,7 @@ class TestUpgradeThroughVersions(Tester):
 
                     self._check_values()
                     self._check_counters()
+                    self._check_select_count()
 
                     debug('Successfully upgraded %d of %d nodes to %s' %
                           (num + 1, len(self.cluster.nodelist()), tag))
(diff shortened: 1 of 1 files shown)


upgrade_through_versions_test.py
```diff
@@ -372,6 +375,23 @@ def _check_counters(self):
...
+        cursor.execute("use upgrade;")
+
+        expected_num_rows = len(self.row_values)
+
+        countquery = SimpleStatement("SELECT COUNT(*) FROM cf;", consistency_level=consistency_level)
+        result = cursor.execute(countquery)
+
+        if result is not None:
+            actual_num_rows = result[0]
+            assert actual_num_rows == expected_num_rows, "SELECT COUNT(*) returned %s when expecting %s" % (actual_num_rows,expected_num_rows)
```
EnigmaCurry: Let's use self.assertEquals here, it provides much better reporting in the case of a failure.

EnigmaCurry:
+1

ptnapoleon:
Can you squash before merge?

ptnapoleon:
Beautiful. +1 from me as well.

PR #234, opened 2015-04-15
convert @require's argument to a string

mambocab:
This allows uses like

    @require(42)
    def func():
        pass

rather than unnecessarily requiring

    @require('42')
    def func():
        pass

Strings will still work; this change just supports the [common use case](https://github.com/riptano/cassandra-dtest/search?utf8=%E2%9C%93&q=%22%40require%22) where the only information passed in represents the integer ID of a Jira ticket.

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -206,7 +206,7 @@ def require(msg):
     """Skips the decorated class or method with a message about which Jira
     ticket it requires."""
     # equivalent to decorating with @unittest.skip
-    return unittest.skip('require ' + msg)
+    return unittest.skip('require ' + str(msg))
 
 class InterruptBootstrap(Thread):
     def __init__(self, node):


ptnapoleon:
What is the need for this change? Require is really just a skip with nicer readability for why we're skipping. The argument is never used.

mambocab:
> What is the need for this t?

It's not strictly necessary; it just makes `@require` a little more Pythonic IMO, and it was easy enough to do.

> The argument is never used.

It's [used to build the message](https://github.com/mambocab/cassandra-dtest/blob/master/tools.py#L207). It uses string concatenation for that, which requires that `msg` be a string. I don't see any need to expose that in the `@require` interface.

ptnapoleon:
>It's used to build the message
I meant it's not used for any logic. But I'm all for making it more Pythonic. +1

PR #235, opened 2015-04-15
Fix the increment of sstable generation numbers

thobbs:
The existing code only worked correctly when a single sstable existed.
For example, if the existing generations were [1, 2], it would increment
them to [2, 3].  The correct behavior is to go to [3, 4].

This fixes the sometimes failing `scrub_test.TestScrub.test_standalone_scrub`


ptnapoleon:
+1

PR #236, opened 2015-04-15
test that memtable size calculation includes tombstones

mambocab:
Defect reported in [#9194](https://issues.apache.org/jira/browse/CASSANDRA-9194).

[Fails](https://gist.github.com/mambocab/87681ed3770e19345a21) on all 2.0 and 2.1 releases going back to 2.0.2. `nodetool cfstats` earlier than that doesn't support the way I get the memtable data size, so I don't know if it worked then.

### DIFF
--- a/deletion_test.py
+++ b/deletion_test.py
@@ -2,6 +2,9 @@ from dtest import Tester
 
 import os, sys, time
 from ccmlib.cluster import Cluster
+from tools import require, since
+from jmxutils import make_mbean, JolokiaAgent
+
 
 class TestDeletion(Tester):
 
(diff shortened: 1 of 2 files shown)


ptnapoleon:
@mambocab I might be terrible here and wonder if it would be better to use the direct JMX interface that @thobbs added.

mambocab:
@ptnapoleon This was a good idea and you should feel good. The test uses JMX now. I left the history for review but I'll squash before merge.

I left `jmxutils` changes in this PR to avoid PRs with dependencies, but I can pull those changes out if you prefer. They change the `make_mbean` signature, but don't break any existing uses in the dtests.

Requesting review from @thobbs, since this touches `jmxutils` and I don't have a great mental model of JMX and MBeans.

thobbs:
+1 on the jmxutils changes, they look good to me.  (I forgot that some of the mbeans were more complex when implementing this.)

ptnapoleon:
Also +1.

PR #237, opened 2015-04-21
Test for CASSANDRA-9220

spodkowinski:
This set of tests can be run to test if server ssl options will have the expected effect on node-to-node communication. It provides a module for interacting with the keytool and will start the cluster with various ssl settings. 
Depends on [CASSANDRA-9220](https://issues.apache.org/jira/browse/CASSANDRA-9220) for hostname verification and log output with detailed handshake errors.
The test needs to have keytool and openssl in the current path.

### DIFF
--- a/sslkeytool.py
+++ b/sslkeytool.py
@@ -1,0 +1,90 @@
+import os
+import os.path
+import tempfile
+import subprocess
+
+
+def generate_credentials(ip, cakeystore=None, cacert=None):
+
+    tmpdir = tempfile.mkdtemp()
+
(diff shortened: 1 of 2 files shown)


mambocab:
@spodkowinski What's the current status on this feature and this test? Just checking to make sure we don't need to do anything just yet.

spodkowinski:
I've rebased and squashed the PR and the test is working again for CASSANDRA-9220. I don't think there's anything to do yet as long as the ticket is still pending.

PR #241, opened 2015-04-22
Properly wait for RPC server to start, schema agreement

thobbs:
The test is typically failing against 2.0 because the thrift connection
is being opened before the RPC server has started.  I also saw failures
due to querying the new tables before there was schema agreement, so I
added a check for that.


ptnapoleon:
+1!

PR #243, opened 2015-04-27
Delete last test dir after closing file handle.

josh-mckenzie:
Getting errors in CI due to having a handle open when we attempt to delete. Trivial patch to resolve that.


ptnapoleon:
+1 Pending Jenkins results with this patch

PR #244, opened 2015-04-27
JSON tests

knifewine:
I built a little doctest-cql supporting code to try this as a test style. These are basically glorified doctests, with the added benefit that they can be a CQL/json syntax reference.

I think it worked nicely for testing a CQL-impacting feature while it was in progress without concerning myself too much with a driver.

If others find the doctest style to be useful we can keep them this way, and possibly relocate the generic support code out of json_tests.py . If anyone dislikes this style of test or can find a good reason not to test this way, we can likely port these over with some work!


knifewine:
CC @riptano/cassandra_test_engineering

Apparently CASSANDRA-7970 finally got committed so I just realized today these tests can now be merged in. Let me know if you have any strong thoughts on the code, and sorry for the giant diff.

knifewine:
Here's a recent run of the json tests running successfully against c* trunk: https://cassci.datastax.com/view/All_Jobs/job/json_doctest_scratch/31/

ptnapoleon:
What is the reasoning behind this format? Just simpler given that JSON is all string based? I like it so far, giving in an in-depth reading still.

knifewine:
Yeah, I wanted to test in a very literal way. I also wanted to prototype using tests as a CQL reference that's also executable to see if the idea catches on (we could also leverage this for any cqlsh tests that live in dtest). Didn't want to worry about the driver not being able to deal with the input/output nor depend on an unreleased driver version, or the driver possibly munging values before giving the data back.

Also it was easy to port over from the manual test plan I started with :)

ptnapoleon:
I'm +1, but I think a second reviewer would be nice.

mambocab:
I'm +1. The tests look good, and I think this is a great testing ground for docstring-based tests.

aboudreault:
yep, I like that Russ. Great work! +1

PR #245, opened 2015-04-28
jmx test + nodetool

shawnkumar:
Additional Testing To Do


ptnapoleon:
What's the status of this?

PR #247, opened 2015-04-29
add metadata reset test

tjake:
This adds a test which does a concurrent compaction, repair and read.  In the middle of this we call nodetool resetlocalschema.  Causes issues. Written for 9249

### DIFF
--- a/metadata_tests.py
+++ b/metadata_tests.py
@@ -1,0 +1,80 @@
+import random
+import subprocess
+import time
+import tempfile
+import threading
+
+from dtest import debug, Tester
+from tools import new_node, query_c1c2, since, InterruptBootstrap
+from assertions import assert_almost_equal
+from ccmlib.node import NodeError
+from cassandra import ConsistencyLevel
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Why the `time.sleep(5)`'s?

tjake:
If you don't the nodetool() calls lockup on file issues

ptnapoleon:
Oh. Well that's unpleasant. Looks very nice though, +1 from me. @mambocab, thoughts?

mambocab:
So this is meant to pass when 9249 is fixed? If so, add `@require(9249)` and I'm +1. Maybe add another test that just runs compaction and resets the schema, as described in 9249 -- if that's the error we're testing for, a minimal test case would be good.

ptnapoleon:
If 9249 is already patch available, why not just wait to merge instead of using a require?

mambocab:
> why not just wait to merge instead of using a require?

Good point, +1 to that.

tjake:
Ok 9249 was committed. This issue will trigger another issue 9270

Issue #246, opened 2015-04-29
`__filter_errors` in `tearDown` TypeError


mambocab:
[When `__filter_errors` is called in `Tester.tearDown`](https://github.com/riptano/cassandra-dtest/blob/master/dtest.py#L430) it takes a list generated by `ccmlib.node.grep_log_for_errors` and [passes each element to `re.search` as the second argument](https://github.com/riptano/cassandra-dtest/blob/master/dtest.py#L487). Unfortunately, the elements generated by `grep_log_for_errors` [are lists](https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L306), which leads to `TypeError`s when there are actually errors in the log. [This](https://gist.github.com/mambocab/233eb953c4e47236aa3c) is an example list, and [this](https://gist.github.com/mambocab/3311787a2a0b066d1625) is the resulting error from `nosetests`.

Am I understanding this correctly? If so, then `__filter_errors`, as called by `Tester.tearDown`, shouldn't have ever worked when there are actually errors.

ptnapoleon:
This is working locally for me, and on cassci, even though it seems like it shouldn't.

ptnapoleon:
Appears to definitely be windows only

mambocab:
Reproduced on Linux. Used `UPGRADE_PATH=1_2:2_1 nosetests -x upgrade_through_versions_test.py:TestRandomPartitionerUpgrade` to trigger an ignored error. Output [here](https://gist.github.com/mambocab/afe95bcfb20ce9f22ad6#file-linux-txt-L215). 

That said, [here's](http://cassci.datastax.com/job/cassandra-2.1_dtest/lastCompletedBuild/testReport/counter_tests/TestCounters/upgrade_test/) an instance that Philip pointed out where it worked on Linux. No idea why.

mambocab:
Fixed in #248. Closing.

PR #248, opened 2015-04-29
Tester.teardown: accounts for lists of error lines

mambocab:
Addresses #246. `ccmlib.Node.grep_log_for_errors` returns a list of lists of strings. This fix joins each of the lists into a single string to be compared in `__filter_errors`.

To demonstrate the problem and its fix, run

```UPGRADE_PATH=1_2:2_1 nosetests -x upgrade_through_versions_test.py:TestRandomPartitionerUpgrade```

That particular upgrade path will always throw an error, and since the tests in this `TestRandomPartitionerUpgrade` use the log filter, the errors will be used in the regex comparison that was throwing the `TypeError`. It should error out with a `TypeError` on HEAD fail correctly after applying this change.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -427,7 +427,8 @@ class Tester(TestCase):
         try:
             for node in self.cluster.nodelist():
                 if self.allow_log_errors == False:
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

PR #249, opened 2015-04-30
Add instructions for installing doxygen

ptnapoleon:
Are the instructions clear enough? Should more examples be added? What would you like to see added or changed?


ptnapoleon:
/cc @EnigmaCurry @aboudreault @mambocab @shawnkumar @knifewine @mshuler 

I'd like feedback from everyone, if possible. We'll all be writing test plans this way now.

CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
...
+        """
+            pass
+
+To run doxygen to generate HTML from these test plans, you will need to do the following
+
+* Unzip doxygen/doxypy-0.4.2.tar.gz and install it
+        cd doxygen/
+        tar xvf doxypy-0.4.2.tar.gz
+        cd doxypy-0.4.2
+        sudo python setup.py install
```
aboudreault: Not sure why pypi said Latest version: 0.4.2 and we can only install 0.4.1. We should be able to use pip sometime soon I guess?

aboudreault:
Seems to be clear enough IMO.

CONTRIBUTING.md
ptnapoleon: Doxypy's website recommends you install directly. I don't think there will be a move to using pip.

mambocab:
A few things:

- If `@test_category` is going away, how about we just not document it?
- I'm not clear on how and when to use `@expected_result` and `@expected_errors`. Unless we have some compelling use case, I think we should hide these by not mentioning them in the documentation -- I'd rather have a readable test than this extra metadata to maintain. If we leave them in, an example would be helpful.
- I don't like the different semantics between the `@since` annotation and the `@since` decorator. I'm assuming that Doxygen's `@since` can't be changed, so maybe renaming the decorator is in order.

CONTRIBUTING.md
EnigmaCurry: They messed up the pypy packaging, but it can be installed via:

    pip install --allow-external doxypy --allow-unverified doxypy doxypy

EnigmaCurry: If we leave the tarball in the the repo, the instuctions can be simplified to just be:

    sudo pip install doxygen/doxypy-0.4.2.tar.gz

ptnapoleon:
1. Doxygen's `since` can't be changed. So we would need to rename our decorator in that case.
2. test_category will NOT be going away. Instead what will happen, is I will be writing a Doxygen pre-processor that will be filling that out automatically based on the file name and class name. Since we use those to hold essentially the same information. There would be no need to be redundant.
3. Example for an auth test 

    @expected_errors [ArgumentError] When an empty username and password is provided.#    
 @expected_errors [Cassandra::Errors::AuthenticationError] When an invalid username and password is provided.
    @expected_result Username and password should not successfully authenticate to the Cassandra cluster.

mambocab:
@ptnapoleon Gotcha on 1 and 2. I just wasn't sure what "deprecated" meant for `test_category`.

Re: 3: I see - I still think that, e.g., good assertions [like this](https://github.com/riptano/cassandra-dtest/blob/master/auth_test.py#L210) make that kind of annotation redundant and a maintenance burden. But, if you think it'd be helpful for us or for test plan readers, let's give it a try.

+1, excited to see how inline test plans get used.

ptnapoleon:
I personally feel that most of that would go into a good description. But as I said, there is no need to use all of the tags. So if you have no particular asserts of that fashion, no need to use the tag. However, it does help if someone doesn't want to read the description, and only wants to skim for the various assertions.

mambocab:
I buy it. Thanks for the explanation.

PR #254, opened 2015-05-05
Remove broken func_auth_test

thobbs:
This test doesn't work with current trunk, and its coverage is
superceded by the new tests in auth_roles_test from CASSANDRA-7557.


ptnapoleon:
I see the superceding tests. +1

PR #255, opened 2015-05-06
Validate data through CQL query, not stress

thobbs:
Stress was not being used correctly for validation of the data, and it's easy to validate the data normally through CQL.

This fixes the failing ``manual_bootstrap_test()``.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -1,9 +1,7 @@
 import random
-import subprocess
 import time
-import tempfile
 
-from dtest import debug, Tester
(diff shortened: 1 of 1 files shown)


ptnapoleon:
This validation is definitely more correct. We seem to have seriously intended to use stress's sequential mode for the write and the read operation, instead of the sample mode. Maybe it's still worth it to do so? Or is this method better/just as good? Otherwise, +1.

aboudreault:
+1. That is effectively a better way to validate the data.

PR #256, opened 2015-05-07
Tests for CASSANDRA-7066 (standalone sstablelister)

stef1927:
We've introduced a new standalone tool (sstablelister) for CASSANDRA-7066 and these are some basic tests for it.

### DIFF
--- a/sstablelist_test.py
+++ b/sstablelist_test.py
@@ -1,0 +1,145 @@
+from dtest import Tester, debug
+from tools import since, require, InterruptCompaction
+from ccmlib import common
+
+import subprocess
+import glob
+import os
+
+# These must match the stress schema names
+KeyspaceName = 'keyspace1'
+TableName = 'standard1'
+
+@require('7066')
+@since('3.0')
+class SSTableListTest(Tester):
+
+    def listfiles_oncompaction_test(self):
+        """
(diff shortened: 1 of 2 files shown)


sstablelist_test.py
```diff
@@ -0,0 +1,145 @@
...
+        else:
+            args = [ tool_bin, '--type', type, ks, table]
+
+        p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        (stdin, stderr) = p.communicate()
+
+        if len(stderr) > 0:
+            debug("Error invoking sstablelister : %s " %(stderr,))
+            assert false
```
ptnapoleon: This should be `assert False`, not `assert false`.

ptnapoleon:
Other than the incorrect assert, +1

sstablelist_test.py
aboudreault: suggestion: would it be clearer to do:
```
debug(stderr)
assert False, "Error invoking sstablelister"
```

Will that make the error more obvious in the exception output? not a strong opinion, it's up to you.

ptnapoleon: I prefer your suggestion, Alan.

stef1927:
Thanks for picking it up, I fixed it with Alan's suggestion and squashed it: commit 4486c1e.

PR #257, opened 2015-05-07
uses platform-agnostic executable-specifying code

mambocab:
The previous version built the path with string formatting. This patch makes it use ccm's cross-platform tools for getting paths for executables.


mambocab:
I should note: this does change the way this works, slightly. The existing implementation uses the `sstableloader` executable that ccm copies into a node-specific `bin` directory. This implementation uses the one provided by `Node.get_tool`, which currently uses the `sstableloader` in the C* installation directory. Shouldn't affect behavior.

ptnapoleon:
Belated +1

PR #260, opened 2015-05-09
Add cassandra-attack.jar from CASSANDRA-6285

mshuler:
Some devs were a little stumped as to where this jar lived. In testing, we have been wget'ing the file from the test master server, prior to dtest runs. Let's just add the jar to dtest, so there is no mystery.


ptnapoleon:
+1

PR #261, opened 2015-05-10
Add token-generator dtest

snazy:
dtest for CASSANDRA-9300

### DIFF
--- a/token_generator.py
+++ b/token_generator.py
@@ -1,0 +1,161 @@
+# coding: utf-8
+
+import re
+import subprocess
+import time
+
+from ccmlib import common
+from dtest import Tester
+from tools import rows_to_list
+from cassandra.util import sortedset
+
+class TokenGenerator(Tester):
+    """
+    Basic tools/bin/token-generator test.
+    """
+
+    def call_token_generator(self, install_dir, randomPart, nodes):
+        executable = install_dir + "/tools/bin/token-generator"
+        if common.is_win():
+            executable += ".bat"
+
+        args = [executable]
+        for n in nodes:
+            args.append(str(n))
+
+        if randomPart:
+            if randomPart:
+                args.insert(1, "--random")
+            else:
+                args.insert(1, "--murmur3")
+
+        token_gen_output = subprocess.check_output(args)
+        lines = token_gen_output.split("\n")
+        dc_tokens = None
+        generated_tokens = []
+        for line in lines:
+            if line.startswith("DC #"):
+                if dc_tokens is not None:
+                    self.assertGreater(dc_tokens.__len__(), 0, "dc_tokens is empty from token-generator %r" % args)
+                    generated_tokens.append(dc_tokens)
+                dc_tokens = []
+            else:
+                if line.__len__() > 0:
+                    m = re.search("^  Node #(\d+): [ ]*([-]?\d+)$", line)
+                    self.assertIsNotNone(m, "Line \"%r\" does not match pattern from token-generator %r" % (line, args))
+                    node_num = int(m.group(1))
+                    node_token = int(m.group(2))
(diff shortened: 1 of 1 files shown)


token_generator.py
```diff
@@ -0,0 +1,167 @@
...
+import re
+import subprocess
+import time
+
+from ccmlib import common
+from dtest import Tester, canReuseCluster, freshCluster
+from tools import rows_to_list
+from cassandra.util import sortedset
+
+cql_version = "3.0.0"
```
ptnapoleon: This is actually not necessary.

token_generator.py
```diff
@@ -0,0 +1,167 @@
...
+        # remove these from cluster options - otherwise node's config would be overridden with cluster._config_options_
+        cluster._config_options.__delitem__('num_tokens')
+        cluster._config_options.__delitem__('initial_token')
+
+        assert not cluster.nodelist(), "nodelist() already initialized"
+        cluster.populate(nodes, use_vnodes=False, tokens=generated_tokens[0]).start(wait_for_binary_proto=True)
+        time.sleep(0.2)
+
+        node = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node, version=cql_version)
```
ptnapoleon: Remove the `, version=cql_version`

token_generator.py
```diff
@@ -0,0 +1,167 @@
...
+                    all_tokens.add(tok)
+            assert all_tokens.__len__() == node_count, "Number of tokens %r and number of nodes %r does not match for %r" % (all_tokens.__len__(), node_count, dc_nodes)
+
+    def multi_dc_tokens_default_test(self):
+        self._multi_dc_tokens()
+
+    def multi_dc_tokens_murmur3_test(self):
+        self._multi_dc_tokens(False)
+
+    def multi_dc_tokens_random_test(self):
```
ptnapoleon: Why can these reuse the cluster, if the single dc tests can not?

snazy: The last three only call token-generator tool and don't need a cluster instance.
Only the first three tests start a 3-node cluster.

token_generator.py
snazy: ok

token_generator.py
snazy: ok

token_generator.py
ptnapoleon: Well in that case there's no need to use the `@canReuseCluster` functionality at all. The first three tests will default to not sharing a cluster, and since the latter three don't even create a cluster, there is nothing to reuse.

snazy: ok - removed the annotation

ptnapoleon: You should also remove all the `@freshCluster` annotations.

token_generator.py
```diff
@@ -0,0 +1,164 @@
...
+                    generated_tokens.append(dc_tokens)
+                dc_tokens = []
+            else:
+                if line.__len__() > 0:
+                    m = re.search("^  Node #(\d+): [ ]*([-]?\d+)$", line)
+                    assert m is not None, "Line \"%r\" does not match pattern from token-generator %r" % (line, args)
+                    node_num = int(m.group(1))
+                    node_token = int(m.group(2))
+                    dc_tokens.append(node_token)
+                    assert node_num == dc_tokens.__len__(), "invalid token count from token-generator %r" % args
```
ptnapoleon: It would be great, but not necessary, if for numeric equality assertions, you used the assertEqual function in assertions.py

ptnapoleon:
Other than remaining comments, +1

token_generator.py
snazy: removed

snazy:
Rebased branch against current master

PR #269, opened 2015-05-12
Update test for 9299

ptnapoleon:
@knifewine and @aboudreault to review

see CASSANDRA-9299

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -2,12 +2,13 @@ import time
 import uuid
 
 from cassandra import ConsistencyLevel as CL
-from cassandra import InvalidRequest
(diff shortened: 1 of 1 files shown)


ptnapoleon:
The logic here is that we only count the actual tombstones being read in a given page now. So we needed to use a bigger fetch_size than the tombstone_failure_threshold and I didn't want to muck with the check_all_paging_results logic for a special case.

ptnapoleon:
Also see 9374d07da2b54a42360106d740690b2e4a832803 and a83ccd803472407676a24ed613bee5889f5befa0

knifewine:
seems fine to me
+1

PR #271, opened 2015-05-13
Fix periodic scrub test failures

thobbs:
This breaks up the sstable generation incrementing process to work on
each table (or index) individually instead of all at once.  When all
sstables were worked on at once, if a single table/index had a
generation that the others did not, the incrementing process would
over-increment those other tables.


ptnapoleon:
Seems reasonable, +1.

Issue #238, opened 2015-04-21
Modify require decorator


ptnapoleon:
Currently the `require` decorator is just a glorified skip. To better serve our new testing model, `require` should be modified to accept the name of a git branch as an argument. This branch will correspond to the name of a git branch on a Cassandra fork. This will allow developers to commit new tests to dtest that will run in CI, but only against the fork with the code those tests are running against. Then, when the patch is merged, merging the tests is as simple as removing the decorator.

mambocab:
closed by https://github.com/riptano/cassandra-dtest/commit/dcaf31e539bfafb611077702073aadeceec8f829

PR #274, opened 2015-05-13
Require filter branch

mambocab:
This addresses #238. My request for review has 3 parts:

1. Verify that the name-matching behavior described in `@require`'s docstring is the desired behavior.
1. This makes a hopefully non-invasive change to `dtest` and `dtest.Tester`. Previously, the directory used was calculated at object-definition/method-call time by grabbing `CASSANDRA_DIR` from the environment. Now, it's calculated at load time. This allows `@require` to grab the directory of the currently-running C* install without actually starting a cluster. Please check that this change makes sense before +1'ing this patch.
1. Since this behavior depends on some external state, I'm concerned that it :tada: _works on my machine_ :tada:. I'd like a reviewer to run the [shell script with the tests linked at this gist](https://gist.github.com/mambocab/28805fef05fda0a11c86). Just follow the instructions in the comment at the bottom.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -28,6 +28,7 @@ config = ConfigParser.RawConfigParser()
 if len(config.read(os.path.expanduser('~/.cassandra-dtest'))) > 0:
     if config.has_option('main', 'default_dir'):
         DEFAULT_DIR=os.path.expanduser(config.get('main', 'default_dir'))
+CASSANDRA_DIR = os.environ.get('CASSANDRA_DIR', DEFAULT_DIR)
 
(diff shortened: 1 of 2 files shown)


ptnapoleon:
Your docstring for `require` should explicitly specify that the git branch we're checking is the one for the checkout of cassandra and not the dtest branch. It's implied, but not clear to someone coming in without context.

ptnapoleon:
Works for me, and it all looks good. +1!

mambocab:
Docstring fixed. If it looks good I'll squash the last 2 commits and merge.

ptnapoleon:
+1 on squash and merge.

Issue #188, opened 2015-03-09
Reorganize package structure to support specific driver versions


thobbs:
We've been working on some protocol changes for the v4 native protocol.  Tests for the native protocol generally require a driver, but the python driver doesn't have support for the new features yet.  So, to properly write tests, we need to run with a forked version of the python driver.

I attempted to make this possible with a test-level decorator, but it's too tricky to replace/reload a package completely in python.

I think the next best option might be a combination of two things:
* An environment variable to specify a repo and branch of the python driver to use for the tests. I already have code to handle a git clone and checkout of the driver, given a repository and branch name.
* A test decorator to skip the test unless that version of the python driver is in use

The main change that would be required is to create a proper package structure for the dtests.  That probably looks like a ``dtests`` directory with the test modules in it, along with an ``__init__.py`` file.  The ``__init__.py`` file is key, because it's loaded before any of the test modules.  That's where the environment variable would be checked, allowing us to import the python driver from a specific location.  If we do this anywhere else, there's no way to guarantee that the driver package hasn't already been imported.

In the short-term, this would cause some pain around updating test runners and imports.  In the long-run, though, a proper package structure should make it easier to do things like this (and hopefully make the code a little cleaner).

What do you guys think?

knifewine:
+1 as long as we can keep the functionality unsurprising for everyday uses of dtest (such as using nose locally to run a handful of tests).

on the topic of reorg I think we should consider moving test modules into separate functional area subpackages (subjective, yes) so there's room for multiple modules within a given functional area, and helper code can have its own module if desired.

thobbs:
bq. +1 as long as we can keep the functionality unsurprising for everyday uses of dtest (such as using nose locally to run a handful of tests).

Yup, it would remain easy to use nose.  The main difference would be pointing nose at a directory (although I think it might work without a directory, too).

bq. I think we should consider moving test modules into separate functional area subpackages

I'm fine with doing this in my changes, but I don't have any strong opinions on what the layout should be.  Feel free to suggest something if you'd like me to make those changes in my PR.

ptnapoleon:
I don't think we need to make you do additional organization of dtest. As long as your changes are future-compatible, i.e., we don't need to rework your changes when we do org dtest, then that should be sufficient.

ptnapoleon:
@thobbs Seems like we are +1 on this.

PR #278, opened 2015-05-15
Cassandra 9306: cqlsh COPY tests

mambocab:
This adds a suite of tests for the COPY command in cqlsh, as described [here](https://issues.apache.org/jira/browse/CASSANDRA-9306).

Note that I've broken off the tests into their own module, as discussed in #188 -- this kept each test file a little more managably-sized and helped me organize utility functions. I can jam all the tests back into one file if anyone has strong feelings, but personally, I prefer them this way.

I've put off testing against a cluster with a failing node -- it may be better suited as part of [9172](https://issues.apache.org/jira/browse/CASSANDRA-9172), or part of the kitchen sink tests.

Requesting review from @thobbs, since he filed the Jira for these.

I'll squash before merging.


ptnapoleon:
You need to merge `master` into this branch and resolve the merge conflicts.

ptnapoleon:
Also you will definitely need to squash at least some of these commits away.

cqlsh_tests/cqlsh_tools.py
```diff
@@ -0,0 +1,39 @@
...
+from nose.tools import assert_items_equal
+
+class DummyColorMap(object):
+    def __getitem__(self, *args):
+        return ''
+
+
+def csv_rows(filename, delimiter=None):
+    '''
+    Given a filename, opens a csv file and yeilds it line by line.
```
ptnapoleon: Typo in `yields`

cqlsh_tests/cqlsh_tools.py
```diff
@@ -0,0 +1,39 @@
...
+    '''
+    reader_opts = {}
+    if delimiter is not None:
+        reader_opts['delimiter'] = delimiter
+    with open(filename, 'r') as csvfile:
+        for row in csv.reader(csvfile, **reader_opts):
+            yield row
+
+
+def assert_csvs_items_equal(filename1, filename2):
```
ptnapoleon: Why `csvs` here?

ptnapoleon:
I'm sorry, but I think you need to comment these so as to satisfy `CONTRIBUTING.md`.

cqlsh_tests/cqlsh_tools.py
mambocab: You mean in the name `assert_csvs_items_equal`? Because it compares 2 csv files, so `csv`, but plural. Happy to change if you prefer a different name.

ptnapoleon: The built in asserts use singular here, such as `assertListEqual`, but it's really a personal choice, so I'll leave it up to you.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -0,0 +1,529 @@
...
+            )""")
+
+        data = [[1, 'no'], [2, 'Yes'],
+                [3, 'True'], [4, 'false']]
+
+        tempfile = NamedTemporaryFile()
+        write_rows_to_csv(tempfile.name, data)
+
+        self.node1.run_cqlsh(
+            """COPY ks.testquoted ("IdNumber", "select") FROM '{name}'""".format(name=tempfile.name))
```
thobbs: Can you also test this without explicitly specifying the column names in the COPY command?

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -0,0 +1,529 @@
...
+            )""")
+
+        data = [[1, 'no'], [2, 'Yes'],
+                [3, 'True'], [4, 'false']]
+        insert_statement = self.session.prepare("""INSERT INTO testquoted ("IdNumber", "select") VALUES (?, ?)""")
+        execute_concurrent_with_args(self.session, insert_statement, data)
+
+        tempfile = NamedTemporaryFile()
+        self.node1.run_cqlsh(
+            """COPY ks.testquoted ("IdNumber", "select") TO '{name}'""".format(name=tempfile.name))
```
thobbs: Ditto

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def test_read_invalid_text(self):
+        self.data_validation_on_read_template('test', expect_invalid=True)
+
+    # The next two tests fail due to differences in cqlsh formatting. At
+    # initialization time, cqlsh monkey-patches the Cassandra driver so blobs
+    # are decoded differently. If this changes, un-skip these tests. In the
+    # meantime, the round-trip test for all datatypes shows that COPY works,
+    # though the contents of the CSV aren't defined in cqlshlib.
+    @unittest.skip('fails due to formatting differences in and out of cqlsh')
```
thobbs: There are a couple of fixes for this I can think of.  One is to essentially do the same thing that cqlsh does and monkeypatch the driver during this test.  The other would be to process the CSV or CQL results to convert the blob columns to a matching format.

ptnapoleon:
Does nose pick up this file when run from the `cassandra-dtest/` directory?

ptnapoleon:
This seems like it could use the cluster reuse functionality, since you aren't changing cluster topology between tests. It would drastically improve run time.

mambocab:
> Does nose pick up these tests when run from the `cassandra-dtest/` directory?

Yep. You can verify with `nosetests --collect-only -v 2>&1 | grep cqlsh`.

> This seems like it could use the cluster reuse functionality

Can do. Working on a PR addressing comments now.

mambocab:
I believe I've addressed all comments. @ptnapoleon and @thobbs, can you re-review? Changes since initial push can be viewed [here](https://github.com/mambocab/cassandra-dtest/compare/dbfe0da...CASSANDRA-9306).

Again, I'll squash before commit.

cqlsh_tests/cqlsh_copy_tests.py
mambocab: addressed in https://github.com/mambocab/cassandra-dtest/commit/70894809a37fb3a291f04712b38171e5d08c6747

ptnapoleon:
All my reservations are gone. Trusting you ran the tests here. +1. No comment on the changes Tyler wanted, I didn't check those out.

thobbs:
My comments all seem to be addressed as well, so +1.  Thanks!

mambocab:
Squashed into 2 commits so there's only one commit where `cqlsh_tests.py`'s content changes. Merging.

PR #281, opened 2015-05-18
add test for CASSANDRA-8801

mambocab:
A node should not be able to rejoin a cluster after being decommissioned.

Should I lower the timeout on the log-watching? It'd be nice for failure not to take a full 10 minutes, but I don't want the test to flap, either.

Addresses https://issues.apache.org/jira/browse/CASSANDRA-8801

### DIFF
--- a/topology_test.py
+++ b/topology_test.py
@@ -1,9 +1,11 @@
 from dtest import Tester
-from tools import insert_c1c2, query_c1c2, no_vnodes, new_node
+from tools import insert_c1c2, query_c1c2, no_vnodes, new_node, debug, require, since
 from assertions import assert_almost_equal
 
 import os, sys, time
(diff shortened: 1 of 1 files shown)


mambocab:
I should mention: locally, this fails on trunk and succeeds with @driftx's patch, v2, as linked at the Jira ticket.

ptnapoleon:
What's the longest you've ever seen it take to show up in the log? I definitely think you should use a shorter time. Maybe ~2-3x the expected time until it prints.

ptnapoleon:
Otherwise, +1

mambocab:
Force-pushed with a generous 15s timeout. Ready to merge if the changed version gets a +1.

ptnapoleon:
+1

PR #283, opened 2015-05-19
makes @require work with regex argument

### LINKED ISSUE #282: Turn require into a regex, not a string match
This will help us use it for running on multiple branches.

@mambocab Thoughts?

mambocab:
re: the discussion in #282: I think regexes are a good fit here. The implementation is pretty simple, the functionality subsumes an `*args`-based approach, and the behavior safely falls back to the way it was before, where it just did case-insensitive string matching, if you just don't use regex escape chars.

closes #282.


ptnapoleon:
+1

PR #285, opened 2015-05-20
Set max_version for some upgrade tests

thobbs:
These tests are only designed to be run on 2.1 at the highest, and for
2.2 and later they fail due to Java 8 compilation errors.

### DIFF
--- a/secondary_indexes_test.py
+++ b/secondary_indexes_test.py
@@ -1,6 +1,8 @@
-import os, random, re, time, uuid
+import random
+import re
+import time
+import uuid
 
-from ccmlib.common import get_version_from_build
 from dtest import Tester, debug
 from tools import since
 from assertions import assert_invalid, assert_one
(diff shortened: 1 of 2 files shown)


ptnapoleon:
I think you should be able to set "2.2" as the max_version, instead of "2.1.99". `"2.2.0" > "2.2"` returns true, so that should exclude any 2.2.X version appropriately.

thobbs:
I checked the ``@since`` code and it's using ``LooseVersion`` for comparison instead of strings, so "2.2" wouldn't work.

ptnapoleon:
Can you use `@since(max_version='2.1.x')` the way auth_test does then? It looks a lot nicer than `2.1.99`. Otherwise, +1, feel free to merge.

thobbs:
Sure, I didn't know that LooseVersion would handle that correctly, thanks.  Done.

PR #286, opened 2015-05-20
CASSANDRA-9409: Add tests to ensure keyspace isolation for UDA, UDF and UDT

aboudreault:


aboudreault:
@ptnapoleon or @mambocab to review.

user_functions_test.py
```diff
@@ -194,3 +194,38 @@ def udf_with_udt_test(self):
         assert_one(session, "select sum(funk(udt)) from tab", [6])
 
         assert_invalid(session, "drop type test;")
+
+    @since('2.2')
+    def udf_with_udt_keyspace_isolation_test(self):
+        "CASSANDRA-9409 - Ensure function dont allow a UDT from another keyspace"
```
ptnapoleon: These need to be docstrings, so use `""""`, not just `"`". Also, if you could format it according to the contribution guidelines, that would be great.

ptnapoleon:
+1 to the tests, but you need to make the docstring fixes first.

aboudreault:
Good point for the docstring formatting. Fixed.

user_functions_test.py
```diff
@@ -194,3 +194,45 @@ def udf_with_udt_test(self):
...
         assert_invalid(session, "drop type test;")
+
+    @since('2.2')
+    def udf_with_udt_keyspace_isolation_test(self):
+        """
+        Ensure functions dont allow a UDT from another keyspace
+        @jira_ticket CASSANDRA-9409
+        @since 2.2
+        """
+        session = self.prepare()
```
aboudreault: @ptnapoleon  is the docstring ok like that? Or I should really ad expected_result etc.?

mambocab:
LGTM too.

user_functions_test.py
ptnapoleon: This is perfect

PR #287, opened 2015-05-21
Fix initial_token for simple_bootstrap_test

thobbs:
The test has been failing because the ``intial_token`` was not being
properly set, and with 32 random tokens there was some variation in the
data distribution.

The root of the problem is that when ``DISABLE_VNODES=false``, the
cluster-level config sets ``{'initial_token': None}``.  In ccm, the cluster
(and node-level) config options are applied _after_ ``self.initial_token``,
and ``self.initial_token`` is not actually in the node-level config.  The
result was that no initial_token was being set.  It seems like this
should be improved in ccm, but in the meantime, this is a simple fix for
the tests.


ptnapoleon:
+1

PR #289, opened 2015-05-21
fixes kwarg passed to unittest.skip()

mambocab:
`unittest.skip` doesn't have that kwarg!


ptnapoleon:
Belated +1. Sorry for the trigger finger.

PR #292, opened 2015-05-22
skip test that fails because of jolokia problems

mambocab:
This test is failing because of an error connecting via Jolokia, e.g. [here](http://cassci.datastax.com/view/trunk/job/trunk_dtest/163/testReport/deletion_test/TestDeletion/tombstone_size_test/). thobbs, author of our `jmxutils`, is on the scene, but for the moment, let's skip this test since we know it fails.

### DIFF
--- a/deletion_test.py
+++ b/deletion_test.py
@@ -1,6 +1,6 @@
 from dtest import Tester
 
-import os, sys, time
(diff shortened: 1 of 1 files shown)


mambocab:
obviated by #293.

PR #294, opened 2015-05-22
skip copy tests pre-2.1

mambocab:
The copy tests are broken pre-2.1, for example [here](http://cassci.datastax.com/job/cassandra-2.0_novnode_dtest/24/testReport/junit/cqlsh_copy_tests/CqlshCopyTest/test_undefined_as_null_indicator/). Until I figure out formatting for 2.0, we should skip these tests.

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -26,6 +26,7 @@ DEFAULT_FLOAT_PRECISION = 5  # magic number copied from cqlsh script
 
 
 @canReuseCluster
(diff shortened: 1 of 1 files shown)


mambocab:
I should note, we don't see this failure on 2.1+: http://cassci.datastax.com/job/cassandra-2.0_novnode_dtest/24/testReport/junit/cqlsh_copy_tests/CqlshCopyTest/test_null_as_null_indicator/

mambocab:
Still failing. I'm confident this is the right thing to do for right now, so I'm merging.

PR #296, opened 2015-05-26
Cassandra-5839

mambocab:
Some basic tests for CASSANDRA-5839 (27596a8) along with a few improvements to `repair_test.py`:

- some style improvements (5fec1d4, 4258a8f),
- removal of [mutable default argument](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) footguns (de5ef7e).

I do want the reviewer to sanity-check [this line](https://github.com/riptano/cassandra-dtest/blob/4258a8f832b7c0b2cc502db5196d2800fbd8e22f/repair_test.py#L310) -- I'm pretty sure we can ignore that error. It's generated [here](https://github.com/apache/cassandra/blob/d96a02a12e20bc2e0b096904ebe7a1d68532d2f3/src/java/org/apache/cassandra/db/lifecycle/LifecycleTransaction.java#L258) and appears to be a guard against invalid updates, not a fatal error.

### DIFF
--- a/repair_test.py
+++ b/repair_test.py
@@ -10,8 +10,13 @@ from tools import insert_c1c2, no_vnodes, query_c1c2, since
 
 class TestRepair(Tester):
 
-    def check_rows_on_node(self, node_to_check, rows, found=[], missings=[], restart=True):
+    def check_rows_on_node(self, node_to_check, rows, found=None, missings=None, restart=True):
+        if found is None:
+            found = []
+        if missings is None:
+            missings = []
         stopped_nodes = []
+
         for node in self.cluster.nodes.values():
             if node.is_running() and node is not node_to_check:
                 stopped_nodes.append(node)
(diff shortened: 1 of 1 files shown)


repair_test.py
```diff
@@ -216,3 +226,104 @@ def _empty_vs_gcable_no_repair(self, sequential):
...
         assert len(l) > 0, "Non GC-able data should be repaired"
+
+
+RepairTableContents = namedtuple('RepairTableContents',
+                                 ['parent_repair_history', 'repair_history'])
+
+
+@since('2.2')
+class TestRepairDataSystemTable(Tester):
+    '''
```
ptnapoleon: https://www.python.org/dev/peps/pep-0257/

These should be using `"""` not `'''`

repair_test.py
```diff
@@ -216,3 +226,104 @@ def _empty_vs_gcable_no_repair(self, sequential):
...
+    def setUp(self):
+        '''
+        Prepares a cluster for tests of the repair history tables by starting
+        a 5-node cluster, then inserting 5000 values with RF=3.
+        '''
+
+        Tester.setUp(self)
+        self.cluster.populate(5).start(wait_for_binary_proto=True)
+        [self.node1, _, _, _, _] = self.cluster.nodelist()
```
ptnapoleon: You can just use `self.node, = self.cluster.nodelist()` and it looks much nicer.

repair_test.py
mambocab: I... I never knew. Gracious!

mambocab:
I addressed @ptnapoleon's comments and squashed them onto 5b7c54d. Since some of these tests flap due to SSTable GC issues, I've marked the new tests as requiring CASSANDRA-9482. With a +1 for that, I'll merge.

ptnapoleon:
+1

PR #297, opened 2015-05-26
monkeypatches driver for all cqlsh tests

mambocab:
Fixes a failing cqlsh test on trunk by using the same driver monkeypatching the cqlsh does in all cqlsh tests, not just the new copy tests.

### DIFF
--- a/cqlsh_tests/cqlsh_tools.py
+++ b/cqlsh_tests/cqlsh_tools.py
@@ -6,6 +6,8 @@ import time
 
 from nose.tools import assert_items_equal
 
+import cassandra
+
 
 class DummyColorMap(object):
     def __getitem__(self, *args):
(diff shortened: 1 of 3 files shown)


cqlsh_tests/cqlsh_tools.py
```diff
@@ -58,3 +60,24 @@ def write_rows_to_csv(filename, data):
         writer = csv.writer(csvfile)
         for row in data:
             writer.writerow(row)
+
+
+def monkeypatch_driver(obj):
+    '''
```
ptnapoleon: See comment on #296

ptnapoleon:
+1 other than the docstrings

mambocab:
docstring quotes fixed. merging. Thanks for the review!

PR #298, opened 2015-05-26
add formatting for 2.1 compatibility

mambocab:
This adds a format string, like in cqlsh itself, for `datetime`s in the cqlsh copy tests. This should fix [this test](http://cassci.datastax.com/job/cassandra-2.1_dtest/106/testReport/cqlsh_copy_tests/CqlshCopyTest/test_all_datatypes_write/), which is broken on 2.1.


ptnapoleon:
+1

PR #299, opened 2015-05-26
Forcefully terminate running cassandra instances in Tester.setUp on W…

josh-mckenzie:
…indows

Interim solution to get tests to run through to completion on Windows. Combined with https://github.com/pcmanus/ccm/pull/289, we should be able to get a clear view of whether specific tests are hanging or if it's inconsistent.


ptnapoleon:
I'm +1. @mambocab, any objections?

mambocab:
LGTM!

PR #300, opened 2015-05-27
fix deserialize problems

mambocab:
Don't attach the cached functions to a class -- that makes it so it's treated as a method, so deserialize can't be called as a function afterwards. This puts them in a dictionary instead.

Causing failures on cassci like [this one](http://cassci.datastax.com/view/All_Jobs/job/trunk_novnode_dtest/65/testReport/cql_tests/TestCQL/select_with_alias_test/). I repro'd the test bug locally with this command:

```
$ CASSANDRA_VERSION=git:trunk nosetests cqlsh_tests/cqlsh_copy_tests.py:CqlshCopyTest.test_all_datatypes_round_trip cql_tests.py:TestCQL.empty_blob_test --verbosity=3
```

Fails on master, succeeds with this PR.

### DIFF
--- a/cqlsh_tests/cqlsh_tools.py
+++ b/cqlsh_tests/cqlsh_tools.py
@@ -62,22 +62,25 @@ def write_rows_to_csv(filename, data):
             writer.writerow(row)
 
 
-def monkeypatch_driver(obj):
+def monkeypatch_driver():
     """
     Monkeypatches the `cassandra` driver module in the same way
-    that clqsh does. Stores the original values of monkeypatched names on
-    `obj`.
+    that clqsh does. Returns a dictionary containing the original values of
+    the monkeypatched names.
     """
-    obj._cached_deserialize = cassandra.cqltypes.BytesType.deserialize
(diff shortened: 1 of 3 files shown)


cqlsh_tests/cqlsh_tools.py
```diff
@@ -62,22 +62,25 @@ def write_rows_to_csv(filename, data):
             writer.writerow(row)
 
 
-def monkeypatch_driver(obj):
+def monkeypatch_driver():
     """
     Monkeypatches the `cassandra` driver module in the same way
     that clqsh does. Stores the original values of monkeypatched names on
     `obj`.
```
ptnapoleon: Is this comment still correct?

mambocab: Nope. Good catch.

mambocab:
Thanks for the review, @ptnapoleon. I've addressed your comment on the docstring. Anything else?

ptnapoleon:
It all looks good to me, and no more confusing than it was before, so +1.

PR #301, opened 2015-05-27
Add test for cassandra-7886

aboudreault:
Ensure nodes return an error message in case of TombstoneOverwhelmingExceptions rather than dropping the request. A drop makes the coordinator waits for the specified  read_request_timeout_in_ms.


ptnapoleon:
The code all looks good to me, +1

PR #302, opened 2015-05-28
Tests for CASSANDRA-8502

thobbs:


ptnapoleon:
Why put this here and not in paging_tests.py?

ptnapoleon:
Other than moving this to paging_tests.py, can you change the `@since` annotation to read `2.0.6`, since that is when static columns were introduced?

Other than that, +1. I'll add the doxygen tags myself after merge.

thobbs:
Okay, should be good to go.

ptnapoleon:
+1.

PR #303, opened 2015-05-28
Ensure NEW_NODE notifications aren't errantly collected

thobbs:
Along with https://github.com/pcmanus/ccm/pull/291, this should
make sure that we don't collect NEW_NODE events from the cluster
starting up.  Sadly, I don't see any simple alternatives to adding the sleep()
line.


ptnapoleon:
+1

PR #304, opened 2015-05-29
CASSANDRA-9492: Using stable ordering for all JDKs

carlyeks:
This goes with CASSANDRA-9492, since the counter column is first alphabetically, we'll use that as the default validator


ptnapoleon:
+1. I'll merge this after 9492 is committed.

PR #305, opened 2015-05-29
simplifies repair table tests and changes require

mambocab:
CASSANDRA-9842 was fixed, but the check for the parent repair history table is still failing, so we skip just that test and let the other tests run.

This also simplifies the tests by removing a parameterized test and the resulting indirection.


ptnapoleon:
+1

PR #306, opened 2015-05-29
dtest for 8989 & 9502

snazy:
dtest for "Reading from table which contains collection type using token function and with CL > ONE causes overwhelming writes to replicas" (CASSANDRA-8989)

### DIFF
--- a/read_repair_test.py
+++ b/read_repair_test.py
@@ -1,0 +1,79 @@
+import time
+
+from cassandra import ConsistencyLevel
+from cassandra.query import SimpleStatement
+
+from dtest import Tester, PRINT_DEBUG
+from tools import since
+
+
+@since('2.0')
+class TestRepairDataSystemTable(Tester):
+    """
+    @jira_ticket CASSANDRA-8989
+    @jira_ticket CASSANDRA-9502
+
+    Range-slice queries with CL>ONE do unnecessary read-repairs.
+    Reading from table which contains collection type using token function and with CL > ONE causes overwhelming writes to replicas.
+
+
+    It's possible to check the behavior with tracing - pattern matching in system_traces.events.activity
+    """
+    def setUp(self):
+        Tester.setUp(self)
+        self.cluster.populate(3).start(wait_for_binary_proto=True)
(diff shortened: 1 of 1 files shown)


read_repair_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+from cassandra import ConsistencyLevel
+from cassandra.query import SimpleStatement
+
+from dtest import Tester, PRINT_DEBUG
+from tools import since
+
+
+@since('2.0')
+#@require('CASSANDRA-9502')
```
ptnapoleon: Remove the commented out line, please.

read_repair_test.py
```diff
@@ -0,0 +1,80 @@
...
+        future = cursor1.execute_async(query, trace=True)
+        future.result()
+        trace = future.get_query_trace(max_wait=120)
+        self.pprint_trace(trace)
+        for trace_event in trace.events:
+            # Step 1, find coordinator node:
+            activity = trace_event.description
+            assert activity != "Appending to commitlog"
+            assert activity != "Adding to cf memtable"
+            assert activity != "Acquiring switchLock read lock"
```
ptnapoleon: Could you make these perform substring searches, instead of using `!=`?

read_repair_test.py
```diff
@@ -0,0 +1,80 @@
...
+from dtest import Tester, PRINT_DEBUG
+from tools import since
+
+
+@since('2.0')
+#@require('CASSANDRA-9502')
+class TestRepairDataSystemTable(Tester):
+    """
+    @jira_ticket CASSANDRA-8989
+    @jira_ticket CASSANDRA-9502
```
ptnapoleon: Thanks for the doxygen markup!

snazy: copy + paste ;)

read_repair_test.py
snazy: is it ok? (python question ;) )

ptnapoleon: Is the substring search okay? The best way to do it would be `assert "Appending to commitlog" not in activity`

ptnapoleon:
Thanks. +1

PR #307, opened 2015-06-01
Add multiple dc repair tests

yukim:
For [CASSANDRA-9521](https://issues.apache.org/jira/browse/CASSANDRA-9521) and part of [CASSANDRA-9109](https://issues.apache.org/jira/browse/CASSANDRA-9109).

### DIFF
--- a/repair_test.py
+++ b/repair_test.py
@@ -227,6 +227,92 @@ class TestRepair(Tester):
         l = node2.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync for cf2")
         assert len(l) > 0, "Non GC-able data should be repaired"
 
+    def local_dc_repair_test(self):
+        cluster = self._setup_multi_dc()
+        node1 = cluster.nodes["node1"]
+        node2 = cluster.nodes["node2"]
+
+        debug("starting repair...")
+        opts = ["-local"]
+        opts += self._repair_options(ks="ks")
+        node1.repair(opts)
+
+        # Verify that only nodes in dc1 is involved in repair
+        l = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
+        assert len(l) == 1, "Lines matching: %d" % len(l)
+        line, m = l[0]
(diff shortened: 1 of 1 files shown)


repair_test.py
```diff
@@ -227,6 +227,92 @@ def _empty_vs_gcable_no_repair(self, sequential):
...
+
+    def _setup_multi_dc(self):
+        """
+        Sets up 3 DCs (2 nodes in 'dc1', and one each in 'dc2' and 'dc3').
+        After set up, node2 in dc1 lacks some data and needs to be repaired.
+        """
+        cluster = self.cluster
+
+        # Disable hinted handoff and set batch commit log so this doesn't
+        # interfer with the test (this must be after the populate)
```
ptnapoleon: This comment says the `set_configuration_options` call must occur after `cluster.populate`, but that isn't happening here. Which is wrong, the code or the comment?

repair_test.py
```diff
@@ -227,6 +227,92 @@ def _empty_vs_gcable_no_repair(self, sequential):
...
+        node2 = cluster.nodes["node2"]
+        node3 = cluster.nodes["node3"]
+
+        debug("starting repair...")
+        opts = ["-dc", "dc1,dc2"]
+        opts += self._repair_options(ks="ks")
+        node1.repair(opts)
+
+        # Verify that only nodes in dc1 is involved in repair
+        l = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
```
ptnapoleon: Can you use a more descriptive variable name than `l`?

repair_test.py
```diff
@@ -227,6 +227,92 @@ def _empty_vs_gcable_no_repair(self, sequential):
...
+        assert len(l) == 2, "Lines matching: " + str([elt[0] for elt in l])
```
ptnapoleon: Can you use `self.assertEqual` rather than the python assert?

yukim:
Thanks for the review. I converted to unittest's assert for whole repair_test, not just for the new ones.

ptnapoleon:
I'm still concerned about the set_configuration_options comment/behavior, but other than that, +1.

ptnapoleon:
Merged. Only the comment is wrong, not the behavior. I'll fix that in a separate commit.

PR #309, opened 2015-06-02
Clean up dtest for output of pyflakes

ptnapoleon:
I ran all of dtest through pyflakes. It found a lot of unused imports, some unused variables worth removing, and even a few unnecessary functions. I also snuck in a stylistic change in expanding tuples.

### DIFF
--- a/auth_roles_test.py
+++ b/auth_roles_test.py
@@ -1,11 +1,11 @@
-import time
+import time, re
 
-from cassandra import AuthenticationFailed, Unauthorized
+from cassandra import AuthenticationFailed, Unauthorized, InvalidRequest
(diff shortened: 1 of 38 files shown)


mambocab:
Looks amazing. As long as pyflakes doesn't complain about any imports, merge away.

knifewine:
You can do a kinda poor man's check by doing something like:

for file in *.py; do echo $file; python $file; done

to make sure the python interpreter doesn't print any errors. You will get some output from the upgrade tests though.... :)

ptnapoleon:
@knifewine Great advice hatch. Everything seems to be in order then.

PR #311, opened 2015-06-04
CASSANDRA-7066, changes for tmp files in offline tools 

stef1927:
I had to change offline tools to cope with the new way to discover temporary files or else it will fail after CASSANDRA-7066 is committed. It will work even if 7066 is not yet committed.

### DIFF
--- a/sstablelist_test.py
+++ b/sstablelist_test.py
@@ -1,6 +1,7 @@
 from dtest import Tester, debug
 from tools import since, require, InterruptCompaction
 from ccmlib import common
+from ccmlib.node import NodetoolError
 
 import subprocess
 import glob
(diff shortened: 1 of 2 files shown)


sstablelist_test.py
```diff
@@ -51,7 +52,11 @@ def listfiles_onabortedcompaction_test(self):
         t = InterruptCompaction(node, TableName)
         t.start()
 
-        node.compact()
+        try:
+            node.compact()
```
ptnapoleon: If this doesn't fail, do you think it's worth either throwing an error, or logging something with `debug(message)`?

ptnapoleon:
Other than potentially making the one suggested change, +1.

sstablelist_test.py
stef1927: If it doesn't fail then there won't be any tmp files and the assertion below will fail. However it may add extra clarity so I will add something, thanks.

stef1927:
Thanks, committed with suggested change.

PR #312, opened 2015-06-04
Convert CRLF to LF in cqlsh_print on Windows.

josh-mckenzie:
Python converts to operating-system specific line-endings on any stream
write or print call which causes failures when coupled with doctest's
<BLANKLINE> replacement logic.


ptnapoleon:
+1

PR #314, opened 2015-06-08
adds cql test for NPE in 'WITH WITH' queries

mambocab:
This tests the bug described in [CASSANDRA-9565](https://issues.apache.org/jira/browse/CASSANDRA-9565).

It might be more appropriate to test this as a unit test, in which case we shouldn't merge this. Thoughts?

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -5295,3 +5295,19 @@ class TestCQL(Tester):
 
         cursor = self.patient_cql_connection(self.cluster.nodelist()[0])
         assert_all(cursor, "SELECT k FROM ks.test WHERE v = 0", [[0]])
+
+    @require(9565)
+    def double_with_npe_test(self):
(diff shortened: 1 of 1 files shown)


ptnapoleon:
-1. CASSANDRA-9160 is almost done. Just put this into CqlTester as a unit test. Alternatively, you could leave it here, but restrict it to only run on 2.0.x. I'm +1 for that.

mambocab:
> you could leave it here, but restrict it to only run on 2.0.x. I'm +1 for that.

Is that the plan for the whole `cql_tests` suite? If so, then let's merge this; otherwise, it'd be weird to only have one test work like that.

ptnapoleon:
It's either that, or deleting all of cql_tests and losing 2.0 coverage. Which I'm -1 on. So that's the plan as I understand it. +1 to merge as is then.

PR #315, opened 2015-06-08
Cassandra-5322 Making dtests more granular

steveandwang:
Updated environment variables DEBUG and TRACE such that they also accept class names along with true/false/yes/no, By inputting class name(s) (separated with a colon), you can set only those classes to have that level of logging.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -152,6 +152,28 @@ class Tester(TestCase):
 
         return cluster
 
+    def var_debug(self, cluster):
+        general_DEBUG = os.environ.get('DEBUG', 'no').lower() not in ('no', 'false', 'yes', 'true')
+        classes_to_debug = []
+        if general_DEBUG:
+            classes_to_debug = os.environ.get('DEBUG').split(":")
+            cluster.set_log_level('DEBUG', None if len(classes_to_debug) == 0 else classes_to_debug)
+
+    def var_trace(self, cluster):
+        general_TRACE = os.environ.get('TRACE', 'no').lower() not in ('no', 'false', 'yes', 'true')
+        classes_to_trace = []
+        if general_TRACE:
+            classes_to_trace = os.environ.get('TRACE').split(":")
+            cluster.set_log_level('TRACE', None if len(classes_to_trace) == 0 else classes_to_trace)
+
+    def modify_log(self, cluster):
+        if DEBUG:
+            cluster.set_log_level("DEBUG")
+        if TRACE:
+            cluster.set_log_level("TRACE")
+        self.var_debug(cluster)
+        self.var_trace(cluster)
+
     def _cleanup_cluster(self):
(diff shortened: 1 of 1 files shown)


dtest.py
```diff
@@ -152,6 +155,22 @@ def _get_cluster(self, name='test'):
 
         return cluster
 
+    def var_debug(self, cluster):
+        not_specific_DEBUG = os.environ.get('DEBUG', 'no').lower() in ('no', 'false', 'yes', 'true')
+        classes_to_debug = []
+        if not not_specific_DEBUG:
```
ptnapoleon: It would be preferable to rename the variable to `specific_DEBUG`, and perform the opposite boolean check. A double negative is undesirable.

dtest.py
```diff
@@ -141,6 +141,9 @@ def _get_cluster(self, name='test'):
         else:
             cluster = Cluster(self.test_path, name, cassandra_dir=cdir)
 
+        self.var_debug(cluster)
+        self.var_trace(cluster)
```
ptnapoleon: Could you move these calls down to where the existing set_log_level calls are? And then maybe break out the four calls into a separate function?

dtest.py
```diff
@@ -152,6 +155,22 @@ def _get_cluster(self, name='test'):
...
+            for cass_class in classes_to_debug:
+                cluster.set_log_level('DEBUG', cass_class)
+
+    def var_trace(self, cluster):
+        not_specific_TRACE = os.environ.get('TRACE', 'no').lower() in ('no', 'false', 'yes', 'true')
+        classes_to_trace = []
+        if not not_specific_TRACE:
+            classes_to_trace = os.environ.get('TRACE').split(":")
+            for cass_class in classes_to_trace:
+                cluster.set_log_level('TRACE', cass_class)
```
ptnapoleon: Does this really work in an append fashion? I worry it would overwrite the previous calls.

ptnapoleon:
Make sure you squash the commits as well

dtest.py
```diff
@@ -152,6 +152,28 @@ def _get_cluster(self, name='test'):
 
         return cluster
 
+    def var_debug(self, cluster):
+        general_DEBUG = os.environ.get('DEBUG', 'no').lower() not in ('no', 'false', 'yes', 'true')
+        classes_to_debug = []
+        if general_DEBUG:
+            classes_to_debug = os.environ.get('DEBUG').split(":")
+            cluster.set_log_level('DEBUG', classes_to_debug if len(classes_to_debug) == 0 else None)
```
ptnapoleon: Isn't this ternary operator wrong? Don't you want `classes_to_debug if len(classes_to_debug) > 0 else None`?

ptnapoleon:
+1, thanks!

knifewine:
nice job, @steveandwang !

PR #317, opened 2015-06-10
Add deprecated repair JMX API tests

yukim:
Test for [CASSANDRA-9570](https://issues.apache.org/jira/browse/CASSANDRA-9570).
In the future, when ```@Deprecated``` repair methods are removed, this test will be obsolete.

### DIFF
--- a/deprecated_repair_test.py
+++ b/deprecated_repair_test.py
@@ -1,0 +1,185 @@
+import time
+
+from cassandra import ConsistencyLevel
+from cassandra.query import SimpleStatement
+
+from dtest import Tester, debug
+from tools import since, insert_c1c2, query_c1c2
+from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
+
+@since("2.2")
+class TestDeprecatedRepairAPI(Tester):
+    """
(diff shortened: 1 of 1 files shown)


deprecated_repair_test.py
```diff
@@ -0,0 +1,185 @@
...
+from cassandra.query import SimpleStatement
+
+from dtest import Tester, debug
+from tools import since, insert_c1c2, query_c1c2
+from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
+
+@since("2.2")
+class TestDeprecatedRepairAPI(Tester):
+    """
+    Test if depreated repair JMX API runs with expected parameters
```
ptnapoleon: Can you add a line `@jira_ticket CASSANDRA-9570` in this docstring?

ptnapoleon:
Do these deprecated methods only exist in 2.2? Why are we not testing them in previous versions?

yukim:
These are deprecated in 2.2 and no longer used from ```nodetool repair```. So 2.1 and before are (kind of) tested through repair test.
From 2.2, ```nodetool``` uses unified ```repairAsync``` method.

ptnapoleon:
+1, thanks.

PR #319, opened 2015-06-10
Add rebuild test

yukim:
```nodetool rebuild``` test seems missing.
Added simple rebuild test that traces the steps of adding data center and rebuild from the other.
Also added verification for my fix for [CASSANDRA-9119](https://issues.apache.org/jira/browse/CASSANDRA-9119).

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -1,0 +1,93 @@
+import time
+from dtest import Tester, debug
+from tools import insert_c1c2, query_c1c2, new_node
+from ccmlib.node import Node, NodetoolError
+from cassandra import ConsistencyLevel
+from threading import Thread
+
+
+class TestRebuild(Tester):
+
+    def __init__(self, *args, **kwargs):
+        kwargs['cluster_options'] = {'start_rpc': 'true'}
+        # Ignore these log patterns:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rebuild_test.py [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+        node2 = cluster.create_node('node2', False,
+                                    ('127.0.0.2', 9160),
+                                    ('127.0.0.2', 7000),
+                                    '7200', '2001', None,
+                                    binary_interface=('127.0.0.2', 9042))
+        cluster.add(node2, False, data_center='dc2')
+        node2.start(wait_other_notice=True, wait_for_binary_proto=True)
+
+        # wait for snitch to reload
+        time.sleep(60)
```
ptnapoleon: Is there no better way of waiting for the snitch than this sleep? Maybe we can parse a log, or query something with jmx?

yukim: It is the way cassandra's PropertyFileSnitch does without log.
I couldn't figure out pre-configure cassandra-topology.properties without adding nodes using ccm.

ptnapoleon: Well than, that will have to do

ptnapoleon:
+1

PR #320, opened 2015-06-10
Use ccm run cqlsh

mambocab:
7681edb is the meat of this PR. `TestCqlsh` uses a custom `run_cqlsh` method that can return the stdout and stderr output from the cqlsh command. Since https://github.com/pcmanus/ccm/pull/264 was merged, this is no longer necessary. This commit removes the function and replaces all instances of its being called with ccmlib.Node.run_cqlsh.

The other commits are more cleanup-y cleanup: they organize imports with `isort`, remove `TestCqlsh.__init__` (which had no effect), correct some indentation that was too deep, and :fireworks: add Doxygen docstrings to the test methods in `TestCqlsh` :fireworks:.

The tests still pass locally on `trunk` and `cassandra-2.{2,1,0}`

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -17,10 +17,6 @@ from tools import create_c1c2_table, insert_c1c2, rows_to_list, since
 
 
 class TestCqlsh(Tester):
-
-    def __init__(self, *args, **kwargs):
-        Tester.__init__(self, *args, **kwargs)
-
     @classmethod
     def setUpClass(cls):
         cls._cached_driver_methods = monkeypatch_driver()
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Can you check if/how much this works on windows?

ptnapoleon:
Other than that, I'm +1

mambocab:
I don't expect problems, but I've just asked for a Windows license to test locally. I'll let you know when I've tested with it.

mambocab:
Something about the implementation of `run_cqlsh` in here is preventing some tests from failing on Windows (specifically `tracing_from_system_traces_test` and `test_eat_glass`). I don't know why. Until ccm's `run_cqlsh` works properly on Windows, let's not merge this. Shall I close the PR until then, or can we leave it open?

ptnapoleon:
Better to leave it open I think, so we feel bad for not fixing it.

knifewine:
not sure this relates at all, but Josh made a change for the json tests that had to deal with windows line endings. I mention that because I used some code from cqlsh_tests.py when I did the json stuff.
https://github.com/riptano/cassandra-dtest/pull/312/files

mambocab:
I'm gonna go ahead and close this. We have an internal ticket to work on this already.

PR #321, opened 2015-06-12
Migration of CQL tests from dtests to unit tests, CASSANDRA-9160

stef1927:
These are the changes for CASSANDRA-9160:

- cql_tests.py now contains only the tests which could not be implemented in java, and it is marked as @since(2.1) because the migration is only effective in 2.1 onward. It also contains one basic test per CQL statement, to exercise the code in StorageProxy that is not covered by CQLTester.

- deprecated_cql_tests.py contails all tests which did not have a @since(>= 2.1) and is marked to run only for version up to 2.0 using the @since max_version property.

Once review is complete I can squash before merging.


ptnapoleon:
`cql3_rename_thrift_test` is marked `since('2.0')` here, but the class it is in is marked `since('2.1')`. I assume the stricter restriction will win.

ptnapoleon:
I'm +1 on this strategy of moving all the existing tests to `deprecated_cql_tests` and restricting it to < 2.1, as well as keeping a cql_tests just for 2.1+. The new 2.1 cql_tests all look good, so I'm +1 overall.

stef1927:
Thanks for your review. I've removed the redundant @since tags in cql_tests.py. We should obviously wait for CASSANDRA-9160 to be committed before merging this.

ptnapoleon:
9160 still hasn't been committed, and this PR now needs rebased.

stef1927:
Squashed and rebased.

In prepared_statement_invalidation_test(), passing InvalidRequestException at line 362 was failing the test so I left InvalidRequest as a parameter, as it was originally.

ptnapoleon:
I'll merge this as soon as Josh finishes review of 9160

stef1927:
Josh has finished 9160, merging.

PR #322, opened 2015-06-12
adds compact storage option to create_cf

mambocab:
This adds a `compact_storage` kwarg to `Tester.create_cf`. If truthy, it creates the table `WITH COMPACT STORAGE`. This is helping me with comparing the old row format to the new format that will be introduced in 3.0.

Since it's a kwarg at the end of the parameter list, it doesn't affect existing calls.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -393,7 +393,7 @@ class Tester(TestCase):
         session.execute('USE %s' % name)
 
     # We default to UTF8Type because it's simpler to use in tests
-    def create_cf(self, session, name, key_type="varchar", speculative_retry=None, read_repair=None, compression=None, gc_grace=None, columns=None, validation="UTF8Type"):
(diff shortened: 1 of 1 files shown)


dtest.py
```diff
@@ -393,7 +393,7 @@ def create_ks(self, session, name, rf):
         session.execute('USE %s' % name)
 
     # We default to UTF8Type because it's simpler to use in tests
-    def create_cf(self, session, name, key_type="varchar", speculative_retry=None, read_repair=None, compression=None, gc_grace=None, columns=None, validation="UTF8Type"):
+    def create_cf(self, session, name, key_type="varchar", speculative_retry=None, read_repair=None, compression=None, gc_grace=None, columns=None, validation="UTF8Type", compact_storage=False):
```
ptnapoleon: Any chance you want to split this header across two lines, since it's getting long?

ptnapoleon:
+1, with or w/o my suggested change.

mambocab:
Good idea. Thanks!

PR #323, opened 2015-06-12
require and clean up rebuild test

mambocab:
[9119](https://issues.apache.org/jira/browse/CASSANDRA-9119) hasn't been merged yet, so this test is failing on cassci; the first commit fixes that.

The next two commits are import and whitespace cleanup.

The final commit removes `allow_log_errors` from the class in `rebuild_test`; it still passes on the 9119 branch without it (at least locally), so we should catch errors if we can.

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -1,9 +1,11 @@
 import time
-from dtest import Tester, debug
-from tools import insert_c1c2, query_c1c2, new_node, require
-from ccmlib.node import Node, NodetoolError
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I'm worried the allow_log_errors will break on expected log errors that might only show up intermittently or on slower machines like cassci. But if that happens, we can investigate those when they show up. +1.

mambocab:
+1 to your thoughts on the errors. I'd rather deal with false positives than false negatives.

PR #324, opened 2015-06-12
adds metadata to thrift_hsha_test tests

mambocab:
Quick little improvement -- adds some Doxygen metadata to the thrift HSHA tests.

### DIFF
--- a/thrift_hsha_test.py
+++ b/thrift_hsha_test.py
@@ -23,7 +23,11 @@ class ThriftHSHATest(Tester):
 
     @unittest.skipIf(sys.platform == "win32", 'Could not be executed on Windows')
     def test_closing_connections(self):
-        """Test CASSANDRA-6546 - do connections get closed when disabling / renabling thrift service?"""
+        """
+        @jira_ticket CASSANDRA-6546
+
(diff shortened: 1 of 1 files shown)


ptnapoleon:
You can feel 100% free to just push these kind of commits. +1 though.

mambocab:
> You can feel 100% free to just push these kind of commits.

Personally, I like having reviews even for the smallest commits -- by definition, I don't know when I'm going to screw up. That said, if it's just noise for you, I can stop; just let me know

I can push this up if you'd prefer I not generate merge commit noise. Do you care about that?

ptnapoleon:
I don't mind merge commit noise, and I don't object to PRs, just wanted to make sure you knew you could "ninja" things like this if you wanted.

But since you're explicitly asking for review on this, could we standardize the value for `@jira_ticket`? In some places we use `CASSANDRA-XYZ` and in others, we use `XYZ`. I'd prefer to use the former, since the html docs we generate with Doxygen will be colocated with docs from other softwares who use other JIRA's. Additionally, we might want to link python driver tickets (unlikely, but possible).

mambocab:
Awesome, thanks. Force pushed with `@jira_ticket CASSANDRA-XYZ`, merging.

PR #325, opened 2015-06-12
skip flapping deletion tests on pre-2.1

mambocab:
This test flaps with LCS on 2.0, so let's skip it until it's fixed. There's a Jira [here](https://issues.apache.org/jira/browse/CASSANDRA-9595).


mambocab:
Closing in favor of #327.

PR #326, opened 2015-06-15
s/@jira_ticket N/@jira_ticket CASSANRA-N/g

mambocab:
Updates CONTRIBUTING to reflect this style choice.

Follow up to feedback on #324. We use CASSANDRA-N instead of just N because some dtests will be related to tickets in other projects, such as the Python driver.

You can verify that there aren't any more instances of `@jira_ticket N` with

```
grep -r 'jira_ticket[ \d]*$'
```


ptnapoleon:
+1

PR #327, opened 2015-06-15
Major compaction lcs

mambocab:
As discussed [here](https://issues.apache.org/jira/browse/CASSANDRA-9595), major compaction doesn't such for tables using LCS before 2.2.

This PR skips tests in `compaction_test.py` that use `nodetool compact` when the version is less than 2.2 and the strategy being tested is LCS.

This PR also includes some whitespace fixes. The changes are in separate commits for easier review.

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -4,9 +4,10 @@ from assertions import assert_none, assert_one
 import tempfile
 import os
 
+
 class TestCompaction(Tester):
 
-    __test__= False
+    __test__ = False
 
     def __init__(self, *args, **kwargs):
         kwargs['cluster_options'] = {'start_rpc': 'true'}
@@ -22,7 +23,6 @@ class TestCompaction(Tester):
 
         cursor = self.patient_cql_connection(node1)
         self.create_ks(cursor, 'ks', 1)
-
 
         cursor.execute("create table ks.cf (key int PRIMARY KEY, val int) with compaction = {'class':'" + self.strategy + "'} and gc_grace_seconds = 30;")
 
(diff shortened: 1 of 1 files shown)


compaction_test.py
```diff
@@ -4,9 +4,10 @@
 import tempfile
 import os
 
+
```
ptnapoleon: Why?

mambocab: [PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines): "Surround top-level function and class definitions with two blank lines."

It's a habit. I'll disable it in my linter if we don't like this part of PEP8.

aboudreault: I think it's good to follow pep8, at least for new stuff. I often do that kind of fixes too.

ptnapoleon: Oh then I'm +1. I didn't realize two blank lines was the rule. All about PEP-8

ptnapoleon:
Do these tests definitely RELY on major compaction working for LCS, or do they just use it?

mambocab:
I left out a word above: s/doesn't as such/doesn't exist as such/.

> Do these tests definitely RELY on major compaction working for LCS, or do they just use it?

They use it, but don't seem to rely on it -- they pass on 2.0 and 2.1, with the exception of `sstable_deletion_test`, which flaps on 2.0.

If these tests have some purpose other than to fail when compaction is performed incorrectly, I think they should be renamed. Maybe they're just compaction _strategy_ tests that make sure tables created with the different compaction strategies behave correctly.

If the purpose is to test the correctness of compactions, then I would shore up these tests so they check that compaction was actually performed -- maybe by watching logs or something similar.

ptnapoleon:
I'd rather first investigate which of these can be rewritten to work around the lack of major compactions. `compaction_throughput_test` definitely can. `data_size_test` can as well. Only `sstable_deletion_test` sticks out to me as possibly an outlier.

ptnapoleon:
Should we close this in favor of #330?

mambocab:
Yes.

PR #328, opened 2015-06-16
New consistency tests introduced by CASSANDRA-9102

stef1927:
There are the tests for [CASSANDRA-9102] (https://issues.apache.org/jira/browse/CASSANDRA-9102) : Consistency levels such as non-local quorum need better tests.

### DIFF
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -38,13 +38,13 @@ class TestHelper(Tester):
         }[cl]
 
     def _is_local(self, cl):
-        return cl == ConsistencyLevel.LOCAL_QUORUM or \
-               cl == ConsistencyLevel.LOCAL_ONE or \
-               cl == ConsistencyLevel.LOCAL_SERIAL
+        return (cl == ConsistencyLevel.LOCAL_QUORUM or
+                cl == ConsistencyLevel.LOCAL_ONE or
+                cl == ConsistencyLevel.LOCAL_SERIAL)
 
     def _is_conditional(self, cl):
-        return cl == ConsistencyLevel.SERIAL or \
-               cl == ConsistencyLevel.LOCAL_SERIAL
+        return (cl == ConsistencyLevel.SERIAL or
+                cl == ConsistencyLevel.LOCAL_SERIAL)
 
     def _required_nodes(self, cl, rf_factors, dc):
         """
@@ -66,7 +66,7 @@ class TestHelper(Tester):
             ConsistencyLevel.LOCAL_ONE : 1,
         }[cl]
 
-    def _should_suceed(self, cl, rf_factors, num_nodes_alive, current):
+    def _should_succeed(self, cl, rf_factors, num_nodes_alive, current):
         """
         Return true if the read or write operation should succeed based on
         the consistency level requested, the replication factors and the
@@ -78,7 +78,7 @@ class TestHelper(Tester):
             for i in xrange(0, len(rf_factors)):
                 if num_nodes_alive[i] < self._required_nodes(cl, rf_factors, i):
                     return False
-                return True
+            return True
         else:
             return sum(num_nodes_alive) >= self._required_nodes(cl, rf_factors, current)
 
@@ -91,7 +91,7 @@ class TestHelper(Tester):
         cluster.populate(nodes).start(wait_for_binary_proto=True, wait_other_notice=True)
 
         self.ksname = 'mytestks'
-        session = self.patient_cql_connection(cluster.nodelist()[0])
+        session = self.patient_exclusive_cql_connection(cluster.nodelist()[0])
 
         self.create_ks(session, self.ksname, rf)
         self.create_tables(session)
@@ -100,15 +100,7 @@ class TestHelper(Tester):
             self.sessions = []
             self.sessions.append(session)
             for node in cluster.nodelist()[1:]:
-                self.sessions.append(self.patient_cql_connection(node, self.ksname))
-
-    def _restart_cluster(self):
-        cluster = self.cluster
-
-        cluster.start(wait_for_binary_proto=True, wait_other_notice=True)
-
-        session = self.patient_cql_connection(cluster.nodelist()[0], self.ksname)
-        self.truncate_tables(session)
+                self.sessions.append(self.patient_exclusive_cql_connection(node, self.ksname))
 
     def create_tables(self, session):
         self.create_users_table(session)
@@ -178,9 +170,13 @@ class TestHelper(Tester):
         return ret
 
     def read_counter(self, session, id, consistency):
+        """
+        Return the current counter value. If we find no value we return zero
+        because after the next update the counter will become one.
+        """
         statement = SimpleStatement("SELECT c from counters WHERE id = %d" % (id,), consistency_level=consistency)
         res = rows_to_list(session.execute(statement))
-        return res[0][0] if len(res) > 0 else 0
+        return res[0][0] if res else 0
 
 class TestAvailability(TestHelper):
     """
@@ -188,7 +184,7 @@ class TestAvailability(TestHelper):
     """
     def _test_simple_strategy(self, combinations):
         """
-        Helper test function for a sinle data center: invoke _test_insert_query_from_node() for each node
+        Helper test function for a single data center: invoke _test_insert_query_from_node() for each node
         and each combination, progressively stopping nodes.
         """
         cluster = self.cluster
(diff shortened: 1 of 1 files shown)


consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+            ConsistencyLevel.SERIAL : 'SERIAL',
+            ConsistencyLevel.LOCAL_SERIAL : 'LOCAL_SERIAL',
+            ConsistencyLevel.LOCAL_ONE : 'LOCAL_ONE',
+        }[cl]
+
+    def _is_local(self, cl):
+        return cl == ConsistencyLevel.LOCAL_QUORUM or \
+               cl == ConsistencyLevel.LOCAL_ONE or \
+               cl == ConsistencyLevel.LOCAL_SERIAL
+
```
aboudreault: return (cl == ConsistencyLevel.LOCAL_QUORUM or
            cl == ConsistencyLevel.LOCAL_ONE or
            cl == ConsistencyLevel.LOCAL_SERIAL)
nit: I'd would prefer this way rather than \

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+    def _is_conditional(self, cl):
+        return cl == ConsistencyLevel.SERIAL or \
+               cl == ConsistencyLevel.LOCAL_SERIAL
+
```
aboudreault: same comment than above

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+        the consistency level requested, the replication factors and the
+        number of nodes alive in each data center.
+        """
+        if self._is_local(cl):
+            return num_nodes_alive[current] >= self._required_nodes(cl, rf_factors, current)
+        elif cl == ConsistencyLevel.EACH_QUORUM:
+            for i in xrange(0, len(rf_factors)):
+                if num_nodes_alive[i] < self._required_nodes(cl, rf_factors, i):
+                    return False
+                return True
```
ptnapoleon: This line is intended one too far, right? We'd be exiting after the first iteration of the for loop

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+        res = rows_to_list(session.execute(statement))
+        return res[0][0] if len(res) > 0 else 0
+
+class TestAvailability(TestHelper):
+    """
+    Test that we can read and write depending on the number of nodes that are alive and the consistency levels.
+    """
+    def _test_simple_strategy(self, combinations):
+        """
+        Helper test function for a sinle data center: invoke _test_insert_query_from_node() for each node
```
ptnapoleon: s/sinle/single

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+                self.log('Testing node %d in dc %d with %s nodes alive' % (n, i, nodes_alive))
+                node = n + sum(nodes[:i])
+                session = self.patient_cql_connection(cluster.nodelist()[node], self.ksname)
+                for combination in combinations:
+                    self._test_insert_query_from_node(session, i, rf_factors, nodes_alive, *combination)
+
+                self.cluster.nodelist()[node].stop()
+                nodes_alive[i] = nodes_alive[i] - 1
+
+    def _test_insert_query_from_node(self, session, dc_idx, rf_factors, num_nodes_alive, write_cl, read_cl, serial_cl=None, check_ret=True):
```
mambocab: This method should provide a little more help in tracking down a failure if one happens. Since this method is called in a loop, e.g. in `test_simple_strategy`, a failure will just look like "FAIL: `test_simple_strategy`", which doesn't specify the particular consistency levels the test failed at. I think a `debug` call that prints the relevant parameters would be enough.

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+            self.sessions.append(session)
+            for node in cluster.nodelist()[1:]:
+                self.sessions.append(self.patient_cql_connection(node, self.ksname))
 
-        #Stop a node and retest
-        self.cluster.nodelist()[2].stop()
-        for n in xrange(0, 100):
-            insert_c1c2(session, n, ConsistencyLevel.QUORUM)
-        assert_unavailable(query_c1c2, session2, 100, ConsistencyLevel.THREE)
+    def _restart_cluster(self):
```
ptnapoleon: I don't see where else this method is called. Is it misnamed? It isn't really restarting a cluster, only booting a previously stopped one.

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+        self.create_ks(session, self.ksname, rf)
+        self.create_tables(session)
 
-    def quorum_three_test(self):
-        session, session2 = self.cl_cl_prepare(ConsistencyLevel.QUORUM, ConsistencyLevel.THREE)
+        if save_sessions:
+            self.sessions = []
+            self.sessions.append(session)
+            for node in cluster.nodelist()[1:]:
+                self.sessions.append(self.patient_cql_connection(node, self.ksname))
```
ptnapoleon: Should these be patient, exclusive connections? Otherwise, they're all essentially equivalent, so I don't see why we would use multiple.

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+            ConsistencyLevel.QUORUM : sum(rf_factors) / 2 + 1,
+            ConsistencyLevel.ALL : sum(rf_factors),
+            ConsistencyLevel.LOCAL_QUORUM : rf_factors[dc] / 2 + 1,
+            ConsistencyLevel.EACH_QUORUM : rf_factors[dc] / 2 + 1,
+            ConsistencyLevel.SERIAL : sum(rf_factors) / 2 + 1,
+            ConsistencyLevel.LOCAL_SERIAL : rf_factors[dc] / 2 + 1,
+            ConsistencyLevel.LOCAL_ONE : 1,
+        }[cl]
+
+    def _should_suceed(self, cl, rf_factors, num_nodes_alive, current):
```
ptnapoleon: s/suceed/succeed

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+        session.execute("""CREATE TABLE users (
+                userid int PRIMARY KEY,
+                firstname text,
+                lastname text,
+                age int
+            ) WITH COMPACT STORAGE""")
+
+    def insert_user(self, session, userid, age, consistency, serial_consistency=None):
+        text = "INSERT INTO users (userid, firstname, lastname, age) VALUES (%d, 'first%d', 'last%d', %d) %s" \
+            % (userid, userid, userid, age, "IF NOT EXISTS" if serial_consistency else "")
```
aboudreault: ***NO IMMEDIATE MODIFICATION REQUIRED***

This is comment for future PRs. Since Python 3 will eventually become default in future OS distributions, we try to use as many as possible Py3 habits in our new code, ie. the string formatting. The formatting use the "".format() function.

https://docs.python.org/3.5/library/string.html#format-string-syntax

***NO IMMEDIATE MODIFICATION REQUIRED***

ptnapoleon: What Alan is saying, is that formatting with `%` here is fine, but in future PRs, if you could use the python3 way, that would be preferable, please.

consistency_test.py
aboudreault: +1. I think adding some more debug to exactly what exactly is tested would be simple and good enough there.

consistency_test.py
```diff
@@ -1,149 +1,601 @@
...
+        expected = [[id, val]] if val else []
+        ret = rows_to_list(res) == expected
+        if check_ret:
+            assert ret, "Got %s from %s, expected %s at %s" % (res, session.cluster.contact_points, expected, self._name(consistency))
+        return ret
+
+    def read_counter(self, session, id, consistency):
+        statement = SimpleStatement("SELECT c from counters WHERE id = %d" % (id,), consistency_level=consistency)
+        res = rows_to_list(session.execute(statement))
+        return res[0][0] if len(res) > 0 else 0
```
mambocab: Nit: It's more idiomatic to write `if res` than `if len(res) > 0`.

I'm not clear on why returning 0 is ok if no result is returned. Maybe add a comment to explain.

mambocab:
Great tests. I agree with Alan and Philip's comments; when the comments are addressed I'll have a look at the changes.

consistency_test.py
stef1927: Got it, thanks!

consistency_test.py
stef1927: We have already this debug line at the beginning of the method:
```python
self.log("Connected to %s for %s/%s/%s" %
            (session.cluster.contact_points, self._name(write_cl), self._name(read_cl), self._name(serial_cl)))
```
We also have debug calls in the loops calling this method. Where exactly would you like to see more information?

stef1927:
Thanks for your comments. I pushed a new commit and I believe I addressed all comments requiring modification, except one for which I asked a question inline above. Please let me know if I missed something.

consistency_test.py
mambocab: You're right, I totally misread that line. This debug line should be plenty, sorry about that!

stef1927: No problem, thanks for confirming! :)

ptnapoleon:
I'm +1. @mambocab or @aboudreault ?

aboudreault:
Yep +1. Thanks @stef1927  for the great work!

mambocab:
+1, thanks!

PR #329, opened 2015-06-16
adds style section to CONTRIBUTING

mambocab:
This is a starting point for the Python 3 `CONTRIBUTING` changes we discussed while reviewing #328. I couldn't think of any changes other than string formatting off the top of my head, but that list should probably get longer at some point.

I've taken the liberty of also mentioning PEP8 and import sorting. [`isort`](https://github.com/timothycrosley/isort) is the tool I used [here](https://github.com/riptano/cassandra-dtest/commit/2f29d9254a692e5954d5140e1dca7fcdb3696edd) and I think it's nice.

Feel free to propose things to add/remove. I tried to describe how we already do things and not to impose anything new.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,15 @@
+## Style
+
+We plan to move to Python 3 in the near future. Where possible, new code should be Python 3-compatible. In particular:
+
+- favor `format` over `%` for formatting strings.
+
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

aboudreault:
Thanks Jim for that addition. +1

PR #330, opened 2015-06-16
Major compaction cleanup

mambocab:
d0f67ce: As discussed in #327, this is a better solution to `TestCompaction_with_LeveledCompactionStrategy.sstable_deletion_test` flapping on 2.0. We still skip that particular test pre-2.2 (eb8c7ab), but the other tests are not skipped as they were in the previous PR. Instead, we use a flush, a major compaction, and log-waiting to block on a compaction.

587f821 should be reviewed on its own. It's justified in its commit message. It may expose a C* bug: compaction throughput runs at about 120-125% of the specified maximum.

The other commits are cleanup.

Requesting review from @shawnkumar since he wrote the tests and from @ptnapoleon since he reviewed the previous PR.

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -4,9 +4,10 @@ from assertions import assert_none, assert_one
 import tempfile
 import os
 
+
 class TestCompaction(Tester):
 
-    __test__= False
+    __test__ = False
 
     def __init__(self, *args, **kwargs):
         kwargs['cluster_options'] = {'start_rpc': 'true'}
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I'm +1. To 587f821 specifically as well

shawnkumar:
I'm +1 as well.

mambocab:
These tests are now fully awesome thanks to an assist from @krummas in 828c2bd . Can I get a quick once-over, @ptnapoleon? The commits since Marcus' commit are new; the others are the commits mentioned above, but rebased onto master.

ptnapoleon:
I left a comment on Marcus's commit. If the tests pass on 2.0, then you're free to ignore that comment. +1, otherwise

mambocab:
Thanks, merging :fireworks:

PR #331, opened 2015-06-17
Adding basic scrub tests for versions 2.0 and 2.1 and one test for CASSANDRA-9591

stef1927:
I have added two basic scrub tests, one for nodetool and one for standalone scrubbing. We already had some tests but they only run for >= 2.2 because scrubbing of secondary indexes is only available in 2.2. There is a test called test_nodetool_scrub() which runs since 2.1 but it doesn't seem to check much.

I have also added a test for [CASSANDRA-9591] (https://issues.apache.org/jira/browse/CASSANDRA-9591).

### DIFF
--- a/scrub_test.py
+++ b/scrub_test.py
@@ -6,16 +6,18 @@ import uuid
 
 from ccmlib import common
 from dtest import Tester, debug
-from tools import since
+from tools import since, require
 import time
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1!

PR #332, opened 2015-06-18
hinted handoff test

shawnkumar:
Please don't merge yet as hinted handoff has not been committed yet

### DIFF
--- a/hinted_handoff_test.py
+++ b/hinted_handoff_test.py
@@ -1,8 +1,8 @@
-from dtest import Tester, debug, DISABLE_VNODES
-from ccmlib.node import Node, NodeError, TimeoutError
-from cassandra import ConsistencyLevel, Unavailable, ReadTimeout
+from dtest import Tester, debug
+from ccmlib.node import Node
+from cassandra import ConsistencyLevel
 from cassandra.query import SimpleStatement
-from tools import since, InterruptBootstrap
+from tools import since
 from time import time, sleep
 from assertions import assert_none, assert_invalid
 from jmxutils import make_mbean, JolokiaAgent, remove_perf_disable_shared_mem
@@ -11,8 +11,8 @@ from jmxutils import make_mbean, JolokiaAgent, remove_perf_disable_shared_mem
 class TestHintedHandoff(Tester):
 
     def check_delivery(self, node):
-        node2_mbean = make_mbean('metrics', type='HintedHandoffManager', name='Hints_created-127.0.0.2')
-        node3_mbean = make_mbean('metrics', type='HintedHandoffManager', name='Hints_created-127.0.0.3')
+        node2_mbean = make_mbean('metrics', type='HintedHandOffManager', name='Hints_created-127.0.0.2')
+        node3_mbean = make_mbean('metrics', type='HintedHandOffManager', name='Hints_created-127.0.0.3')
         handedoff = False
         timeout = time() + 90.00
         while not handedoff and time() < timeout:
@@ -20,12 +20,12 @@ class TestHintedHandoff(Tester):
                 try:
                     node2 = jmx.read_attribute(node2_mbean, 'Count')
                     node3 = jmx.read_attribute(node3_mbean, 'Count')
-                    if node2 == "10200" and node3 == "10200":
+                    if 10200 <= node2 <= 10210 and 10200 <= node3 <= 10210:
                         handedoff = True
                     else:
                         debug(node3)
                         debug(node2)
-                        self.fail()
+                        return handedoff
                 except Exception,e:
                     debug(str(e))
         return handedoff
@@ -43,7 +43,7 @@ class TestHintedHandoff(Tester):
         """
         cluster = self.cluster
         cluster.populate(3)
-        [node1, node2, node3] = cluster.nodelist()
+        node1, node2, node3 = cluster.nodelist()
 
         remove_perf_disable_shared_mem(node1)
         cluster.start(wait_for_binary_proto=True)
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Since every method is marked `since('3.0')`, can you just move that to a class level annotation?

ptnapoleon:
It looks like some of these imports aren't used, like `DISABLE_VNODES`

ptnapoleon:
I love the doxygen markup.

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+        - sethintedhandoffthrottlekb to low kb
+        - take down node, write data to others
+        - bring up node
+        - check statushandoff is reporting accurate stats
+        - try pausing handoff, check status again to ensure, resume handoff
+        - verify data
+        """
+        cluster = self.cluster
+        cluster.populate(3)
+        [node1, node2, node3] = cluster.nodelist()
```
ptnapoleon: Please change to `node1, node2, node3 = cluster.nodelist()`

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+        for x in range(0, 100):
+            delq = "DELETE FROM hhtest.hhtab WHERE key = {key}".format(key=str(x))
+            cursor.execute(delq)
+
+        for x in range(100, 200):
+            updateq = "INSERT INTO hhtest.hhtab(key,val) VALUES ({key}, 100000)".format(key=str(x))
+            cursor.execute(updateq)
+
+        node1.stop(gently=False)
+        node1.set_install_dir(version = '3.0')
```
ptnapoleon: Does this really work?

ptnapoleon:
Do you want to move `hintedhandoff_test` from consistency.py into this file now? As well as evaluate if that test should change?

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+        node2_mbean = make_mbean('metrics', type='HintedHandoffManager', name='Hints_created-127.0.0.2')
+        node3_mbean = make_mbean('metrics', type='HintedHandoffManager', name='Hints_created-127.0.0.3')
+        handedoff = False
+        timeout = time() + 90.00
+        while not handedoff and time() < timeout:
+            with JolokiaAgent(node) as jmx:
+                try:
+                    node2 = jmx.read_attribute(node2_mbean, 'Count')
+                    node3 = jmx.read_attribute(node3_mbean, 'Count')
+                    if node2 == "10200" and node3 == "10200":
```
mambocab: This looks like it will always end the loop on the first execution, either by setting `handedoff` to true or by `fail`ing the test. Is that right? and if so, why is it in a loop?

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+        node2.start()
+        node3.start()
+
+        handedoff = self.check_delivery(node1)
+
+        self.assertTrue(handedoff)
+
+        node1.stop(gently=False)
+
+        for node in [node2,node3]:
```
mambocab: PEP8 violation: should be `[node2, node3]`

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+
+        node1.stop(gently=False)
+        node1.set_install_dir(version = '3.0')
+        node1.start()
+
+        #check system.hints truncated
+        assert_invalid(cursor, "SELECT * FROM system.hints LIMIT 10")
+
+        node2.set_install_dir(version = '3.0')
+        node3.set_install_dir(version = '3.0')
```
mambocab: PEP8 violation: no whitespace around the `=`.

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+from cassandra import ConsistencyLevel, Unavailable, ReadTimeout
+from cassandra.query import SimpleStatement
+from tools import since, InterruptBootstrap
+from time import time, sleep
+from assertions import assert_none, assert_invalid
+from jmxutils import make_mbean, JolokiaAgent, remove_perf_disable_shared_mem
+
+
+class TestHintedHandoff(Tester):
+
```
mambocab: I'm not clear on what part of this feature is new. Where applicable, could you please add a `@jira_ticket` annotation if there's one tracking a new feature?

hinted_handoff_test.py
```diff
@@ -0,0 +1,234 @@
...
+        for x in range(0, numhints):
+            insq = SimpleStatement("INSERT INTO hhtest.hhtab(key,val) VALUES ({key}, {key})".format(key=str(x)), consistency_level=ConsistencyLevel.TWO)
+            cursor.execute(insq)
+        
+        node3.start()
+
+        notstarted = True
+        while notstarted:
+            output = node1.nodetool("statushandoff", capture_output=True)
+            if "running" in output:
```
mambocab: Asking as someone who doesn't know anything about the nodetool handoff commands: How confident are you that `'running'` will never appear in the output unless handoff is actually running? Same goes for below where the code checks for `'paused'` and `'running'`. If there's a JMX endpoint for this info, I'd be more comfortable with assertions based on that.

hinted_handoff_test.py
mambocab: Also, I don't know why we expect the counts to have these values. It looks like they depend on the number of rows being written to the nodes before this method is called? Even if it's always the same, pass it in as a parameter so it's clearer what it is and where it comes from.

mambocab:
Good tests! My PEP8 comments are nits, obviously. If you think it might be more readable, consider using prepared statements and `execute_concurrent_with_args` instead of string formatting and loops.

mambocab:
Tracking this PR internally with CSTAR-440.

ptnapoleon:
Closing this, for obvious reasons

PR #334, opened 2015-06-19
Trying to fix Jenkins failure, CASSANDRA-9102

stef1927:
It seems consistency_test.py:TestAvailability.test_network_topology_strategy fails on Jenkins on all branches. However I cannot reproduce it on my box, it passes every time. At LOCAL_QUORUM a write should fail if only one node is up and two are down, given that rf is three. So I am guessing we need to wait for other nodes to notice, but it is only my best guess as I don't know how to download the node log files from Jenkins.

### DIFF
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -221,7 +221,7 @@ class TestAvailability(TestHelper):
                 for combination in combinations:
                     self._test_insert_query_from_node(session, i, rf_factors, nodes_alive, *combination)
 
-                self.cluster.nodelist()[node].stop()
+                self.cluster.nodelist()[node].stop(wait_other_notice=True)
                 nodes_alive[i] = nodes_alive[i] - 1
 
     def _test_insert_query_from_node(self, session, dc_idx, rf_factors, num_nodes_alive, write_cl, read_cl, serial_cl=None, check_ret=True):


mambocab:
I poked around on Jenkins to see if I could find the logs for this. Somehow they didn't get saved on the most recent build, but on build 255 they did. You can find them by:

- going [here](http://cassci.datastax.com/view/trunk/job/trunk_dtest/255/)
- clicking 'Build Artifacts' to look at the saved files from the test
- downloading `jenkins-trunk_dtest-255_logs.000.tar.gz` there.

That archive should contain the node logs you want. It won't always be in the `000` archive; that's just where the job bucketizer put it this time.

mambocab:
That said, I don't think this change could possibly hurt, so I'm +1. Thanks for keeping an eye on the failing tests!

ptnapoleon:
I do recall fixing this test originally, so I would totally believe the current issues are related to some sort of race condition like this. Also, +1.

stef1927:
It seems it's still failing. I noticed we are still using shared connections in TestAvailability and I pushed [a commit] (https://github.com/stef1927/cassandra-dtest/commit/ca913534ee056b691191971d28e26921bc5e0f74) to fix this. Is there a way to run my branch on Jenkins and to add DEBUG=true? Otherwise we keep on merging stuff into master that we might not need.

PR #335, opened 2015-06-19
Test to check JMX metrics

steveandwang:
Test to see whether JMX changes as you'd expect it to after running C*.

### DIFF
--- a/jmxmetrics_test.py
+++ b/jmxmetrics_test.py
@@ -1,0 +1,120 @@
+from dtest import Tester
+from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
+
+# Currently only have attributes that are incrementing. 
+# MBEAN_VALUES are expressed in tuple with the first value being the class, 
+# the package (may be tuple), the attribute, and then the value. 
+MBEAN_VALUES_PRE = [('metrics', ('ColumnFamily', 'AllMemtablesLiveDataSize'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'AllMemtablesHeapSize'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'AllMemtablesOffHeapSize'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'BloomFilterDiskSpaceUsed'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'BloomFilterFalsePositives'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'IndexSummaryOffHeapMemoryUsed'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'LiveDiskSpaceUsed'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'LiveSSTableCount'), 'Value', 'MBeanIncrement'), 
+                        ('metrics', ('ColumnFamily', 'MemtableColumnsCount'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'MemtableLiveDataSize'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'MemtableOnHeapSize'), 'Value', 'MBeanIncrement'),
+                        ('metrics', ('ColumnFamily', 'MemtableSwitchCount'), 'Value', 'MBeanIncrement'),
+                        ('db', 'IndexSummaries', 'MemoryPoolSizeInMB', 'MBeanIncrement'),
+                        ('db', 'IndexSummaries', 'IndexIntervals', 'MBeanIncrement'),
+                        ('db', 'Caches', 'CounterCacheKeysToSave', 2147483647),
+                        ('db', 'Caches', 'CounterCacheSavePeriodInSeconds', 7200),
+                        ('metrics', ('ColumnFamily', 'MaxRowSize'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'MemtableOffHeapSize'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'MinRowSize'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'PendingCompactions'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'RowCacheHit'), 'Value', 'MBeanEqual'),
+                        ('metrics', ('ColumnFamily', 'CompressionRatio'), 'Value', 'MBeanDecrement'),
+                        ('metrics', ('ColumnFamily', 'MeanRowSize'), 'Value', 'MBeanDecrement'),
+                        ('db', 'BatchlogManager', 'TotalBatchesReplayed', 0),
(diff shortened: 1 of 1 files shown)


jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+                before.append(jmx.read_attribute(mbean, attribute))
+
+            node.stress(['write', 'n=100K', '-port jmx=7100'])
+
+            attr_counter = 0
+            for package, bean, attribute, expected in MBEAN_VALUES:
+                a_value = jmx.read_attribute(mbeans[attr_counter], attribute)
+                b_value = before[attr_counter]
+                if expected == 'MBeanIncrement':
+                    if b_value > a_value:
```
EnigmaCurry: I think we want >= here

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+                        errors += mbeans[attr_counter] + " has a before value of " + str(b_value) + " and after value of " + str(a_value) + " and did not increment" + "\n"
```
EnigmaCurry: String concatentation is usually an expensive operation. Better to make errors a list and then append the errors to that. If you still want it to be a string at the end, you can do : 

    errors = "\n".join(errors)

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+                elif expected == 'MBeanZero':
+                    if not (b_value == 0 and a_value == 0):
+                        errors += mbeans[attr_counter] + " has a before value of " + str(b_value) + " and after value of " + str(a_value) + " and they do not equal zero" + "\n"
+                # If expected is none of the above, then expected should be a number.
+                else:
+                    if a_value != expected:
+                        errors += mbeans[attr_counter] + " has an after value of " + str(a_value) + " which does not equal " + str(expected) + "\n"
+                attr_counter += 1
+
+            assert len(errors) == 0, errors
```
EnigmaCurry: assert works, but it doesn't print as nice a message as the unittest assert methods do. Better to do:

    self.assertEqual(len(errors), 0)

EnigmaCurry:
I haven't run this yet, but it looks good.

EnigmaCurry:
Find some other metrics for decrement, equal, and zero so we're sure the other tests work too, then looks good to merge to me.

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+                ('metrics', ('ColumnFamily', 'LiveSSTableCount'), 'Value', 'MBeanIncrement'), 
+                ('metrics', ('ColumnFamily', 'MemtableColumnsCount'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'MemtableLiveDataSize'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'MemtableOnHeapSize'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'MemtableSwitchCount'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'PendingFlushes'), 'Value', 'MBeanIncrement'),
+                ('db', 'IndexSummaries', 'MemoryPoolSizeInMB', 'MBeanIncrement'),
+                ('db', 'IndexSummaries', 'IndexIntervals', 'MBeanIncrement')]
+
+class JMXValueTest(Tester):
```
ptnapoleon: For consistency's sake with the other files, this should be `TestJMXMetrics`

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
+
+    def begin_test(self):
```
ptnapoleon: Add some doxygen markup on this, please and thanks. Include at least the `@jira_ticket` tag. Also, could you give the test a more descriptive name?

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+            for package, bean, attribute, expected in MBEAN_VALUES:
+                # In the case that the file name is longer, then we put it in the form of tuple in bean.
+                if type(bean) == tuple:
+                    mbean = make_mbean(package, type = bean[0], name = bean[1])
+                else:
+                    mbean = make_mbean(package, bean)
+                mbeans.append(mbean)
+                before.append(jmx.read_attribute(mbean, attribute))
+
+            node.stress(['write', 'n=100K', '-port jmx=7100'])
```
ptnapoleon: Is this test going to run on 2.0 as well? If so, this stress call will fail.

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+class JMXValueTest(Tester):
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
+
+    def begin_test(self):
+        cluster = self.cluster
+        cluster.populate(1)
+        node = cluster.nodelist()[0]
+        remove_perf_disable_shared_mem(node)
```
ptnapoleon: Could you leave a comment explaining why you did this? It's obvious to anyone familiar with the issue, but not everyone reading the test will be.

jmxmetrics_test.py
```diff
@@ -0,0 +1,94 @@
...
+                ('metrics', ('ColumnFamily', 'MemtableLiveDataSize'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'MemtableOnHeapSize'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'MemtableSwitchCount'), 'Value', 'MBeanIncrement'),
+                ('metrics', ('ColumnFamily', 'PendingFlushes'), 'Value', 'MBeanIncrement'),
+                ('db', 'IndexSummaries', 'MemoryPoolSizeInMB', 'MBeanIncrement'),
+                ('db', 'IndexSummaries', 'IndexIntervals', 'MBeanIncrement')]
+
+class JMXValueTest(Tester):
+
+    def __init__(self, *args, **kwargs):
```
ptnapoleon: This is totally redundant, no need to include this `__init__` if you aren't modifying anything.

stef1927:
Please check [CASSANDRA-9448] (https://issues.apache.org/jira/browse/CASSANDRA-9448), some of the metrics names will change in 3.0.

jmxmetrics_test.py
```diff
@@ -0,0 +1,120 @@
...
+# MBEAN_VALUES_POST_3.0 = 
+
+class TestJMXMetrics(Tester):
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
+
+    def begin_test(self):
+        """
+        @jira_ticket 7436
```
ptnapoleon: Should be `@jira_ticket CASSANDRA-7436` please

jmxmetrics_test.py
```diff
@@ -0,0 +1,120 @@
...
+                # In the case that the file name is longer, then we put it in the form of tuple in bean.
+                if type(bean) == tuple:
+                    mbean = make_mbean(package, type = bean[0], name = bean[1])
+                else:
+                    mbean = make_mbean(package, bean)
+                mbeans.append(mbean)
+                before.append(jmx.read_attribute(mbean, attribute))
+
+            if cluster.version() < "2.1":
+                node.stress(['-o', 'insert', '-n', '100', '-p', '7100'])
```
ptnapoleon: You're inserting 100K lines in 2.1, but only 100 in 2.0

ptnapoleon:
+1! Great job! :+1: :fireworks:

PR #336, opened 2015-06-22
Patch for CASSANDRA-9448

stef1927:
This is a patch for [CASSANDRA-9448] (https://issues.apache.org/jira/browse/CASSANDRA-9448) to support renaming of metrics in 3.0. I also modified the tests for CASSANDRA-7436, P.R. no 335.

It should be merged only when CASSANDRA-9448 is committed, else tests on trunk will fail.

### DIFF
--- a/deletion_test.py
+++ b/deletion_test.py
@@ -61,16 +61,18 @@ class TestDeletion(Tester):
 def memtable_size(node, keyspace, table):
     version = node.get_cassandra_version()
     name = 'MemtableOnHeapSize' if version >= '2.1' else 'MemtableDataSize'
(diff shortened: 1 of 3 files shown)


stef1927:
Please review this as we are close to committing CASSANDRA-9448.

ptnapoleon:
I'm +1. @mambocab ?

mambocab:
I haven't had a chance to run these, but if you have, @ptnapoleon, I'm +1.

stef1927:
I don't think the last commit has been reviewed. I'd like to merge this if possible, since CASSANDRA-9448 has already been committed.

PR #150, opened 2015-01-15
7653 Test for upgrading to new auth schema

ptnapoleon:
This test is not working due to a problem with the python-driver. Once that is resolved, this should be merged in.

PR #338, opened 2015-06-22
7653 with upgrade test

ptnapoleon:
Rebase of #150

### DIFF
--- a/upgrade_internal_auth_test.py
+++ b/upgrade_internal_auth_test.py
@@ -1,0 +1,141 @@
+from cassandra import Unauthorized
+from dtest import Tester, debug
+from assertions import assert_all, assert_invalid
+from tools import since
+import time
+
+from ccmlib import repository
+
+@since('3.0')
+class TestAuthUpgrade(Tester):
+
+    def __init__(self, *args, **kwargs):
+        # Ignore these log patterns:
+        self.ignore_log_patterns = [
+            # This one occurs if we do a non-rolling upgrade, the node
+            # it's trying to send the migration to hasn't started yet,
+            # and when it does, it gets replayed and everything is fine.
+            r'Can\'t send migration request: node.*is down',
+        ]
+
+        # Force cluster options that are common among versions:
+        kwargs['cluster_options'] = {'authenticator': 'PasswordAuthenticator',
+                                     'authorizer': 'CassandraAuthorizer'}
+        Tester.__init__(self, *args, **kwargs)
+
+    def upgrade_with_internal_auth_test(self):
+        """
+        Tests upgrade between 2.1->3.0 as the schema and apis around authn/authz changed
+
+        @jira_ticket CASSANDRA-7653
+        """
+        cluster = self.cluster
+
+        # Forcing cluster version on purpose
+        cluster.set_install_dir(version="git:cassandra-2.1")
+        cluster.populate(3).start()
+
+        node1, node2, node3 = cluster.nodelist()
+
+        # wait for default superuser creation
+        # The log message
+        # node.watch_log_for('Created default superuser')
+        # will only appear on one of the three nodes, and we don't know
+        # which ahead of time. Grepping all three in parallel is unpleasant.
+        # See auth_test and auth_roles test for instances of this as well.
+        # Should be fixed by C*-6177
+        time.sleep(15)
+
+        session = self.patient_cql_connection(node1, user='cassandra', password='cassandra')
+        session.execute("CREATE USER klaus WITH PASSWORD '12345' SUPERUSER")
+        session.execute("CREATE USER michael WITH PASSWORD '54321' NOSUPERUSER")
+        session.execute("CREATE KEYSPACE ks WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}")
+        session.execute("CREATE TABLE ks.cf1 (id int primary key, val int)")
+        session.execute("CREATE TABLE ks.cf2 (id int primary key, val int)")
+        session.execute("GRANT MODIFY ON ks.cf1 TO michael")
+        session.execute("GRANT SELECT ON ks.cf2 TO michael")
+
+        self.check_permissions(node1)
+        session.shutdown()
+        # upgrade the cluster to 3.0, restarting node 1 in the process
+        # note - will need changing when we branch for 3.0
+        self.upgrade_to_version("git:trunk", nodes=[node1])
+        # conversion of legacy auth info won't complete on the first upgraded node
+        node1.watch_log_for('Unable to complete conversion of legacy permissions')
+        time.sleep(1)
+        # run the permissions checking queries on the upgraded node
+        # this will be using the legacy tables as the conversion didn't complete
+        self.check_permissions(node1)
+        # and check on those still on the old version
+        self.check_permissions(node2)
+        self.check_permissions(node3)
+
+        # now upgrade the remaining nodes
+        self.restart_nodes_to_upgrade([node2, node3])
+        time.sleep(15)
+        self.check_permissions(node2)
+        self.check_permissions(node3)
+
+        # we should now be able to drop the old auth tables
+        session = self.patient_cql_connection(node1, user='cassandra', password='cassandra')
+        session.execute('DROP TABLE system_auth.users')
+        session.execute('DROP TABLE system_auth.credentials')
+        session.execute('DROP TABLE system_auth.permissions')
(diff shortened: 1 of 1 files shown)


ptnapoleon:
@mambocab to review. This does not pass on OSX due to CASSANDRA-8220, but that is totally unrelated to this test/feature.

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,136 @@
...
+        cluster = self.cluster
+
+        # Forcing cluster version on purpose
+        cluster.set_install_dir(version="git:cassandra-2.1")
+        cluster.populate(3).start()
+
+        node1, node2, node3 = cluster.nodelist()
+
+        # wait for default superuser creation
+        time.sleep(15)
```
mambocab: Waiting on the log entry for this would be better: https://github.com/riptano/cassandra-dtest/blob/daafea370b6e83ec84eaa330ca89978902f5eea1/cqlsh_tests/cqlsh_tests.py#L456

ptnapoleon: My understanding might be wrong, but only one node in the cluster will see that log message, so it's unpleasant to grep for in parallel. I'll see if we grep in parallel anywhere, so I can just copy the logic.

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,136 @@
...
+
+        cassandra = self.patient_cql_connection(node1, user='cassandra', password='cassandra')
```
mambocab: This variable is confusingly-named, as there's a module called `cassandra`. `session`, I think, is the going name for these objects.

upgrade_internal_auth_test.py
mambocab: That could be a good utility method for `ccmlib.cluster.Cluster`.

upgrade_internal_auth_test.py
ptnapoleon: This naming scheme was copied from auth_test and auth_roles test, where instead of `session`, the variable is named after the current user. I understand the desire not to collide with the module though, so I've renamed all instances of `cassandra` to just `session`.

upgrade_internal_auth_test.py
ptnapoleon: Agreed.

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,141 @@
...
+        session.execute("CREATE KEYSPACE ks WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}")
+        session.execute("CREATE TABLE ks.cf1 (id int primary key, val int)")
+        session.execute("CREATE TABLE ks.cf2 (id int primary key, val int)")
+        session.execute("GRANT MODIFY ON ks.cf1 TO michael")
+        session.execute("GRANT SELECT ON ks.cf2 TO michael")
+
+        self.check_permissions(node1)
+        session.shutdown()
+        # upgrade the cluster to 3.0, restarting node 1 in the process
+        ## note - will need changing when we branch for 3.0
```
mambocab: PEP8 nit -- comment lines start with '`# `'.

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,141 @@
...
+        self.upgrade_to_version("git:trunk", nodes=[node1])
+        # conversion of legacy auth info won't complete on the first upgraded node
+        node1.watch_log_for('Unable to complete conversion of legacy permissions')
+        time.sleep(1)
```
mambocab: What are we waiting on here? A comment would be nice.

ptnapoleon: I have no idea. @beobal ?

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,141 @@
...
+        session.execute("GRANT SELECT ON ks.cf2 TO michael")
+
+        self.check_permissions(node1)
+        session.shutdown()
+        # upgrade the cluster to 3.0, restarting node 1 in the process
+        # note - will need changing when we branch for 3.0
+        self.upgrade_to_version("git:trunk", nodes=[node1])
+        # conversion of legacy auth info won't complete on the first upgraded node
+        node1.watch_log_for('Unable to complete conversion of legacy permissions')
+        time.sleep(1)
```
ptnapoleon: Github covered up the comment, @beobal, what are we sleeping here for?

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,141 @@
...
+        # run the permissions checking queries on the upgraded node
+        # this will be using the legacy tables as the conversion didn't complete
+        self.check_permissions(node1)
+        # and check on those still on the old version
+        self.check_permissions(node2)
+        self.check_permissions(node3)
+
+        # now upgrade the remaining nodes
+        self.restart_nodes_to_upgrade([node2, node3])
+        time.sleep(15)
```
ptnapoleon: Or here?

upgrade_internal_auth_test.py
beobal: I don't remember, does it removing the sleep(s) have any effect?

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,139 @@
...
+            node.drain()
+            node.watch_log_for("DRAINED")
+            node.stop(wait_other_notice=False)
+
+        # Update Cassandra Directory
+        for node in nodes:
+            repository.clean_all()
+            debug(repository.GIT_REPO)
+            node.set_install_dir(version=tag, verbose=True)
+            debug("Set new cassandra dir for %s: %s" % (node.name, node.get_install_dir()))
```
mambocab: Here and elsewhere, favor `format` over `%`.

upgrade_internal_auth_test.py
```diff
@@ -0,0 +1,139 @@
...
+        node1.watch_log_for('Unable to complete conversion of legacy permissions')
+        # run the permissions checking queries on the upgraded node
+        # this will be using the legacy tables as the conversion didn't complete
+        self.check_permissions(node1)
+        # and check on those still on the old version
+        self.check_permissions(node2)
+        self.check_permissions(node3)
+
+        # now upgrade the remaining nodes
+        self.restart_nodes_to_upgrade([node2, node3])
```
mambocab: I had to dig through the different methods called to make sure that node 1, 2, and 3 would all have the same version. The story here is: `restart_nodes_to_upgrade` upgrades `node{2,3}` to the cluster's current version. The only reason this is set to the same version as `node1` is because in `upgrade_to_version` the cluster's current version is set. I didn't expect that, given the signature of `upgrade_to_version`.

This is too dependent on implicit state for me. I'd prefer the cluster's version be set explicitly in the test, or be passed as an argument to `restart_nodes_to_upgrade`.

mambocab:
Overall these are well written tests. However, I haven't seen them pass -- the test times out waiting for `node1` to start after upgrading to `git:trunk`. [Here](https://gist.github.com/mambocab/854ddb468b9d18ebd2b0) is the traceback..

ptnapoleon:
@beobal Does the trace Jim linked look like a bug to you?

beobal:
There's a couple of things we need to address in light of changes to the anticipated C* release plan. Initially, 7653 was intended to be included in 3.0, but since we added the 2.2 release the test is slightly outdated.
First, of all we need to change the @since annotation from 3.0 to 2.2. Second, the test currently tries to upgrade from 2.1 to trunk (i.e. 3.0). As I understand it, we're *not* planning on making a step through 2.2 a mandatory part of the upgrade path to 3.0, so we'll need to test both 2.1->2.2 and 2.1->3.0 upgrades.

Regarding the stacktrace, NoClassDefFound means that either a class definition was changed underneath the JVM, so the class is found on the classpath but it is incompatible with whatever is referencing it, or some initialization error occured during the loading of that class. My guess would be the former, given that the dtest approach to upgrading is to switch the dir containing the class files. I've reworked the test to upgrade the nodes individually rather than setting the cluster version and having them update to that. That seems to have improved things as since doing that I'm not seeing any failures going from 2.1 to either 2.2 or trunk. I've left a todo in to change the target branch for the 3.0 upgrade from trunk once we branch for 3.0.

ptnapoleon:
@mambocab, want to re-review?

mambocab:
Gorgeous, I'm +1.

ptnapoleon:
Thanks @beobal

PR #339, opened 2015-06-22
Cql test failure

mambocab:
There's some incorrect error checking that causes failures on trunk ([cassci](http://cassci.datastax.com/view/trunk/job/trunk_dtest/lastCompletedBuild/testReport/cql_tests/TestCQL/validate_counter_regular_test/)). 3cf8ce9 fixes that.

3e8e3e8 and aad92ce deal with some code that was commented out.

The rest of the commits are whitespace, markup, and various PEP8 cleanup.

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -50,7 +50,9 @@ class TestCQL(Tester):
         return session
 
     def static_cf_test(self):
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

PR #340, opened 2015-06-23
Tests for CASSANDRA-9035

stef1927:
These are some basic hinted handoff tests, for CASSANDRA-9035.

### DIFF
--- a/hintedhandoff_test.py
+++ b/hintedhandoff_test.py
@@ -1,8 +1,6 @@
 from dtest import Tester, debug, DISABLE_VNODES
-from assertions import assert_unavailable, assert_none
-from tools import require, create_c1c2_table, insert_c1c2, query_c1c2, insert_columns, rows_to_list
+from tools import require, create_c1c2_table, insert_c1c2, query_c1c2
 from cassandra import ConsistencyLevel
-from cassandra.query import SimpleStatement
 
 @require("9035")
 class TestHintedHandoff(Tester):
@@ -29,7 +27,7 @@ class TestHintedHandoff(Tester):
         Launch a nodetool command and check there is no error, return the result
         """
         out, err = node.nodetool(cmd, capture_output=True)
-        assert len(err) == 0
+        self.assertEqual('', err)
         return out
 
     def _do_hinted_handoff(self, node1, node2, enabled):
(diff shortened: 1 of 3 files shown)


ptnapoleon:
Can you possibly use the unittest asserts, such as `self.assertEqual` rather than the python `assert` keyword?

hintedhandoff_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+        if enabled:
+            node1.watch_log_for(["Finished hinted"], from_mark=log_mark, timeout=120)
+
+        node1.stop(wait_other_notice=True)
+
+        # Check node2 for all the keys that should have been delivered via HH if enabled or not if not enabled
+        cursor = self.patient_cql_connection(node2, keyspace='ks')
+        for n in xrange(0, 100):
+            query_c1c2(cursor, n, ConsistencyLevel.ONE, tolerate_missing=not enabled, must_be_missing=not enabled)
```
ptnapoleon: Is this boolean correct? If HH is enabled, why would tolerate missing keys?

hintedhandoff_test.py
```diff
@@ -0,0 +1,142 @@
...
+            cluster.populate([2], tokens=tokens).start()
+
+        return cluster.nodelist()
+
+    def _launch_nodetool_cmd(self, node, cmd):
+        """
+        Launch a nodetool command and check there is no error, return the result
+        """
+        out, err = node.nodetool(cmd, capture_output=True)
+        assert len(err) == 0
```
mambocab: Comparing `len(iterable)` to 0 isn't idiomatic Python. As Philip mentioned, this should use the `unittest` asserts as well, so this can be written `self.assertFalse(err)`. It might be clearer as `self.assertEqual(err, '')` (as long as `node.nodetool` returns a string, which I think it does). It's up to you, whichever you think is better.

hintedhandoff_test.py
stef1927: If HH enabled then not enabled is false, hence tolerate_missing=false so we don't tolerate missing keys right?

stef1927:
Thanks, I've pushed a commit with the assertions fixed. I've also cleaned up the imports and switched to exclusive connections.

hintedhandoff_test.py
ptnapoleon: No, you're right. I'm not sure why I was reading it backwards. It seems clear enough now.

ptnapoleon:
+1

PR #341, opened 2015-06-23
Add a small pause when waiting for inbound pages

knifewine:
@EnigmaCurry spotted this problem yesterday, we're probably needlessly consuming cpu
while waiting for something that might take a second or two.


ptnapoleon:
+1

Issue #251, opened 2015-05-01
replace_address_test: replace_with_reset_resume_state_test and resumable_replace_test hang server


mshuler:
replace_with_reset_resume_state_test and resumable_replace_test both completely hang the test server frequently, although they periodically do pass.

The line they are both hanging on is:
```python
try:
    node4.start(jvm_args=["-Dcassandra.replace_address_first_boot=127.0.0.3"])
```

Node logs and stdout from one run: http://12.am/tmp/resumable_replace_test.tar.gz - the other test looks the same with node4 repeating. These tests are excluded from trunk, currently.

mambocab:
I wasn't able to reproduce this locally or on CassCI:

http://cassci.datastax.com/job/scratch_mambocab-replace-address/6/testReport/history/

(Test run #1 failed due to misconfiguration; ignore that one.)

Does it look to you like whatever problem there was is fixed now? If it is, I'll prepare a PR to re-include these tests.

ptnapoleon:
If they don't hang up the server, re-include them. We can always revert the inclusion

PR #342, opened 2015-06-23
re-includes tests from replace_address_test

mambocab:
At least running locally, they succeed now, as discussed over in #251.


ptnapoleon:
+1

PR #344, opened 2015-06-25
Add jdk switching to upgrade tests

ptnapoleon:


ptnapoleon:
@mambocab for review.

This fixes ~40 failures out of 64 upgrade tests.

upgrade_through_versions_test.py
```diff
@@ -118,6 +118,29 @@ def make_branch_str(_tuple):
 
     return 'cassandra-{}.{}'.format(_tuple[0], _tuple[1])
 
+def sanitize_version(version):
```
mambocab: Over in `tools.py` we use `LooseVersion`. It looks like it'll probably do for our purposes here too:

```python
In [1]: from distutils.version import LooseVersion

In [2]: LooseVersion('2.2.0rc1') < '2.2.0'  # though it isn't perfect
Out[2]: False

In [3]: LooseVersion('2.2.0rc1') > '2.1'
Out[3]: True

In [4]: LooseVersion('2.0.16') < '2.1'
Out[4]: True
```

Comparing version strings directly always worries me, which is why I suggest using `LooseVersion`.

mambocab: Oh, duh, I see the problem with the `cassandra-X.Y` strings now. But after that's stripped, this should probably just return a `LooseVersion` of the remaining string.

upgrade_through_versions_test.py
```diff
@@ -118,6 +118,29 @@ def make_branch_str(_tuple):
...
+
+def switch_jdks(version):
+    version = sanitize_version(version)
+    try:
+        if version < '2.1':
+            os.environ['JAVA_HOME'] = os.environ['JAVA7_HOME']
+        else:
+            os.environ['JAVA_HOME'] = os.environ['JAVA8_HOME']
+    except KeyError as e:
+        print("You need to set JAVA7_HOME and JAVA8_HOME to run these tests!")
```
mambocab: I think this should raise a `RuntimeError` with this string instead, maybe collecting information about which of the variables is missing first. We'll lose the `KeyError` on Python 2, but once we move to Python 3, we'll start getting [awesome double tracebacks](http://blog.ionelmc.ro/2014/08/03/the-most-underrated-feature-in-python-3/).

mambocab:
I feel fairly strongly about avoiding string comparisons for version numbers, and not-super-strongly about the `RuntimeError`. Definitely address the first, address the second if you agree with me, and I'll be +1.

PR #347, opened 2015-06-26
Prettify data

knifewine:
This is mostly cosmetic change, but allows data given to datahelp.py
to have a line that breaks between the headers and the data descriptions.
I also took this opportunity to fix up the imports in paging_test.py

I vetted the change by running some of the paging tests and all looks well.


ptnapoleon:
Beautiful. So Pythonic. +2.

PR #349, opened 2015-06-26
Add travis config with linter

mambocab:
(Thank you, @knifewine)

This adds a travis configuration file to the repo. It'll run `pep8` on changed lines, so idiomatic-ness will only improve.

It also runs `pyflakes` via `flake8` on all files. It only reports on importing errors, since @ptnapoleon fixed all those and all of this type of error found should be new.

This PR also removes unused imports and `from module import *`, as found by `pyflakes`, so we don't start off with failing linters. It also cleans up the imports with `isort` because why not?

For the errors we've ignored and for PEP8 checking on the entire codebase, we can take it in chunks, or just improve things as we change tests, with the help of this linting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: python
+python:
+  - "2.7"
+install:
+  - pip install pep8 flake8
+script:
+  # we want pyflakes to check all files for unused imports only
+  # we use flake8 because it allows us to ignore other warnings
+  # exclude the thrift directories - they contain auto-generated code
+  - flake8 --ignore=E,W,F811,F812,F821,F822,F823,F831,F841,N8,C9 --exclude=thrift_bindings,cassandra-thrift .
+  # feed changed lines with no context around them to pep8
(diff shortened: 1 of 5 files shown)


mambocab:
I don't know who has permissions to do this, but someone will have to set up the Travis service for this repo. Is that helpdesk?

mambocab:
Bah, not ready for merge yet. Need to add a warning to the ignored `flake8` errors.

mambocab:
Force pushed the fix. Good to go. I'll still hold off on merging until helpdesk can turn the service on.

mambocab:
[Here is an example](https://github.com/mambocab/cassandra-dtest/pull/2#partial-pull-merging) of the checking in action on a PR. It should demonstrate that the config works.

ptnapoleon:
-1 to forgetting "Linter is coming" from that image macro. +1 to the PR.

mambocab:
It looks like the plumbing is in place. Merging.

PR #350, opened 2015-06-29
Test for CASSANDRA-9601

stef1927:


cqlsh_tests/cqlsh_tests.py
```diff
@@ -1043,6 +1043,17 @@ def test_client_warnings(self):
 Warnings :
 Unlogged batch covering 2 partitions detected against table [client_warnings.test]. You should use a logged batch for atomicity, or asynchronous writes for performance.""")
 
+    @require("9601")
+    @since('2.2')
+    def test_connect_timeout(self):
+        self.cluster.populate(1)
```
ptnapoleon: Please add the docstring
```
"""
@jira_ticket CASSANDRA-9601
"""
```

ptnapoleon:
+1 after the listed change. Thanks

mambocab:
It'd be nice if this could test the timeout behavior itself -- but I'm not sure how to do that without somehow messing with or stubbing out Cassandra, which we don't don't have the tools to do. I'm +1 on the existing test (with the change ptnapoleon mentioned), but if you have any ideas on how to test the timeout itself, it'd be good to get those in too.

knifewine:
Maybe we could send SIGSTOP to cassandra, wait a little bit and then SIGCONT to resume.

mambocab:
I tried stopping the node via ccm before running the cqlsh statement -- the cqlsh call didn't time out, it just failed. The timeout seems to be for negotiating the connection, not for establishing it. Is what you're proposing different, @knifewine?

knifewine:
I'm overcomplicating things, a stop would accomplish the same, bummer that won't do the trick. Probably more effort than it's worth to mess around with connections, etc.

stef1927:
I've added the comment, thanks.

In order to test the real behavior, we could create an authenticator or authorizer that sleeps for a few seconds before authorizing all. According to the python documentation, this timeout covers the entire connection process, including authentication, and so this should work in theory.

It does require writing a java class, in a specific namespace, and adding this class to the cassandra process CLASSPATH. Probably best to add it in a test folder in the cassandra project, then make sure this code is compiled into the jar or a new jar file is also loaded. Then the authenticator is selected via cassandra.yaml.

Do we already have the capability to attach test java code to the cassandra process or would this have to be set-up entirely from scratch?

cqlsh_tests/cqlsh_tests.py
```diff
@@ -1043,6 +1043,20 @@ def test_client_warnings(self):
...
+    def test_connect_timeout(self):
+        """
+        @jira_ticket CASSANDRA-9601
+        """
+        self.cluster.populate(1)
+        self.cluster.start(wait_for_binary_proto=True)
+
+        node1, = self.cluster.nodelist()
+
+        stdout, stderr = self.run_cqlsh(node1, cmds = 'USE system', cqlsh_options=['--debug', '--connect_timeout=10'])
```
ptnapoleon: The PEP-8 linter is complaining about you having `cmds = 'USE system'` instead of `cmds='USE system'`. Unless I'm interpreting the output wrong, @mambocab ?

aboudreault: That seems correct. It should be without spaces.

mambocab: That's right, bullet point three here: https://www.python.org/dev/peps/pep-0008/#other-recommendations

mambocab:
> Do we already have the capability to attach test java code to the cassandra process or would this have to be set-up entirely from scratch?

As far as I know we have no such thing. If we could magically get this infrastructure for free, I'd love to have it. I don't know if it's worth the effort right now. Thoughts from @riptano/cassandra_test_engineering ? If nothing else, we could start cataloging cases where we'd like these tools, and implement it when our list reaches some critical mass.

cqlsh_tests/cqlsh_tests.py
stef1927: I fixed it and we've also renamed "connect_timeout" to "connect-timeout", see ticket. What is the linter command to run to check you have not introduced new warnings?

stef1927:
Agreed that there aren't many cases yet so let's see if more cases come up first.

cqlsh_tests/cqlsh_tests.py
aboudreault: Check the flake8 and pep8 command in 
https://github.com/riptano/cassandra-dtest/blob/master/.travis.yml . 
That should do it.

stef1927: Thanks

PR #351, opened 2015-06-29
introduce linter errors to test travis

mambocab:
Ignore this.


mambocab:
Ok, looks like it works \o/

PR #354, opened 2015-06-29
Remove redefinition of is_win function

ptnapoleon:
We're already importing this from ccmlib.common here. In
zero tests do we ever use the dtest implementation, only ever the
ccmlib one. They're identical code.

@mambocab to review


mambocab:
Nice find, looks good to me. +1.

PR #355, opened 2015-06-30
add test for stress sparseness distribution

mambocab:
This adds a couple smoke tests for the new feature added in [CASSANDRA-9522](https://issues.apache.org/jira/browse/CASSANDRA-9522).

I don't know if this is where we want tests of stress itself to live -- I'm happy to move them or to wait until other external tools' tests get merged.

### DIFF
--- a/stress_tool_test.py
+++ b/stress_tool_test.py
@@ -1,0 +1,51 @@
+from __future__ import division
+
+from dtest import Tester
+from tools import rows_to_list, since
+
+
+@since('3.0')
+class TestStressSparsenessRatio(Tester):
+    """
+    @jira_ticket CASSANDRA-9522
+
+    Tests for the `row-population-ratio` parameter to `cassandra-stress`.
+    """
+    def uniform_ratio_test(self):
+        """
+        Tests that the ratio-specifying string 'uniform(5..15)/50' results in
+        ~80% of the values written being non-null.
+        """
+        self.distribution_template(ratio_spec='uniform(5..15)/50',
+                                   expected_ratio=.8,
+                                   delta=.1)
+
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I don't object to having tests for stress, or for putting them here. What do others think?

stress_tool_test.py
```diff
@@ -0,0 +1,51 @@
...
+        @param delta the acceptable delta between the expected and actual ratios
+
+        A parameterized test for the `row-population-ratio` parameter to
+        `cassandra-stress`.
+        """
+        self.cluster.populate(1).start(wait_for_binary_proto=True)
+        node = self.cluster.nodelist()[0]
+        node.stress(['write', 'n=1000', '-rate', 'threads=50', '-col', 'n=FIXED(50)',
+                     '-insert', 'row-population-ratio={ratio_spec}'.format(ratio_spec=ratio_spec)])
+        session = self.patient_exclusive_cql_connection(node)
```
ptnapoleon: Why add the extra overhead of an exclusive connection if we only have one node?

ptnapoleon:
I'm +1

mambocab:
I've pushed up a change that makes the connection non-exclusive; thanks for the review @ptnapoleon.

Waiting for some feedback from others before I merge.

ptnapoleon:
@knifewine or @aboudreault, what are your feelings on this?

aboudreault:
No objection. +1. I don't think we need to add a test for all options but it is reasonable to add some when an issue is found.

mambocab:
Ok, commentary period's up. Thanks for the reviews/opinions, y'all, merging.

PR #358, opened 2015-07-01
Fix failing tests 9064

mambocab:
`test_describe` started failing (I assume) because of changes to the Python driver introduced in [CASSANDRA-9064](https://issues.apache.org/jira/browse/CASSANDRA-9064) -- the test assumed that the `min_threshold` and `max_threshold` would always be printed, but they aren't if they use the default values.

I also added a test for the other case, where the thresholds are set to non-default values, to make sure they are printed.

I also removed a few line continuations because I think they're gross.

The tests still all pass on `cassandra-2.{0,1,2}` and trunk, except for `cqlsh_tests/cqlsh_tests.py:TestCqlsh.test_describe_round_trip`, which still fails on 2.0 because the fix for 9064 wasn't applied there.


ptnapoleon:
+1

mambocab:
Thanks, merged.

BTW, I've filed [CASSANDRA-9703](https://issues.apache.org/jira/browse/CASSANDRA-9703) to track the failure on 2.0.

PR #360, opened 2015-07-02
Test for materialized views implementation (CASSANDRA-6477)

aboudreault:
Do not merge, but ready for some review. 

Test for materialized views implementation.

https://issues.apache.org/jira/browse/CASSANDRA-6477

I am waiting on 2 fixes/changes to complete this PR:

* Adding a datacenter or a single node produce unexpected results. (Bad data propagation)
* The drop table and drop column has to be changed to NOT drop the materialized views silently. (see ticket for more info)


materialized_views_test.py
```diff
@@ -0,0 +1,399 @@
...
+from tools import since, require, new_node
+from assertions import assert_one, assert_invalid
+
+
+@since('3.0')
+@require('6477')
+class TestMaterializedViews(Tester):
+    """
+    Test materialized views implementation.
+    @jira_ticket 6477
```
ptnapoleon: Make that `@jira_ticket CASSANDRA-6477` please

aboudreault: sure. fixed

aboudreault:
@mambocab When you have some time, I would appreciate a first review here :)

materialized_views_test.py
```diff
@@ -0,0 +1,399 @@
...
+from assertions import assert_one, assert_invalid
+
+
+@since('3.0')
+@require('6477')
+class TestMaterializedViews(Tester):
+    """
+    Test materialized views implementation.
+    @jira_ticket CASSANDRA-6477
+    @since 3.0
```
mambocab: Do we have a standard meaning for `@since`?

ptnapoleon: No :(

materialized_views_test.py
```diff
@@ -0,0 +1,399 @@
...
+
+        session.execute("CREATE TABLE t (id int PRIMARY KEY, v int)")
+
+        for i in xrange(1000):
+            session.execute("INSERT INTO t (id, v) VALUES ({v}, {v})".format(v=i))
+
+        session.execute(("CREATE MATERIALIZED VIEW t_by_v AS SELECT * FROM t PRIMARY KEY (v)"))
+
+        # wait propagation
+        time.sleep(10)
```
mambocab: Are there any logs you can wait on here? If they're distributed logs then keep an eye on https://github.com/pcmanus/ccm/issues/311, which could help you out.

materialized_views_test.py
```diff
@@ -0,0 +1,399 @@
...
+        assert len(result) == 2, "Expecting {} materialized view, got {}".format(2, len(result))
+
+        session.execute("DROP MATERIALIZED VIEW ks.users_by_state;")
+
+        result = session.execute(("SELECT * FROM system.schema_materializedviews "
+                                 "WHERE keyspace_name='ks' AND columnfamily_name='users'"))
+        assert len(result) == 1, "Expecting {} materialized view, got {}".format(1, len(result))
+
+    def drop_column_test(self):
+        """Test that we ... wrong REQUIRE CHANGE IN THE IMPLEMENTATION"""
```
mambocab: What is this docstring about? Same below.

aboudreault: nothing to be done here until the implementation is done

materialized_views_test.py
aboudreault: I will recheck that but I don't think I can watch for anything in the log. data are simply replicated async in BG.

materialized_views_test.py
```diff
@@ -0,0 +1,399 @@
...
+        session.execute("CREATE TABLE t (id int PRIMARY KEY, v int, v2 int, v3 int)")
+        session.execute("CREATE MATERIALIZED VIEW t_by_v2 AS SELECT * FROM t PRIMARY KEY (v2)")
+
+        for i in xrange(1000):
+            session.execute("INSERT INTO t (id, v, v2, v3) VALUES ({v}, {v}, {v}, {v}) USING TTL 30".format(v=i))
+
+        for i in xrange(1000):
+            assert_one(session, "SELECT * FROM t_by_v2 WHERE v2 = {}".format(i), [i, i, i, i])
+
+        time.sleep(30)
```
mambocab: Oof, this is a long sleep. Anything we can block on?

materialized_views_test.py
mambocab: We should probably request that feature. If it'd be useful for us, I imagine it'd be useful for diagnosing problems in the field, no?

materialized_views_test.py
aboudreault: Same issue. Not really. I can maybe reduce the the number rows read.... and maybe add a smart_sleep that could wait (30 seconds - *the time of the 1000 assert_one just before*)

mambocab:
I find the use of and relationship between the `prepare` and `prepare_user_table` a little confusing. I think there's sort of already a conventional meaning for the method name `prepare` in the codebase, which `prepare_user_table` subverts. If it were me, I'd add a kwarg (maybe named  `user_table`) to `prepare` and only create that table if the kwarg is truthy.

That's just me being a big ol' nerd, though, and if you disagree don't change anything.

aboudreault:
sounds good, will modify that prepare method.

materialized_views_test.py
aboudreault: @tjake Can I use nodetool("replaybatchlog") on all nodes to replace that kind of sleep?

aboudreault:
@mambocab  I pushed more changes since yesterday, your comments/review are welcome.

ptnapoleon:
You use a lot of the python `assert a == b, msg` syntax, rather than the unittest `self.assertEqual(a, b, msg)` which is preferred.

aboudreault:
Ready to merge. 

Big thanks to @ptnapoleon  and @tjake  for their multiple contributions in that big PR.

Issue #361, opened 2015-07-02
remove compaction_delete_test's dependency on sstable2json


mambocab:
`compaction_delete_test` in [`compaction_test.py`](compaction_test.py) checks for tombstones by using `sstable2json` and looking for the substring `'markedForDeleteAt'` in the output. Now that `sstable2json` has been removed from trunk, we need a different way to check for tombstones.

I can take this one on, I'm just not going to this afternoon, so if someone starts work on it before me, go ahead and assign yourself.

PR #362, opened 2015-07-02
Remove sstable2json tests

mambocab:
This deals with problems related to `sstable2json`. #361 is followup for some of it. The justification for each commit is in its message.


ptnapoleon:
+1. Make sure #361 doesn't go uncompleted though

mambocab:
I've fixed up some importing errors and the pep8 error Travis complained about. Once it's run through Travis clean, I'll merge.

PR #363, opened 2015-07-03
Removed @require tags for some committed tickets

stef1927:
I also changed the warning message of 9643, which was modified during commit. 

hinted_handoff.py will fail due to an NPE in the logs but this is not related to 9035 at all. This NPE is happening for other dtests too.


aboudreault:
LGTM when the minor linter issue is fixed.

stef1927:
Fixed.

PR #364, opened 2015-07-06
Fix to run the test for CASSANDRA-9591 on trunk as well

stef1927:
I had to change a previously written test to run it on trunk, since we now have a new sstable file, the statistics file, which seems to be required when scrubbing.


ptnapoleon:
+1

PR #365, opened 2015-07-06
adds tools for running tests with @require

mambocab:
This adds some new functionality surrounding the `@require` decorator. If you set the environment variable `IGNORE_REQUIRE` to `'yes'` or `'true'`, then `@require`d tests will run as though there were no decorator. It also modifies `@require`d tests with the `attr` decorator, so you can select them with `nosetests -a required`. 

Unfortunately, they'll still be skipped if you don't set `IGNORE_REQUIRE`; to have these play together nicely would require digging into the behavior of plugins, which is out of scope right now.

@aboudreault and/or @shawnkumar to review. Asking for extra-careful review of the docstring changes -- can you think of a better way to document this?

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -40,7 +40,7 @@ NUM_TOKENS = os.environ.get('NUM_TOKENS', '256')
 RECORD_COVERAGE = os.environ.get('RECORD_COVERAGE', '').lower() in ('yes', 'true')
 REUSE_CLUSTER = os.environ.get('REUSE_CLUSTER', '').lower() in ('yes', 'true')
 SILENCE_DRIVER_ON_SHUTDOWN = os.environ.get('SILENCE_DRIVER_ON_SHUTDOWN', 'true').lower() in ('yes', 'true')
-
+IGNORE_REQUIRE = os.environ.get('IGNORE_REQUIRE', '').lower() in ('yes', 'true')
 
 CURRENT_TEST = ""
 
--- a/tools.py
+++ b/tools.py
@@ -1,13 +1,20 @@
-from ccmlib.node import Node
+import fileinput
+import functools
+import re
+import subprocess
+import sys
+import time
+import unittest
 from distutils.version import LooseVersion
 from threading import Thread
-import re, sys, fileinput, time, unittest, functools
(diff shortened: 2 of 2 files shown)


mambocab:
@mshuler should have a look, too, to make sure this does what we need.

aboudreault:
Sorry Jim, what is the goal of this? I thought @require were only used for no-ready feature, when waiting a patch to be merged. Why would we want to run only tests with @require? I have probably missed the discussion on hipchat.

ptnapoleon:
@aboudreault I assume the goal is for when we annotate a test with `@require(ticket)` where the ticket is not a feature in development, but a bug-fix for the feature being tested.

mshuler:
Basically, we discussed using @require a little more liberally, and this would give us a simple way to run with or only the tests with @require. This gives us a jenkins job like http://cassci.datastax.com/view/All_Jobs/job/scratch_mshuler-trunk_dtest-experimental/

mambocab:
High-level: what @mshuler said. More detailed: here's the workflow this supports:

- A test is failing in one way or other -- in some cases, it'll be a flaky test.
- we file a ticket and `@require` the test, pointing to the ticket

At this point, most test runs will run normally, skipping `@require`d tests. However, we'll also have a job that runs daily (probably) that either includes `@require`d tests, or only runs `@require`d tests. This will help us keep an eye on flapping and failing tests without adding failure noise to dev branch builds.

I understand it's a little weird to `@require` a flapping test -- if [flaky](https://github.com/box/flaky) works out, then we won't have to use it that way. But for the moment, this will help us reduce failure noise in dev branches and normal builds.

aboudreault:
Thanks Jim for the clarification. +1 for this -- I agree that the flaky plugin would be in favor in the future.

mshuler:
This also serves as an audit job, since there is a disconnect between fixing a JIRA bug and removing the @requires annotation in dtest. For example, a few tests passed in that run, why are we skipping them (assuming they were features/fixes that aren't just flakey tests)?

mambocab:
Thanks, all. Merging.

PR #366, opened 2015-07-06
2.2 flakes

mambocab:
This deals with a couple flaky tests on 2.2. See the commit messages for more details.

I'm not confident that the new `wait_on_binary_proto` will make the test stop flaking, as I noted in the commit message for that change -- I don't understand the interaction between `wait_for_binary_proto` and `wait_other_notice` well enough to say. However, adding `wait_for_binary_proto` can't hurt, and it might help. We'll figure it out as we keep an eye on flaky dtests in the coming weeks.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -74,6 +74,10 @@ will often need to modify them in some fashion at some later point:
 
         sudo apt-get install python-nose
 
+* flaky
+
+		sudo pip install flaky
+
 * cassandra
 
         cd ~/git/cstar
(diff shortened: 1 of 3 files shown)


aboudreault:
Following your previous PR,  we would use IGNORE_REQUIRE to run flaky tests? But this also mean all other tests with @require will run too. Am I understanding correctly?

mambocab:
Yep, you are. See [my comment over there](https://github.com/riptano/cassandra-dtest/pull/365#issuecomment-119219787) for a more detailed account of the workflow.

aboudreault:
LGTM!

mambocab:
Thanks for the review. Since we're just about ready to integrate `flaky` on Jenkins, though, I've replaced the `@require` with `flaky`. @aboudreault sorry to ask for a second review, but could you have another look?

@mshuler I'll wait for your go-ahead before merging too.

mshuler:
LGTM!

aboudreault:
+1

mambocab:
Ok, looks like cassci is ready, Travis is happy, and so are y'all. Thanks for all the reviews! Merging.

PR #367, opened 2015-07-07
fix sstable_deletion tests

mambocab:
I think I introduced this bug. By default, `block_on_compaction_log` tries to compact the default stress table. This works fine as long as you run it after stress, but in one case I ran it after adding things manually, to a keyspace and table with different names. This caused `sstable_deletion_test` to break everywhere, and I am ashamed.

This fixes that problem by adding `ks` and `table` params to `block_on_compaction_log` while leaving the default behavior the same.

@aboudreault and/or @shawnkumar to review.


aboudreault:
+1, when CI is done.

mambocab:
Thanks, merging.

PR #368, opened 2015-07-07
Rename hinted handoff test

mambocab:
@stef1927, could you please review these docstring and test name changes and make sure they're accurate? My concern is that someone might add hinted handoff tests to this test class without realizing that the hinted handoff feature is older than 3.0.


stef1927:
LGTM, thanks!

PR #369, opened 2015-07-07
data dir/jbod test

shawnkumar:
Please do not merge as feature has not yet been committed

### DIFF
--- a/jbod_test.py
+++ b/jbod_test.py
@@ -3,23 +3,17 @@ from ccmlib.node import Node, NodeError, TimeoutError
 from cassandra import ConsistencyLevel, Unavailable, ReadTimeout
 from cassandra.query import SimpleStatement
 from tools import since
-import subprocess
 import os
-import time
 import shutil
 
 @since('3.0')
 class TestJBOD(Tester):
-
-    def __init__(self, *args, **kwargs):
-        self.ignore_log_patterns = [
-            # This one occurs when trying to send the migration to a
-            # node that hasn't started yet, and when it does, it gets
-            # replayed and everything is fine.
-            r"doesn't exist",
-        ]
-        Tester.__init__(self, *args, **kwargs)
-        self.allow_log_errors = True
+    """
+    Cassandra generally doesn't have an idea of the state of disks.
+    In these tests we use multiple data directories to simulate JBOD events.
+    Ie. adding a disk, a disk failing, keys being distributed and so on.
+    @jira_ticket CASSANDRA-6696
+    """
 
     def verify_keys_test(self):
         """
@@ -35,46 +29,42 @@ class TestJBOD(Tester):
         cluster.set_configuration_options({'disk_failure_policy':'best_effort'})
 
         cluster.set_data_dirs(['data1', 'data2'])
-        cluster.start()
+        cluster.start(wait_for_binary_proto=True)
 
         cursor = self.patient_cql_connection(node1)
-        ksq = "CREATE KEYSPACE test WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':3}"
-        cfq = "CREATE TABLE test.jtest(key int primary key, val int);"
-        cursor.execute(ksq)
-        cursor.execute(cfq)
-
+        cursor.execute("CREATE KEYSPACE test WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':3}")
+        cursor.execute("CREATE TABLE test.jtest(key int primary key, val int);")
+        
         #insert a bunch of keys
         numkeys=100
-        for x in range(0, numkeys):
-            insq = "INSERT INTO test.jtest(key,val) VALUES ({key}, {key})".format(key=str(x))
-            cursor.execute(insq)
+        for x in range(numkeys):
+            cursor.execute("INSERT INTO test.jtest(key,val) VALUES ({key}, {key})".format(key=x))
         node1.flush()
 
         initial_sstables = [None]*numkeys
         #figure out which dir each key is in
-        for x in range(0, numkeys):
+        for x in range(numkeys):
             sstable_path = node1.nodetool('getsstables test jtest ' + str(x), capture_output=True)
             if "data1" in sstable_path[0]:
                 initial_sstables[x] = "data1"
             elif "data2" in sstable_path[0]:
                 initial_sstables[x] = "data2"
             else:
-                self.fail(sstable_path)
+                self.fail("Invalid sstable path:" + sstable_path)
 
         #update added keys to ensure that keys remain on the same disk
-        for x in range(0, numkeys):
-            insq = "INSERT INTO test.jtest(key,val) VALUES ({key}, 66)".format(key=str(x))
-            cursor.execute(insq)
+        for x in range(numkeys):
+            cursor.execute("INSERT INTO test.jtest(key,val) VALUES ({key}, 66)".format(key=x))
         node1.flush()
 
         #check that keys are on same disks as prior
         final_sstables = [None]*numkeys
-        for x in range(0, numkeys):
+        for x in range(numkeys):
             sstable_paths = node1.nodetool('getsstables test jtest ' + str(x), capture_output=True)
             for path in sstable_paths:
                 debug(initial_sstables[x])
                 debug(path)
-                if path != "":
+                if path:
                     self.assertTrue(initial_sstables[x] in path)
     
     def add_disk_test(self):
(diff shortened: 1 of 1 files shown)


mambocab:
Did you want a review on this? I feel like you answered this question in our meeting, so sorry for asking you to repeat yourself.

I can give a general review, though I think @aboudreault could speak more directly to JBOD stuff.

shawnkumar:
Sure - a general review is mostly what I'm looking for.

mambocab:
:+1: cool cool, thanks

aboudreault:
will take a look at this too, but tomorrow morning.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+import subprocess
+import os
+import time
+import shutil
+
+@since('3.0')
+class TestJBOD(Tester):
+
+    def __init__(self, *args, **kwargs):
+        self.ignore_log_patterns = [
```
mambocab: This can be set at the class level, as seen [here](https://github.com/riptano/cassandra-dtest/blob/52501f0377a3e4a058eea45df2f792dcf6b8b677/offline_tools_test.py#L13). Same goes for `allow_log_errors` below. That way, you can get rid of this `__init__ method entirely, since it'd basically be a noop at that point.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+            # This one occurs when trying to send the migration to a
+            # node that hasn't started yet, and when it does, it gets
+            # replayed and everything is fine.
+            r"doesn't exist",
+        ]
+        Tester.__init__(self, *args, **kwargs)
+        self.allow_log_errors = True
```
mambocab: This should probably be replaced with more patterns in `ignore_log_patterns`; ignoring all errors could let some failures slip past us.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+from cassandra import ConsistencyLevel, Unavailable, ReadTimeout
+from cassandra.query import SimpleStatement
+from tools import since
+import subprocess
+import os
+import time
+import shutil
+
+@since('3.0')
+class TestJBOD(Tester):
```
mambocab: If there's an associated ticket, could you add a `@jira_ticket` to this class?

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        node1, node2, node3 = cluster.nodelist()
+        cluster.set_configuration_options({'disk_failure_policy':'best_effort'})
+
+        cluster.set_data_dirs(['data1', 'data2'])
+        cluster.start()
+
+        cursor = self.patient_cql_connection(node1)
+        ksq = "CREATE KEYSPACE test WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':3}"
+        cfq = "CREATE TABLE test.jtest(key int primary key, val int);"
+        cursor.execute(ksq)
```
mambocab: These are only used here, right? If so, I'd personally prefer they be inlined, like `cursor.execute("CREATE KEYSPACE...")`. Just less to reason about. This is personal taste, though; if you disagree, go with what you like.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        node1.flush()
+
+        #check that keys are on same disks as prior
+        final_sstables = [None]*numkeys
+        for x in range(0, numkeys):
+            sstable_paths = node1.nodetool('getsstables test jtest ' + str(x), capture_output=True)
+            for path in sstable_paths:
+                debug(initial_sstables[x])
+                debug(path)
+                if path != "":
```
mambocab: This can be written `if path:`.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        numkeys=100
+        for x in range(0, numkeys):
+            insq = SimpleStatement("INSERT INTO test.jtest(key,val) VALUES ({key}, {key})".format(key=str(x)), consistency_level=ConsistencyLevel.ALL)
+            cursor.execute(insq)
+        node1.flush()
+
+        cluster.drain()
+        cluster.stop(gently=False)
+        cluster.set_data_dirs(['data1', 'data2', 'data3'])
+        shutil.rmtree(os.path.join(node1.get_path(), 'data1'))
```
mambocab: I'll go ahead and thank you on Josh's behalf for using os-agnostic path and `rm` stuff here.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        Add additional disk to yaml and remove one existing disk (ie remove directory) and start cluster.
+        Ensure that data files properly distributed across disks.      
+        """
+        cluster = self.cluster
+        cluster.populate(3)
+        node1, node2, node3 = cluster.nodelist()
+        cluster.set_data_dirs(['data1', 'data2'])
+        cluster.set_configuration_options({'disk_failure_policy':'best_effort'})
+
+        cluster.start()
```
mambocab: This might not wind up being a problem in practice, but here and elsewhere, could you use `wait_for_binary_proto`? It can't hurt and will head off potential problems on slow machines.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+@since('3.0')
+class TestJBOD(Tester):
+
+    def __init__(self, *args, **kwargs):
+        self.ignore_log_patterns = [
+            # This one occurs when trying to send the migration to a
+            # node that hasn't started yet, and when it does, it gets
+            # replayed and everything is fine.
+            r"doesn't exist",
```
mambocab: There seem to be a lot of errors that contain `doesn't exist`:

https://github.com/apache/cassandra/search?utf8=%E2%9C%93&q=%22doesn%27t+exist%22

Could you make this pattern match more precisely?

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        Tester.__init__(self, *args, **kwargs)
+        self.allow_log_errors = True
+
+    def verify_keys_test(self):
+        """
+        Test to see that all keys are on the same disk:
+        Start up 3 node cluster, with 2 data disks.
+        Write a number of partitions to cluster, remembering keys.
+        Use getsstables to see which sstables each key is on. 
+        Check that all keys are on same disk.
```
mambocab: nice markup, thanks.

mambocab:
The fixes that are important to me are adding Doxygen metadata where applicable, adding `wait_for_binary_proto`, addressing my comments about the log error ignoring and pattern matching, and making Travis happy. My other comments are just personal preferences/Python nerdery, not shared style guidelines, so feel free to ignore those if you disagree with my opinions there.

One other thing that would be helpful is to explain the relationship between the JBOD-aware features and `set_data_dirs` -- the connection isn't totally obvious, at least not to me. Maybe add a short explanation to the class-level docstring about using it to simulate JBOD (if that's the case) and explain why that's valid for testing in an environment where you don't actually have a JBOD array.

With those changes, I'm +1, but would like to see @aboudreault's review too, since I know nothing about JBOD stuff.

shawnkumar:
Thanks, will get those changes in. Regarding the JBOD stuff I can see how it'd be confusing - cassandra doesn't actually know anything about the state of disks - only data directories. So I'm pretty much using data directories to simulate JBOD - might be worth changing some of the naming here too.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        cursor = self.patient_cql_connection(node1)
+        ksq = "CREATE KEYSPACE test WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':3}"
+        cfq = "CREATE TABLE test.jtest(key int primary key, val int);"
+        cursor.execute(ksq)
+        cursor.execute(cfq)
+
+        #insert a bunch of keys
+        numkeys=100
+        for x in range(0, numkeys):
+            insq = "INSERT INTO test.jtest(key,val) VALUES ({key}, {key})".format(key=str(x))
```
aboudreault: It is not mandatory to cast your int argument in format(), it will be handled internally. ".format(key=x) this should be good.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+        cursor = self.patient_cql_connection(node1)
+        ksq = "CREATE KEYSPACE test WITH REPLICATION = {'class':'SimpleStrategy', 'replication_factor':3}"
+        cfq = "CREATE TABLE test.jtest(key int primary key, val int);"
+        cursor.execute(ksq)
+        cursor.execute(cfq)
+
+        #insert a bunch of keys
+        numkeys=100
+        for x in range(0, numkeys):
```
aboudreault: range starts at 0 by default, you can omit it if you want: range(numkeys)

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+                initial_sstables[x] = "data2"
+            else:
+                self.fail(sstable_path)
+
+        #update added keys to ensure that keys remain on the same disk
+        for x in range(0, numkeys):
+            insq = "INSERT INTO test.jtest(key,val) VALUES ({key}, 66)".format(key=str(x))
+            cursor.execute(insq)
+        node1.flush()
+
```
aboudreault: Are we sure the compaction won't be a problem here? Maybe we could disableautocompaction for that table?

jbod_test.py
aboudreault: Should we raise an error if the path is "". I don't think the test should pass if so.

jbod_test.py
```diff
@@ -0,0 +1,125 @@
...
+        numkeys=100
+        for x in range(0, numkeys):
+            insq = SimpleStatement("SELECT val FROM test.jtest WHERE key={key};".format(key=str(x)), consistency_level=ConsistencyLevel.ALL)
+            try:
+                res = cursor.execute(insq)
+                debug(res)
+                self.assertEqual(res[0][0], x)
+            except Exception, e:
+                self.fail("Failed during key verification: " + str(e))
```
aboudreault: Before verifying that keys can be fetched, I think I would  confirm that all keys from node1.data1 are impossible to fetch with an exclusive connection to node1. What do you think?

aboudreault:
LGTM, with comments addressed and linter fixed. Thanks Shawn!

mambocab:
Tracking internally with CSTAR-441.

ptnapoleon:
We already have a test suite for this. Let's just close this PR

mambocab:
:+1: Closing.

PR #370, opened 2015-07-07
Flaky repair test

mambocab:
One of the repair tests is flaky on 2.2 (see [this jira ticket](https://issues.apache.org/jira/browse/CASSANDRA-9752)). This decorates it as such.

Unfortunately, it also seems to fail hard on trunk. If we merge this PR, it will be run twice on trunk until it's fixed. Are we ok with that? This is marked with `@attr(long)`, so I take it it's a long-running test. I think @mshuler is the person to answer that question.

### DIFF
--- a/incremental_repair_test.py
+++ b/incremental_repair_test.py
@@ -1,13 +1,15 @@
-from dtest import Tester, debug
-from tools import insert_c1c2, since
+import os
+import time
+from re import findall
+
 from cassandra import ConsistencyLevel
 from ccmlib.node import Node
-from re import findall
-import time
-import os
-from assertions import assert_one, assert_almost_equal
 from nose.plugins.attrib import attr
+
+from assertions import assert_almost_equal, assert_one
+from dtest import Tester, debug
 from flaky import flaky
+from tools import insert_c1c2, since
 
 
 @since('2.1')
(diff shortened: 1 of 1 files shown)


aboudreault:
I don't think we are ok with that. This test takes like 40 minutes. The attribute long has been added for that: basically, to not run it everytime and only running it in a long dtest project (never been created I think)

mshuler:
Yes, multiple_subsequent_repair_test is a long one and has been pretty irregular. Running it twice will suck, since it already takes 30 minutes or so to run, but we've also talked about excluding "long" and creating a separate "long" test run job. It's not the end of the world and we can modify jobs if it really causes pain, but at the same time, this is a good intermittent test that may help us see the usefulness of flaky usage. :)

mambocab:
I think I'm +0.5 with @mshuler -- this is sort of a worst-case test for retrying flapping tests, so we can try it out and see how bad it hurts.

@aboudreault are you convinced that it's worth a try, or are you still -1? I can totally see your perspective and could be convinced either way.

aboudreault:
+0.67, let's give it a try to see if it at least fix that flaky test. We can also test that in a scratch-test project that only run *long* test

mshuler:
I think we might also see what the top few longest running tests are and
decide if we might annotate more long tests to see what sort of time we
might save moving them to a separate job. It's probably a good time to do
that. :)

mambocab:
Looks like between the three of us we're +2.17. I think that's enough for a merge? That's enough for a merge. Merging.

PR #371, opened 2015-07-08
Tests for CASSANDRA-7392

stef1927:
Some tests to check that we can abort queries when they are already in progress, CASSANDRA-7392.

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -7,14 +7,14 @@ from cassandra import ConsistencyLevel, InvalidRequest
 from cassandra.protocol import ProtocolException
 from cassandra.query import SimpleStatement
 
-from assertions import assert_invalid, assert_one
+from assertions import assert_invalid, assert_one, assert_unavailable
 from dtest import Tester, canReuseCluster, freshCluster
 from thrift_bindings.v22.ttypes import \
     ConsistencyLevel as ThriftConsistencyLevel
 from thrift_bindings.v22.ttypes import (CfDef, Column, ColumnOrSuperColumn,
                                         Mutation)
 from thrift_tests import get_thrift_client
-from tools import rows_to_list, since
+from tools import rows_to_list, since, require, debug
 
 
 class CQLTester(Tester):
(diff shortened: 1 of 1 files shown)


cql_tests.py
```diff
@@ -394,3 +394,96 @@ def range_slice_test(self):
...
+@since('3.0')
+@require("7392")
+class AbortedQueriesTester(CQLTester):
+    """
+    @jira_ticket CASSANDRA-7392
+    Test that we can abort long queries that are already in progress
+    """
+    def local_query_test(self):
+        """
+        Abort a query running on the local coordinator node
```
mambocab: This docstring implies (to me) that this tests the correctness of some command that aborts the query. Upon further inspection, though, it looks like the query aborts because it takes longer than 1 second? If that's the case, maybe rephrase this to something like "Test that reads that take longer than `read_request_timeout_in_ms` time out".

cql_tests.py
```diff
@@ -394,3 +394,96 @@ def range_slice_test(self):
...
+        """
+        cluster = self.cluster
+        cluster.set_configuration_options(values={'read_request_timeout_in_ms': 1000})
+
+        cluster.populate(1).start(wait_for_binary_proto=True, jvm_args=["-Dcassandra.test.read_iteration_delay_ms=100"])
```
mambocab: I don't know what this JVM argument does; could you please add a comment to explain?

cql_tests.py
```diff
@@ -394,3 +394,96 @@ def range_slice_test(self):
...
+                id int PRIMARY KEY,
+                col int,
+                val text
+            );
+        """)
+
+        cursor.execute("CREATE INDEX ON test3 (col)")
+
+        for i in xrange(500):
+            cursor.execute("INSERT INTO test3 (id, col, val) VALUES ({}, {}, 'foo')".format(i, i / 10))
```
mambocab: We plan to transition to Python 3 in the near future. When we do, the return type of `/` will change from `int` to `float`:

```python
$ python -c 'print type(3/1)'
<type 'int'>
$ python3 -c 'print(type(3/1))'
<class 'float'>
```

It looks like you actually want floor division here, so could you please change this to `i // 10` so the behavior will be consistent between Python 2 and 3? I missed this incompatibility in `CONTRIBUTING.md`; I'll add that now.

mambocab:
If it's worth testing that queries on indexes time out properly, it's probably also worth checking that it works correctly on materialized views. If you agree, and if you can add a test for that at this point, could you please do so with `@require('6477')`? If you're not able right now, this should be on @aboudreault's radar so he can implement the test (see #360).

stef1927:
Thanks for the comments, see second commit. 

I cannot easily create a materialized view test right now as it would be too time consuming to rebase the 7392 branch on to the 6477 branch. Later, when both features are on the same branch (trunk, 6477 or 7392 - whichever comes first) we can then write a test. I've added a stub with a @require(6477) that fails so we don't forget to add this test once MV are committed.

mambocab:
Perfect, thanks! Merging.

PR #372, opened 2015-07-08
adds information on py3 division to CONTRIBUTING

mambocab:
This is follow up for [a comment I left on a PR today](https://github.com/riptano/cassandra-dtest/pull/371#discussion_r34169469). Related to #84.


mambocab:
@aboudreault This is low-priority, but small; when you get a chance could you please review?

aboudreault:
+1 Jim. I think we should create a ticket to also include absolute_import and unicode_literals. Thoughts?

mambocab:
As in, "new files should use absolute_imports and unicode_literals where applicable"? I'd be +1 on that. Any reason not to tack it on to this PR?

aboudreault:
yep. We can add a note in this PR for sure. The issue is more for all existing files. (When we have time to do it and test properly, so mostly after 3.0 release?)

mambocab:
I've pushed up a note about them -- it's pretty minimal. I'll add a note to the Python 3 issue. Any objections or clarifications you want, @aboudreault?

aboudreault:
All good my friend!

mambocab:
:+1: thanks!

PR #374, opened 2015-07-08
adds require to failing tests mentioned in 9722

mambocab:
These are failing at the moment on trunk. As discussed previously, let's `@require` it.

This does introduce a problem -- these tests should be run on, e.g. 2.2, but won't be with require. I think it should be pretty easy to add some `@since` machinery to `@require`, so I'll try to add that.


mambocab:
Superceded by #376; closing for now. Will reopen if #376 doesn't get merged.

PR #375, opened 2015-07-08
adding tests demonstrating CASSANDRA-9749

bdeggleston:


mambocab:
Thank you for the tests! Is this PR ready for review?

mambocab:
I'm happy to fix up stylistic errors like the ones Travis is complaining about, but I do have some questions about the content -- I'll make some comments now.

commit_log_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+from dtest import Tester
+from ccmlib.node import Node
+import struct
+
+class TestCommitLogFailurePolicy(Tester):
+
+    def test_bad_crc(self):
+        """
+        if the commit log header crc is doesn't the data, and the commit_failure_policy is stop, C* shouldn't startup
```
mambocab: Here and in the method name, it's not clear what the "crc" is.

I'm not sure what "is doesn't the data" is supposed to mean.

commit_log_test.py
```diff
@@ -0,0 +1,181 @@
...
+        self.create_ks(cursor, 'ks', 1)
+        cursor.execute("CREATE TABLE ks.tbl (k INT PRIMARY KEY, v INT)")
+
+        for i in range(10):
+            cursor.execute("INSERT INTO ks.tbl (k, v) VALUES ({0}, {0})".format(i))
+
+        results = cursor.execute("SELECT * FROM ks.tbl")
+        self.assertEqual(len(results), 10)
+
+        time.sleep(1)
```
mambocab: What is this sleep for? Is there some condition you can wait on? For instance, you can watch for log messages on a particular node using a node's `watch_log_for` method.

commit_log_test.py
```diff
@@ -0,0 +1,181 @@
...
+
+class TestCommitLogFailurePolicy(Tester):
+
+    def test_bad_crc(self):
+        """
+        if the commit log header crc is doesn't the data, and the commit_failure_policy is stop, C* shouldn't startup
+        """
+        self.cluster.populate(nodes=1)
+        node = self.cluster.nodelist()[0]
+        assert isinstance(node, Node)
```
mambocab: This shouldn't be necessary. What problem was it preventing for you?

commit_log_test.py
```diff
@@ -0,0 +1,181 @@
...
+                f.seek(crc_pos)
+                crc = struct.unpack('>i', f.read(4))[0]
+                self.assertEqual(crc, 123456)
+
+        # try starting normally
+        try:
+            node.start(wait_for_binary_proto=True)
+        except:
+            # ok
+            return
```
mambocab: So the test ends without checking the data if the node fails to start? Why is that ok? and why does the node sometimes fail to start?

mambocab: Ah, I think I understand a little better -- we want C* to fail to start. I'm not sure why the assertions below exist, then. And to check that a particular exception is thrown, use

```
with self.assertRaises(ExceptionType):
    # code that should raise an exception
```

commit_log_test.py
```diff
@@ -0,0 +1,181 @@
...
+        self.create_ks(cursor, 'ks1', 1)
+        cursor.execute("CREATE TABLE ks1.tbl (k INT PRIMARY KEY, v INT)")
+
+        for i in range(10):
+            cursor.execute("INSERT INTO ks1.tbl (k, v) VALUES ({0}, {0})".format(i))
+
+        results = cursor.execute("SELECT * FROM ks1.tbl")
+        self.assertEqual(len(results), 10)
+
+        time.sleep(1)
```
mambocab: As above, this sleep should either be commented or replaced.

commit_log_test.py
bdeggleston: asserting the type hints the type information to intellij, which helps when working with unfamiliar classes. I'll remove it

commit_log_test.py
bdeggleston: Right. The assertions below are mainly there for demonstration purposes of CASSANDRA-9749. They show that even though C* does start up, the data we inserted is no longer there.

mambocab: Gotcha, thanks for the confirmation. @ptnapoleon could you have a look at this? In the long run, we just want to assert that the `start` call raises an exception, but until CASSANDRA-9749 is closed, these extra assertions are useful. Should we wait to merge until the ticket is closed and we can get rid of the extra assertions?

mambocab:
Thanks for addressing my comments; the changes are looking good. We'll figure out if we want to merge now or once the Jira is closed.

commit_log_test.py
mambocab: How about we get rid of the code below this point in this test and merge, now that CASSANDRA-9749 is resolved?

ptnapoleon:
@mambocab 9749 was committed. Should we merge this?

mambocab:
@ptnapoleon I think the thing to do is to assert that the exception we expect is raised as per [my comment](https://github.com/riptano/cassandra-dtest/pull/375#discussion_r35253827), and remove the extraneous stuff at the end. Does that sound reasonable?

ptnapoleon:
Yes, it does.

mambocab:
Closing in favor of #546, which uses platform-agnostic path-building code and asserts that the node fails to start.

PR #376, opened 2015-07-09
Require with version

mambocab:
This changes the `require` decorator so you can skipping the `require`d test only after a particular version. So, imagine:

- A commit is merged to 2.2 and all greater versions, causing a test to fail or flap.
- File Jira ticket number 42 to request the bug be fixed.
- We want to run the test on versions before 2.2, but skip it on later versions 2.2.

The changes in this PR allow us to use `@require(42, broken_in='2.2')` to tag the test's failure as dependent on `CASSANDRA-42` and skips the test on all versions after 2.2.

This also changes the `since` decorator to use a more standard design. This allows a class to be `since`d multiple times, which is more likely to come up now because the new `require` feature uses `since` internally.

My testing of this basically consisted of playing with it, mostly using the tests in `validation.py` as a testbed. It :tada: :balloon: works on my machine :balloon: :tada:, but I'd appreciate it if whoever reviews this could run it locally and confirm it behaves as they expect.

@aboudreault and/or @shawnkumar, can you review?

This also adds some uses of the new `require` feature, superceding #374 -- see the 3rd commit.

### DIFF
--- a/thrift_tests.py
+++ b/thrift_tests.py
@@ -20,7 +20,7 @@ from thrift_bindings.v22.Cassandra import (CfDef, Column, ColumnDef,
                                            Mutation, NotFoundException,
                                            SlicePredicate, SliceRange,
                                            SuperColumn)
-from tools import since
+from tools import since, require
 
 
 def get_thrift_client(host='127.0.0.1', port=9160):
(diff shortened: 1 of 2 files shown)


ptnapoleon:
+1. Ran it locally as well, and it is working for me.

mambocab:
awesome, thank you @ptnapoleon. merging.

PR #381, opened 2015-07-12
fix "Invalid yaml. Please remove properties [memtable_allocation_type…

snazy:
…] from your cassandra.yaml"

### DIFF
--- a/secondary_indexes_test.py
+++ b/secondary_indexes_test.py
@@ -621,6 +621,8 @@ class TestUpgradeSecondaryIndexes(Tester):
 
         # Forcing cluster version on purpose
         cluster.set_install_dir(version="2.0.12")
+        if "memtable_allocation_type" in cluster._config_options:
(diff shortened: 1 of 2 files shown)


ptnapoleon:
In which versions does this option belong in the yaml? It's unclear why it needs set like this in these tests.

snazy:
2.0 doesn't know about ``memtable_allocation_type``. ccm tests that. But for some tests, the cluster object is initialized with the version to test (e.g. 2.1, 2.2) - so that config option makes it into the yaml, which is bad for 2.0-to-X upgrade tests.
The fix is about to remove that option for 2.0 (if you start the dtest with ``OFFHEAP_MEMTABLES=yes CASSANDRA_DIR=/Users/snazy/devel/cassandra/2.1 nosetests``).

ptnapoleon:
+1 then.

PR #382, opened 2015-07-12
Fix nodetool invocation + node-live recognition

snazy:


ptnapoleon:
Are these the correct options for all current C* versions?

mambocab:
```
node2.start(wait_for_binary_proto=True, wait_other_notice=True)
node1.watch_log_for_alive(node2)
```

Isn't the `watch_for_alive` line a noop when it happens right after `start(... wait_other_notice=True)`?

ptnapoleon:
Yes, it should be. The wait_other_notice should be doing the same thing.

snazy:
Repair option: trying ``nodetool repair -dc dc1,dc2`` gives an error (tries with 2.1+). ``nodetool repair -dc dc1 -dc dc2`` works.

``node1.watch_log_for_alive(node2)`` explicitly greps the log file for ``"%s.* now UP" % node.address()`` - it didn't work for me without that test. Or better: the test was flappy (tending to fail more often). Maybe related to that some sleeps were removed recently?

ptnapoleon:
I don't object to the redundancy if it is causing some empirical increase in stability. Just means there's something wrong with one of those methods. Seems like we're +1 now, right @mambocab ?

mambocab:
Yep, +1.

mambocab:
Merging. Thanks, @snazy!

PR #383, opened 2015-07-13
Use python driver metadata api

ptnapoleon:
Removes the existing FIXME. Review, but don't merge. I'm checking with Aleksey to make sure I'm fixing this the way he wanted before I do the rest of the tests. It seems wrong since it's more verbose, and I'm still specifying the keyspace/table names.

### DIFF
--- a/configuration_test.py
+++ b/configuration_test.py
@@ -96,12 +96,18 @@ class TestConfiguration(Tester):
         result = cursor.cluster.metadata.keyspaces['ks'].tables['test_table'].as_cql_query()
         # Now extract the param list
         params = ''
-        if 'sstable_compression' in result:
-            params = result
 
-        assert params is not '', "Looking for the string 'sstable_compression', but could not find it."
+        if self.cluster.version() < '3.0':
+            if 'sstable_compression' in result:
+                params = result
+        else:
+            if 'compression' in result:
+                params = result
 
-        chunk_length = int(re.search("chunk_length_kb.*?:.*?'(\d*?)'", result).groups()[0])
+        assert params is not '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result)
+
+        chunk_string = "chunk_length_kb" if self.cluster.version() < '3.0' else "chunk_length_in_kb"
+        chunk_length = int(re.search("{chunk}.*?:.*?'(\d*?)'".format(chunk=chunk_string), result).groups()[0])
 
         assert chunk_length == value, "Expected chunk_length: %s.  We got: %s" % (value, chunk_length)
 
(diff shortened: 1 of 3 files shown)


ptnapoleon:
@mambocab This is ready for review now.

cqlsh_tests/cqlsh_tests.py
```diff
@@ -1289,14 +1289,7 @@ def test_drop_table(self):
...
 
-        # FIXME: use python-driver metadata API
-        if self.cluster.version() >= '3.0':
-            query = "SELECT table_name FROM system_schema.tables WHERE keyspace_name='ks';"
-        else:
-            query = "SELECT columnfamily_name FROM system.schema_columnfamilies WHERE keyspace_name='ks';"
-
-        result = rows_to_list(self.cursor.execute(query))
-        self.assertEqual([], result)
+        self.assertEqual({}, self.cursor.cluster.metadata.keyspaces['ks'].tables)
```
mambocab: I don't care that this is a dictionary. I'd rather do

```
self.assertFalse(self.cursor.cluster.metadata.keyspaces['ks'].tables)
```
 
or

```
self.assertEqual(0, len(self.cursor.cluster.metadata.keyspaces['ks'].tables))
```

configuration_test.py
```diff
@@ -93,23 +93,15 @@ def new_commitlog_cluster_node():
...
-        rows = cursor.execute(describe_table_query)
-        results = rows[0]
+        result = cursor.cluster.metadata.keyspaces['ks'].tables['test_table'].as_cql_query()
         # Now extract the param list
         params = ''
-        for result in results:
-            if 'sstable_compression' in str(result):
-                params = result
+        if 'sstable_compression' in result:
+            params = result
```
mambocab: Heck yes. I really like how you've condensed this.

cqlsh_tests/cqlsh_tests.py
ptnapoleon: The latter is better, and more consistent with how I do it elsewhere. I think I only did it like this to match the old code.

mambocab:
I'm running the changed tests now locally; I'll report back when they're done. In terms of style stuff, though, looks great. Make Travis happy and I'm happy.

ptnapoleon:
Travis is now happy. ST3 was indenting silly. Tests pass locally for me. I'll wait for you to repro, then merge.

mambocab:
There are a few failures when I run this branch:

https://gist.github.com/mambocab/4d2eb751c97cc989a533

but I don't think any are related to these changes. If you agree, feel free to merge.

PR #385, opened 2015-07-13
looks for correct error on map key search

mambocab:
This has been [failing consistently on cassci](http://cassci.datastax.com/view/trunk/job/trunk_dtest/lastCompletedBuild/testReport/secondary_indexes_test/TestSecondaryIndexesOnCollections/test_map_indexes/history/). This is because one of the asserts was expecting an old error message. This updates the error message for that assert and consolidates some logic with a similar assert nearby.


ptnapoleon:
+1

mambocab:
Thanks!

PR #387, opened 2015-07-13
Fix commitlog tests

mambocab:
Not ready to merge; requesting review.

This PR fixes `default_segment_size_test` and `small_segment_size_test` as described in [this Jira comment](https://issues.apache.org/jira/browse/CASSANDRA-9717?focusedCommentId=14622145&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14622145). It also has a tentative (not working yet :frowning:) change to the `_commitlog_test` method -- the goal is to allow the final segment of compressed commitlogs to be smaller than expected. Unfortunately, this does not seem to fix the test for the small, compressed commitlog -- smaller commitlogs are encountered before the final one.

@blambov Could you please review these tests, and help me figure out where I went wrong special-casing the test for the small, compressed commitlog? Thank you.

### DIFF
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -11,7 +11,7 @@ import ccmlib
 from ccmlib.common import is_win
 from assertions import assert_almost_equal, assert_none, assert_one
 from dtest import Tester, debug
-from tools import require, since
+from tools import since
 
 
 class TestCommitLog(Tester):
@@ -151,7 +151,6 @@ class TestCommitLog(Tester):
                 self.node1.stress(['write', 'n=1M', '-rate', 'threads=25'], stdout=devnull, stderr=subprocess.STDOUT)
 
     @since('2.1')
-    @require(9717, broken_in='3.0')
     def default_segment_size_test(self):
         """ Test default commitlog_segment_size_in_mb (32MB) """
 
@@ -159,7 +158,6 @@ class TestCommitLog(Tester):
         self._commitlog_test(32, 64, 2, files_error=0.5)
 
     @since('2.1')
-    @require(9717, broken_in='3.0')
     def small_segment_size_test(self):
         """ Test a small commitlog_segment_size_in_mb (5MB) """
         segment_size_in_mb = 5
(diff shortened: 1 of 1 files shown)


blambov:
The first two commits are fine.

On the last one, could we try sorting the log list returned by _get_commitlog_files? I'm not sure the files are guaranteed to be listed in the right order.
What kind of difference in size are you getting?

mambocab:
I've seen it produce sizes like 2291640, where the test expects 3670016.

You're probably right; it's probably a sort order problem. Is there a way to determine which is the last segment? and if not, I was thinking this could be tested by allowing at most one segment to be smaller. Does that sound like a reasonable test for correctness?

aweisberg:
Ensuring all but the one segment are within the expected range sounds fine. I would also assert that the sum of all the segments is within some expected range as well.

mambocab:
I've pushed up some new changes. Requesting review from @aweisberg and/or @blambov.

The changes I want the most scrutiny on:

- https://github.com/riptano/cassandra-dtest/pull/387/files#diff-8e7e61c5fac01771e96022884e936371R125 -- Is having this secondary assertion ok for compressed commitlogs? At least in 3.0, they are consistently much smaller than the expected size range.
- https://github.com/riptano/cassandra-dtest/pull/387/files#diff-8e7e61c5fac01771e96022884e936371R186 -- Is this version logic ok? Again, in 3.0, everything is so small as to be out of range of our `assert_almost_equal`

@ptnapoleon Could you have a look at this too? I know these tests are kind of inconsistent, but I've been looking at them on and off for days, so I'd like someone with some distance to see if they make enough sense.

ptnapoleon:
They make sense to me.

mambocab:
Review from Ariel here: https://issues.apache.org/jira/browse/CASSANDRA-9717?focusedCommentId=14644886&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14644886

Merging.

PR #388, opened 2015-07-14
Modified existing test to make it more likely to reproduce CASSANDRA-9460

stef1927:
I've added a loop to try and catch the NPE of CASSANDRA-9460 more frequently.


ptnapoleon:
If we're creating the table in each loop iteration, but only calling truncate, won't that throw an error?

stef1927:
My bad, the create_cf call should have been out of the loop. I was experimenting with creating new tables for each iteration rather than truncating, and forgot to move it back. Force-pushed the fix.

ptnapoleon:
Looks good to me then, +1.

Issue #393, opened 2015-07-14
fix snapshot_test on trunk


mambocab:
A number of the snapshot tests fail:

- snapshot_test.TestArchiveCommitlog.dont_test_archive_commitlog
- snapshot_test.TestArchiveCommitlog.test_archive_commitlog
- snapshot_test.TestArchiveCommitlog.test_archive_commitlog_point_in_time
- snapshot_test.TestArchiveCommitlog.test_archive_commitlog_point_in_time_with_active_commitlog
- snapshot_test.TestArchiveCommitlog.test_archive_commitlog_with_active_commitlog

I've tracked down one source of failure:

https://github.com/riptano/cassandra-dtest/compare/master...mambocab:fix-snapshot-tests

but I'm not familiar with snapshotting or with this test. If someone familiar with them has the time, I'm sure you could fix this quicker than I can. I can take it on if nobody gets to it, but go for it if you like.

mshuler:
Seems this is working in trunk, so I'm going to remove the requires to see what happens.
http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/18/testReport/snapshot_test/TestArchiveCommitlog/

PR #394, opened 2015-07-14
Skip snapshot tests

mambocab:
... until we can fix them with a solution to #393.


mambocab:
Just made this PR to check in with Travis. Travis is happy, merging.

PR #395, opened 2015-07-15
fix cqlsh describe round trip test

mambocab:
Found this while reviewing for [CASSANDRA-9703](https://issues.apache.org/jira/browse/CASSANDRA-9703). Looks like this test should be passing everywhere now, for some reason.


mambocab:
Since this is a trivial change and Travis is happy: merging.

PR #396, opened 2015-07-15
switch decorating order so attr works

mambocab:
With the existing implementation, plain `@require`d tests (with no
`broken_in`) don't get tagged with the `required` attribute. This
switches the order of the application of `skip` and `attr` to fix that.


mambocab:
Never mind, this works already.

PR #398, opened 2015-07-16
test fixes

tjake:
Needed for 9795

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -2,9 +2,11 @@
 import codecs
 import csv
 import datetime
+import time
 import locale
 import os
 import sys
+import platform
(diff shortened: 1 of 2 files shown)


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -2,9 +2,11 @@
 import codecs
 import csv
 import datetime
+import time
 import locale
 import os
 import sys
+import platform
```
ptnapoleon: You're not using either of these two new imports.

ptnapoleon:
Also, travis is complaining about several whitespace violations, but those aren't supercritical. I can fix them after the merge if you don't have time

tjake:
fixed

ptnapoleon:
Merging, then fixing in new PR

Issue #389, opened 2015-07-14
How is replica_availability_test related to CASSANDRA-8640


mambocab:
This is low-priority; just metadata bookkeeping stuff. [CASSANDRA-8640](https://issues.apache.org/jira/browse/CASSANDRA-8640) has @ptnapoleon tagged as tester, so he might be the person to look at this.

How is https://github.com/riptano/cassandra-dtest/blob/master/paxos_tests.py#L33 related to it? doesn't make much sense to me.

ptnapoleon:
In 8640, we were failing LWT queries if all nodes were not up. We should only need quorum nodes. Replica_availability_test checks that the query succeeds with 2/3 nodes live, but fails with 1/3.

PR #399, opened 2015-07-16
documents purpose of replica_availability_test

mambocab:
addresses #389. Can you review, @ptnapoleon?


ptnapoleon:
I would specify that it should work with quorum, rather than the more ambiguous "without all nodes up"

mambocab:
@ptnapoleon thank you. Does it look good now?

ptnapoleon:
+1

mambocab:
Thanks, merging.

PR #400, opened 2015-07-17
Clean up Jake's PR

ptnapoleon:
@mambocab to review.

Also fixes Travis complaints


tjake:
Looks good, there are other instances of this we should fix (I saw some in cqlsh_tests as well)

ptnapoleon:
Yes, but `delete=False` isn't set there.

tjake:
I'm saying we need to put delete=False everywhere

mambocab:
The explicit `unlink`s make me sad, but if that's what fixes it, that's what fixes it. LGTM.

Issue #204, opened 2015-03-24
collapse verbose *_connection methods down to one with boolean flags


knifewine:
For methods: cql_connection, exclusive_cql_connection, patient_cql_connection, and patient_exclusive_cql_connection can all be collapsed down to a single method that takes booleans for "patient" and "exclusive".

Reasonable defaults would be patient=True and exclusive=False of course.

Then we just need to mass update tests to invoke the method the correct way.

aboudreault:
+1, I remember that discussion on HC. Good for the ticket so we don't forget.

PR #403, opened 2015-07-20
Change all instances of cursor to session

### LINKED ISSUE #402: make "connection" name consistent in codebase and across tests
We've got variables and methods with a soup of variations of session/connection/cursor.

We should try and settle on a single name, preferably the same one used in the cassandra python driver, and make it consistent across dtest.

I think some of this is vestigial from dbapi2 which offered somewhat different abstraction.

ptnapoleon:
Fixes #402.

Cassci results here: http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/connection-rename-dtest/1/testReport/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,9 @@ Contributions will be evaluated by PEP8. We now strictly enforce compliance, via
 
 We do not enforce import sorting, but if you choose to organize imports by some convention, use the `isort` tool (`pip install isort`).
 
+Please use `session`, and not `cursor` when naming your connection variables, to match the style preferred by the DataStax Python Driver, which is how these tests
+connect to C*.
+
 ## Doxygen Docstrings
 
 We are now colocating our test plans directly with the source code. We have decided to do so in a manner compatible with Doxygen, to turn the test plans into easily navigated HTML. Please view the following list of tags, as well as an example test. While this full list of tags is available for use, there is no need to use every tag for a given test. The description and since fields should be included, but most others should only be used when appropriate. The test plan will live in a comment block below the test method declaration.
--- a/assertions.py
+++ b/assertions.py
@@ -27,21 +27,21 @@ def assert_invalid(session, query, matching=None, expected=InvalidRequest):
         if matching is not None:
             assert re.search(matching, msg), "Error message does not contain " + matching + " (error = " + msg + ")"
 
-def assert_one(cursor, query, expected, cl=ConsistencyLevel.ONE):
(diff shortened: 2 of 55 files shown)


ptnapoleon:
@knifewine to review.

@aboudreault or @mambocab to look as well

aboudreault:
+1 for sure on this. "session" is a better name.

ptnapoleon:
Justification for new name: http://datastax.github.io/python-driver/api/cassandra/cluster.html#cassandra.cluster.Session

ptnapoleon:
CONTRIBUTING.md also altered to reflect this new, hard requirement.

mambocab:
In principle, I love it. However, it looks like there are some problems with the bucketizer right now, so some of the tests aren't getting run:

``` bash
$ nosetests --collect-only --with-id -vv --test-select-config conf/trunk_test-select.cfg 2>&1 |  wc -l
728
```

vs. the `502 + 165 + 2 = 669` that actually got run on CassCI. It's up to you if you want to merge now and deal with any potential fallout once the bucketizer starts working correctly on trunk, but if it were me I'd wait until I could run all the tests on trunk.

ptnapoleon:
I'm aware the auth tests didn't run. I'll run them locally?

ptnapoleon:
Also, I'll add a second commit to fix the travis complaints

mambocab:
Sounds good -- if you run the tests that didn't run locally, this gets an unqualified +1 from me (though, are the auth_tests the only ones that didn't run?). And making Travis happy is a nice onus, thanks for taking that on.

knifewine:
Overall looks good to me. What do you guys think about renaming the connection methods in dtest.py as well (I think they all return session objects), so no reason to cling to "connection" either unless I'm missing something.

ptnapoleon:
I don't mind the current method names. There does seem to be no need for `cql` in the method names anymore, now that the default connection type is native.

knifewine:
+1 with or without the "cql"/"connection" change.

mambocab:
I'm still +1 on this PR without the `cql`/`connection` change, but I do think that we should remove `cql` and change `connection` to `session` at some point. We could do that as part of tackling #204, but I don't think it's a priority.

ptnapoleon:
auth_tests are passing

ptnapoleon:
@mambocab to review the nine thousand pep8 fixes I just made

mambocab:
I left some comments about reformatting multi-line strings, which I'm ambivalent about. I won't hoot and holler about the existing changes getting merged, but I'm meh on pretty much every solution to it.

But the rest of the PEP8 fixes are gorgeous.

PR #404, opened 2015-07-20
Make cqlsh tests not care about order

dkua:
This PR is related to my work on 9304 here: https://issues.apache.org/jira/browse/CASSANDRA-9304

The changes make the tests not care about the order of the rows in the COPY TO command.


ptnapoleon:
+1, Thanks

PR #406, opened 2015-07-20
Fix PEP-8 Violations

ptnapoleon:
@mambocab to review.

### DIFF
--- a/super_counter_test.py
+++ b/super_counter_test.py
@@ -1,11 +1,10 @@
 import time
 
 from dtest import Tester, debug
-
 from thrift_tests import get_thrift_client
 
 from cql.cassandra.ttypes import CfDef, ColumnParent, CounterColumn, \
-        ConsistencyLevel, ColumnPath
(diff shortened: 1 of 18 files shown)


ptnapoleon:
I'm aware of the failing Travis complaing, and it is okay. Just an existing 600 char line that I didn't have the stamina to fix

super_counter_test.py
```diff
@@ -1,11 +1,10 @@
 import time
 
 from dtest import Tester, debug
-
 from thrift_tests import get_thrift_client
 
 from cql.cassandra.ttypes import CfDef, ColumnParent, CounterColumn, \
-        ConsistencyLevel, ColumnPath
+                                 ConsistencyLevel, ColumnPath
```
mambocab: Not super important now, but at some point these should be changed from line continuations to this style:

```python
from module import (this_thing, that_thing,
                    another_thing)
```

I believe this can be automated with `isort`.

mambocab:
Go for it. Thanks!

PR #409, opened 2015-07-21
6177 Removing Sleep from Dtests

steveandwang:
Added functions that help to remove sleep from dtests.


dtest.py
```diff
@@ -167,6 +168,34 @@ def modify_log(self, cluster):
...
+            while not moved:
+                after = jmx.read_attribute(mbean, attributes[2])
+                if before == after:
+                    moved = True
+                before = after
+                time.sleep(0.2)
+
+    # Determines when node.compact is still running.
+    def compaction(self, node, time):
+        node.compaction(time)
```
ptnapoleon: What function is this calling? I don't see anything in ccmlib

dtest.py
```diff
@@ -167,6 +168,34 @@ def modify_log(self, cluster):
...
         self.var_trace(cluster)
 
+    # Pauses flow of the test until a certain message is seen in the log.
+    # Here are a few:
+    # CQL command with Create TABLE - "Initializing"
+    # CQL command with Create KEYSPACE - "Create new Keyspace"
+    # Move - "MOVING"
+    def wait_till_msg(self, node, first_fun, msg, *args, **kwargs):
+        first_fun(*args, **kwargs)
+        while not node.watch_log_for(msg):
```
ptnapoleon: Don't you need to mark the log prior to running `first_fun` then resume from that mark?

ptnapoleon:
The Travis complaints do appear related to your changes, please fix those as well.

ptnapoleon:
Can you reformat the in-line comments for the methods, so they use docstrings that come after the header, instead of the current format?

dtest.py
```diff
@@ -167,6 +169,36 @@ def modify_log(self, cluster):
...
+    # Pauses flow of the test until a certain message is seen in the log.
+    # Here are a few:
+    # CQL command with Create TABLE - "Initializing"
+    # CQL command with Create KEYSPACE - "Create new Keyspace"
+    # Move - "MOVING"
+    def wait_till_msg(self, node, first_fun, msg, *args, **kwargs):
+        mark = node.mark_log()
+        first_fun(*args, **kwargs)
+        while not node.watch_log_for(msg, from_mark=mark):
+            continue
```
ptnapoleon: I'm confused by this logic here, what is the goal?

PR #410, opened 2015-07-22
Add linter_check.sh

aboudreault:
A convenient script to ensure our changes will pass on travis.

### DIFF
--- a/linter_check.sh
+++ b/linter_check.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# linter_check ensures your changes will pass on travis.
+# Requires pep8 and flake8: pip install pep8 flake8
+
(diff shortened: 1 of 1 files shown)


linter_check.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# linter_check ensures your changes will pass on travis.
+# Requires pep3 and flake3: pip install pep8 flake8
```
ptnapoleon: s/pep3/pep8

aboudreault: haha!

ptnapoleon:
I like it. Did not run locally to verify correctness

linter_check.sh
ptnapoleon: Oh also flake3. What keyboard do you have that 3 and 8 are near each other?

aboudreault: no idea what happened... both keys are not near at all :)

PR #412, opened 2015-07-22
import require correctly

mambocab:
`require` doesn't live in `dtest`. This makes `offline_tools_test` import it correctly.


ptnapoleon:
+1

PR #416, opened 2015-07-24
Fix user_function_test

snazy:


snazy:
C* issue https://issues.apache.org/jira/browse/CASSANDRA-9887

PR #417, opened 2015-07-24
patch for CASSANDRA-9889

snazy:
https://issues.apache.org/jira/browse/CASSANDRA-9889

PR #418, opened 2015-07-24
Revert "patch for CASSANDRA-9889"

ptnapoleon:
Reverts riptano/cassandra-dtest#417


ptnapoleon:
I jumped the gun on the original merge. Fixed

Issue #421, opened 2015-07-27
fix hanging paxos test


mambocab:
Last we ran `paxos_tests.py:TestPaxos.contention_test_multi_iterations` on CassCI, it hung periodically ([example CassCI run](http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/10/consoleFull)). The test itself is fine at time of writing, so we should fix this so we can run the test on trunk.

PR #422, opened 2015-07-27
Unskip paxos contention test

mambocab:
This un-`@require`s the contention test in `paxos_tests.py` -- it passes locally, and it looks like the underlying bug is fixed (see the discussion [in CASSANDRA-9764](https://issues.apache.org/jira/browse/CASSANDRA-9764).

Unfortunately, the test has hung periodically on CassCI; I've filed #421 to track progress on fixing that. Until then, it's skipped.

This PR includes the usual isort cleanup as well.


ptnapoleon:
+1

PR #423, opened 2015-07-28
For CASSANDRA-9741, reduce number of client threads to avoid overload…

aweisberg:
…ing the ccm cluster and generating errors that cause the test to flap


ptnapoleon:
+1

PR #424, opened 2015-07-29
Fix for bootstrap tests on Jenkins

stef1927:
Some of the tests I added for CASSANDRA-9765 pass locally but fail on Jenkins. I'm guessing the timeout is too short so I reverted to the default timeout.


ptnapoleon:
Seems reasonable, +1.

PR #425, opened 2015-07-29
add tests for enable/disable compaction

krummas:


compaction_test.py
```diff
@@ -256,6 +256,97 @@ def skip_if_no_major_compaction(self):
...
             self.skipTest('major compaction not implemented for LCS in this version of Cassandra')
 
+    def disable_autocompaction_nodetool_test(self):
+        cluster = self.cluster
+        cluster.populate(1).start(wait_for_binary_proto=True)
+        [node] = cluster.nodelist()
+        session = self.patient_cql_connection(node)
+        self.create_ks(session, 'ks', 1)
+        j = 0
+        session.execute('CREATE TABLE to_disable_%d (id int PRIMARY KEY, d TEXT) WITH compaction = {\'class\':\'%s\'}'%(j, self.strategy));
```
mambocab: I'm happy to fix it up for this PR, but in the future, could you please use strings' `.format` method for string formatting instead of `%`, as noted in CONTRIBUTING?

compaction_test.py
```diff
@@ -256,6 +256,97 @@ def skip_if_no_major_compaction(self):
         if self.cluster.version() < '2.2' and self.strategy == 'LeveledCompactionStrategy':
             self.skipTest('major compaction not implemented for LCS in this version of Cassandra')
 
+    def disable_autocompaction_nodetool_test(self):
+        cluster = self.cluster
+        cluster.populate(1).start(wait_for_binary_proto=True)
+        [node] = cluster.nodelist()
+        session = self.patient_cql_connection(node)
+        self.create_ks(session, 'ks', 1)
+        j = 0
```
mambocab: I don't understand the purpose of `j`. Looks like it is incremented at the end of the test and never used again?

compaction_test.py
```diff
@@ -256,6 +256,97 @@ def skip_if_no_major_compaction(self):
...
+        session.execute('CREATE TABLE to_disable_%d (id int PRIMARY KEY, d TEXT) WITH compaction = {\'class\':\'%s\'}'%(j, self.strategy));
+        node.nodetool('disableautocompaction ks to_disable_%d'%j)
+        for i in range(1000):
+            session.execute('insert into to_disable_%d (id, d) values (%d, \'%s\')'%(j, i, 'hello'*100))
+            if i % 100 == 0:
+                node.flush()
+        assert len(node.grep_log('Compacting.+to_disable_%d'%j)) == 0, 'Found compaction log items for %s'%self.strategy
```
mambocab: Could you please change the `assert` statements to use `self.assert*` (in this case, `self.assertEqual`)? The `.assert*` methods have better error reporting.

compaction_test.py
```diff
@@ -256,6 +256,97 @@ def skip_if_no_major_compaction(self):
...
+        node.nodetool('enableautocompaction ks to_disable_%d'%j)
+        time.sleep(2)
```
mambocab: If there's something you can wait on, like a line in a log, could you please use that instead of a `sleep`? Or, if sleeping is the only practical way to do this, could you please add a comment noting what it's for?

compaction_test.py
```diff
@@ -256,6 +256,97 @@ def skip_if_no_major_compaction(self):
...
+            if i % 100 == 0:
+                node.flush()
+        assert len(node.grep_log('Compacting.+to_disable_%d'%j)) == 0, 'Found compaction log items for %s'%self.strategy
+        mark = node.mark_log()
+        # should still be disabled after restart:
+        node.stop()
+        node.start()
+        session = self.patient_cql_connection(node)
+        session.execute("use ks")
+        node.watch_log_for("Starting listening for CQL clients", from_mark=mark)
```
mambocab: `Node#start` with the `wait_for_binary_proto` argument uses the `from_mark` logic you've used here, so you can get rid of this logic by using that kwarg.

mambocab:
Done with initial review. With the changes I've noted in comments, I'll be +1.

These tests are very useful, thank you.

mambocab:
Oh, and could you add docstrings to each of the new methods to explain the purpose of each test? Something short is fine, e.g.

```python
    def disable_autocompaction_schema_test(self):
        """
        Test that autocompaction disabling settings in table schema are respected. 
        """
```

krummas:
merged with the comments fixed, thanks!

mambocab:
Great, thank you!

PR #427, opened 2015-07-29
added internode ssl test and util for generating jvm keystores - CASS…

nutbunnies:
…ANDRA-9884

### DIFF
--- a/internode_ssl_test.py
+++ b/internode_ssl_test.py
@@ -1,6 +1,7 @@
 from dtest import Tester, debug
 from tools import generate_ssl_stores, putget
 
+
 class TestInternodeSSL(Tester):
 
     def __init__(self, *args, **kwargs):
@@ -20,4 +21,3 @@ class TestInternodeSSL(Tester):
         self.create_ks(session, 'ks', 3)
         self.create_cf(session, 'cf', compression=None)
         putget(cluster, session)
-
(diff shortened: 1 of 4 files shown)


internode_ssl_test.py
```diff
@@ -0,0 +1,23 @@
+from dtest import Tester, debug
+from tools import generate_ssl_stores, putget
+
+class TestInternodeSSL(Tester):
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
+
+    def putget_with_internode_ssl_test(self):
+        """ Simple putget test with internode ssl enabled - https://issues.apache.org/jira/browse/CASSANDRA-9884 """
```
ptnapoleon: Can you reformat this? `"""` on their own lines, and instead of the full url, just have a line `@jira_ticket CASSANDRA-9884`

tools.py
```diff
@@ -173,6 +174,38 @@ def replace_in_file(filepath, search_replacements):
...
+
+    Args:
+        base_dir (str):  directory where keystore.jks, truststore.jks and ccm_node.cer will be placed
+
+        passphrase (Optional[str]): currently ccm expects a passphrase of 'cassandra' so it's the default but it can be
+            overridden for failure testing
+    Returns:
+        None
+    Raises:
+        CalledProcessError: If the keytool fails during any step
```
ptnapoleon: While I love (and secretly prefer) this formatting of the docstring, we use Doxygen to generate markup, so if you could switch to the format described in `CONTRIBUTING.md`, that would be appreciated.

Well written though

ptnapoleon:
I'm +1, everything looks good to me.

PR #428, opened 2015-07-31
For CASSANDRA-9868, test is sensitive to filesystem iteration order. …

aweisberg:
…Fix broken list comprehension.


ptnapoleon:
+1

PR #430, opened 2015-07-31
Unrequire paging ttl test

mambocab:
This test has been passing for a while: http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/22/testReport/paging_test/TestPagingWithDeletions/test_ttl_deletions/history/ so, gonna un-require it.

So, this test works by:

1. inserting data with a 3-second TTL in a loop
- checking that the data is there (new with this PR)
- waiting 5 seconds
- checking that no data is there

I'm concerned that, if there's a long delay between steps 1 and 2, or if the test is running on a slow machine, TTL'd data may have already expired before its presence is checked for. Is this a realistic concern, or is 3 seconds plenty of time?

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -1467,7 +1467,6 @@ class TestPagingWithDeletions(BasePagingTester, PageAssertionMixin):
         self.check_all_paging_results(expected_data, 8,
                                       [25, 25, 25, 25, 25, 25, 25, 25])
 
-    @require(9831, broken_in='2.2')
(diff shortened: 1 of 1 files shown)


ptnapoleon:
There's no way we won't get that done in 5 seconds, even on a slow machine. It would indicate to me that something is wrong. So I'm +1

mambocab:
Thanks for the review/confirmation. Merging.

PR #431, opened 2015-07-31
Unrequire scrub test

mambocab:
This does the mechanical work of un`@requiring` `test_scrub_collections_table`. I've run the dtests over the C* versions where it was fixed, and it looks like these stopped failing [at b70f7ea](https://github.com/apache/cassandra/commit/b70f7ea0ce27b5defa0a7773d448732364e7aee0). I can't figure out why; anyone have ideas?

### DIFF
--- a/scrub_test.py
+++ b/scrub_test.py
@@ -6,7 +6,7 @@ import uuid
 
 from ccmlib import common
 from dtest import Tester, debug
-from tools import since, require
+from tools import since
 import time
 
 KEYSPACE = 'ks'
@@ -254,7 +254,6 @@ class TestScrubIndexes(TestHelper):
         users = self.query_users(session)
         self.assertEqual(initial_users, users)
 
-    @require('9814', broken_in='3.0')
     def test_scrub_collections_table(self):
         cluster = self.cluster
         cluster.populate(1).start()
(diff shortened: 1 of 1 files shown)


mambocab:
@pauloricardomg This test of `nodetool scrub` started succeeding with your recent changes around deleting snapshots on startup. Do you have any idea why? The changes don't seem related as far as we can tell.

pauloricardomg:
after a quick glance I didn't find any relationship between my commit and this test, which is really strange. do you know why it was failing before? does it pass only by applying this snapshot cleanup patch?
the only possible reason I can think of, is that some garbage snapshots were being left from previous tests that are now being cleaned.. but I think ccm already takes care of isolating test directories, so don't think this could be a problem.

ptnapoleon:
Bump.

mambocab:
Well, looks like I dealt with this here: https://github.com/riptano/cassandra-dtest/pull/510

Are we comfortable letting sleeping dogs lie and not investigating this mysterious success further?

ptnapoleon:
If this has already been merged elsewhere, no need to keep it open.

mambocab:
Filed [CASSANDRA-10308](https://issues.apache.org/jira/browse/CASSANDRA-10308) to investigate this.

PR #433, opened 2015-07-31
Upgrade improvements

knifewine:
This is still kinda a WIP while I see how it behaves on jenkins, but would love feedback in the meantime.

The basic gist of things here is to write/read and increment/read at quorum during the (rolling) upgrade process to make sure clusters can keep serving quorum during an upgrade.

### DIFF
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -188,8 +188,7 @@ def data_writer(tester, to_verify_queue, verification_done_queue, rewrite_probab
                     # we wanted a re-write but the re-writable queue was empty. oh well.
                     pass
 
-            if not key:
-                key = uuid.uuid4()
+            key = key or uuid.uuid4()
 
             val = uuid.uuid4()
 
@@ -289,8 +288,7 @@ def counter_incrementer(tester, to_verify_queue, verification_done_queue, rewrit
                     # we wanted a re-write but the re-writable queue was empty. oh well.
                     pass
 
-            if not key:
-                key = uuid.uuid4()
+            key = key or uuid.uuid4()
 
             session.execute(prepared, (key))
 
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -24,8 +24,8 @@ from tools import generate_ssl_stores, new_node
 # Used to build upgrade path(s) for tests. Some tests will go from start to finish,
 # other tests will focus on single upgrades from UPGRADE_PATH[n] to UPGRADE_PATH[n+1]
 
-TRUNK_VER = (3, 0)
-DEFAULT_PATH = [(1, 2), (2, 0), (2, 1), (2, 2), TRUNK_VER]
+TRUNK_VER = (3, 1)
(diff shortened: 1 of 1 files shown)


knifewine:
pinging @ptnapoleon @mambocab @aboudreault to bring out the red markers.

upgrade_through_versions_test.py
```diff
@@ -139,15 +149,215 @@ def switch_jdks(version):
...
+            key = None
+
+            if (rewrite_probability > 0) and (random.randint(0, 100) <= rewrite_probability):
+                try:
+                    key = verification_done_queue.get_nowait()
+                except Empty:
+                    # we wanted a re-write but the re-writable queue was empty. oh well.
+                    pass
+
+            if not key:
```
mambocab: Teeny Python nicety: this can be more succinctly written `key = key or uuid.uuid4()`. (from a type nerd perspective this is kinda gross because `type(x or y) == bool` is not always true, but `bool(x or y) == bool(x) or bool(y)` so whatever i guess. Anyway, if you don't think it's gross, you can do `key = key or uuid.uuid4()`.)

ptnapoleon: TIL, cool trick.

upgrade_through_versions_test.py
```diff
@@ -225,33 +444,51 @@ def upgrade_scenario(self, populate=True, create_schema=True, mixed_version=Fals
...
 
-                    self.upgrade_to_version(tag, mixed_version=True, nodes=(node,))
+            # upgrade through versions
+            for tag in self.test_versions[1:]:
+                for num, node in enumerate(self.cluster.nodelist()):
+                    # sleep (sigh) because driver needs extra time to keep up with topo and make quorum possible
+                    # this is ok, because a real world upgrade would proceed much slower than this programmatic one
+                    # additionally this should provide more time for timeouts and other issues to crop up as well, which we could
+                    # possibly "speed past" in an overly fast upgrade test
+                    time.sleep(60)
```
ptnapoleon: 1 minute sleep, so gross. I assume you checked, but is there really no way for us to call a driver method that blocks until it's caught up?

ptnapoleon:
At the high level, this is really clean, and easy to follow. I haven't found any little logic errors anywhere either. I still maintain long-term, CVH is better suited for this, but this goes a long way towards making these dtest-based upgrade tests more robust. I'm +1

upgrade_through_versions_test.py
knifewine: I actually didn't pursue a driver call. I *think* part of the issue here is one node still coming up while another is being dropped. Sometimes quorum fails, other times it doesn't. There could be a legit problem as well, but I doubt that C* is doing something wrong, seems more likely there's just a disagreement between ccm/logs/c* about readiness to handle requests or something.

That aside, if the goal here is to get closer to real world upgrades I don't think we're going to find fast-as-possible programmatic upgrades in the wild, so I feel a blind sleep is actually better suited to mimicking "real world" upgrades (and realistically we should stand up clusters for even longer times because stability of an upgrade is a dimension we're not checking with existing tests).

PR #434, opened 2015-08-01
fixed and optimized sstableofflinerevel and sstablelevelreset dtests

pauloricardomg:
Refactored offline_tools_test.py:TestOfflineTools to create a cluster with only one node and write less keys: tests were taking too long with 3 machines, RF=3 and millions of rows. Essence of tests was maintained while making them faster.

Also, the test sstablelevelreset  was not properly testing this tool, so I fixed that.

### DIFF
--- a/offline_tools_test.py
+++ b/offline_tools_test.py
@@ -93,14 +93,15 @@ class TestOfflineTools(Tester):
     def sstableofflinerelevel_test(self):
         """
         Generate sstables of varying levels.
+        Reset sstables to L0 with sstablelevelreset
         Run sstableofflinerelevel and ensure tables are promoted correctly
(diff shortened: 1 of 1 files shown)


pauloricardomg:
fixed travis checks
context: https://issues.apache.org/jira/browse/CASSANDRA-9942

ptnapoleon:
+1 on these changes, except for one small style nit. Stress in 2.1+ supports a more readable syntax for listing operation count, `n=1M` instead of `n=1000000`, `n=50K` instead of `n=50000`, etc. I'll merge this and make the changes, but just an fyi for the future.

pauloricardomg:
Nice, thanks!

2015-08-03 11:28 GMT-03:00 Philip Thompson <notifications@github.com>:

PR #435, opened 2015-08-03
Test for CASSANDRA-9871

stef1927:
A simple test for CASSANDRA-9871.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -6,7 +6,7 @@ import subprocess
 import tempfile
 import re
 from dtest import Tester, debug
-from tools import new_node, query_c1c2, require, since, KillOnBootstrap, InterruptBootstrap
+from tools import new_node, query_c1c2, since, KillOnBootstrap, InterruptBootstrap
 from assertions import assert_almost_equal
 from ccmlib.node import NodeError
 from cassandra import ConsistencyLevel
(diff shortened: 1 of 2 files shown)


bootstrap_test.py
```diff
@@ -422,3 +422,42 @@ def failed_bootstap_wiped_node_can_join_test(self):
         mark = node2.mark_log()
         node2.start(wait_other_notice=True)
         node2.watch_log_for("JOINING:", from_mark=mark)
+
+    @require("9871")
```
ptnapoleon: No need for this now that it has been committed.

bootstrap_test.py
```diff
@@ -422,3 +422,42 @@ def failed_bootstap_wiped_node_can_join_test(self):
...
+        if version < "2.1":
+            node1.stress(['-n', '10000'])
+        else:
+            node1.stress(['write', 'n=10000', '-rate', 'threads=8'])
+
+        # Stop node 3
+        node3.stop(gently=True)
+
+        # Sleep a bit to let GOSSIP settle
+        time.sleep(2)
```
ptnapoleon: Wouldn't using wait_other_notice=True in the `node3.stop()` call remove the need to let gossip settle?

ptnapoleon:
This test should go in replace_address_test.py, not bootstrap_test.py. Also, any ideas why the existing replace_address tests didn't catch 9871? Was it really just the gently=True stop? That's upsetting.

stef1927:
Moved the test to replace_address_test.py and adapted it to reuse the existing test, which requires indeed gently=True in order to reproduce the problem.

ptnapoleon:
Thanks! +1

PR #439, opened 2015-08-04
CASSANDRA 6717 -- schema migrations - 3.0

nutbunnies:


schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
...
+              """
+
+        session.execute(cql.format(table_name))
+        index_name = cls._table_name_builder("idx_" + table_name_prefix, table_name)
+        session.execute("CREATE INDEX {0} ON {1}( d )".format(index_name, table_name))
+
+    @classmethod
+    def verify_indexes_table(cls, version, keyspace, session, table_name_prefix=""):
+        table_name = cls._table_name_builder(table_name_prefix, "test_indexes")
+        assert 1 == len(session.cluster.metadata.keyspaces[keyspace].indexes)
```
EnigmaCurry: You should use self.assert* methods wherever possible, it makes for the logging messages to be much nicer.

schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
...
+    @classmethod
+    def verify_collection_datatype_table(cls, version, keyspace, session, table_name_prefix=""):
+        table_name = cls._table_name_builder(table_name_prefix, "test_collection_datatypes")
+        meta = session.cluster.metadata.keyspaces[keyspace].tables[table_name]
+        if version > '2.1':
+            assert 13, len(meta.columns)
+        else:
+            assert 7, len(meta.columns)
+
+        assert str(cqltypes.ListType.apply_parameters([cqltypes.Int32Type])) == str(meta.columns['a'].data_type)
```
EnigmaCurry: For these cases where you repeat yourself a lot, you could write a helper:

    from cqltypes import ListType, SetType, MapType, Int32Type, UTFT8Type

    ....
    
    def collectionTypeEqual(collection_type, cql_type, meta_column):
        self.assertEqual(str(collection_type.apply_parameters([cql_type])), str(meta_column.data_type))

    collection_asserts = (
        (ListType, Int32Type, meta.columns['a']),
        (ListType, UTFT8Type, meta.columns['b']),
        (SetType, Int32Type, meta.columns['c']),
        ....
    )

    for thing in collection_asserts:
        collectionTypeEqual(*thing)

schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
...
+        meta = session.cluster.metadata.keyspaces[keyspace].tables[table_name]
+        if version > '2.2':
+            assert 19 == len(meta.columns)
+        else:
+            assert 15 == len(meta.columns)
+
+        assert 1 == len(meta.primary_key)
+        assert 'b' == meta.primary_key[0].name
+
+        assert cqltypes.AsciiType == meta.columns['a'].data_type
```
EnigmaCurry: Another inline helper could clean this up

schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
...
+        assert cqltypes.UUIDType == meta.columns['m'].data_type
+        assert cqltypes.UTF8Type == meta.columns['n'].data_type
+        assert cqltypes.IntegerType == meta.columns['o'].data_type
+        if version > '2.2':
+            assert cqltypes.SimpleDateType == meta.columns['p'].data_type
+            assert cqltypes.ShortType == meta.columns['q'].data_type
+            assert cqltypes.TimeType == meta.columns['r'].data_type
+            assert cqltypes.ByteType == meta.columns['s'].data_type
+
+    @staticmethod
```
mambocab: I don't see a need to make this a static method, or to make the others class methods -- I'd pull these out and just make them top-level functions.

schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
+from dtest import Tester, debug
+from tools import since
+from cassandra import cqltypes
+import re
```
ptnapoleon: Run this through isort, if you could, please.

schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
...
+
+class TestSchemaMetadata(Tester):
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
```
ptnapoleon: If we're not special casing the `__init__`, why bother having it?

schema_metadata_test.py
```diff
@@ -0,0 +1,317 @@
...
+        self.verify_udt_table(self.cluster.version(), 'ks', session)
+
+    # @since('2.2')
+    # def udf_table_test(self):
+    #     cluster = self.cluster
+    #     cluster.populate(1).start()
+    #     session = self.patient_cql_connection(cluster.nodelist()[0])
+    #     self.create_ks(session, 'ks', 1)
+    #     self.establish_udf(session)
+    #     self.establish_udf(session)
```
ptnapoleon: What's going on with this test?

ptnapoleon:
I'm with Jim, why make those class methods?

upgrade_through_versions_test.py
```diff
@@ -325,6 +338,24 @@ def _log_current_ver(self, current_tag):
...
+            debug("**SCHEMA ESTABLISH CALLING: [%s]" % m)
+            getattr(TestSchemaMetadata, m)(self.cluster.version(), session, tag)
+
+    def _check_metadata_schemas(self, version, tag):
+        session = self.patient_cql_connection(self.node2)
+        session.execute('use upgrade')
+        debug("***schema verify*** [%s]" % tag)
+
+        for m in filter(lambda mtd: mtd.startswith('verify_'), dir(TestSchemaMetadata)):
+            debug("**SCHEMA VERIFY CALLING: [%s]" % m)
```
ptnapoleon: Use `"str".format()` syntax instead of `"str" % var`, please

schema_metadata_test.py
nutbunnies: currently.... ignorance of python and reusing test methods,  the goal is to reuse these establish and verify methods in the upgrade test

ptnapoleon: Yeah, then no need to attach them to a class at all. Just pull them out. Put them either as top level functions here, or in a separate, helper file.

upgrade_through_versions_test.py
```diff
@@ -325,6 +338,24 @@ def _log_current_ver(self, current_tag):
...
+            getattr(TestSchemaMetadata, m)(version, 'upgrade', session, tag)
```
mambocab: You were asking about style earlier -- stylistically, this is just fine. Using a list comprehension instead of a `filter(lambda...` is a little more idomatic, but ¯\\\_(ツ)\_/¯. (I don't know if you're familiar with the filter-esque comprehension syntax -- it'd be `method for method in dir(thing) if method.startswith('verify_')`)

As discussed above, these `verify_` helper methods should probably get moved to the top level of a module, but the `dir` trick should work on modules, so that wouldn't change this code much.

ptnapoleon:
The asserts still all need changed as Ryan mentioned in the first review.

nutbunnies:
Yep.  Still trying to get that working.  How do you use assertEqual in plain method that isn't in a class that subclasses from TestCase?  Do those establish/verify methods need to get put into class?

ptnapoleon:
Ahh yes, that would be impossible. So if you're doing assertions in those methods, you have to use the non unit test ones.

ptnapoleon:
In which case, I'm not sure which of the two evils is better: To put them into the class when really they don't belong there, or to use the bad assertions. @EnigmaCurry will know

EnigmaCurry:
That's a shame, I can't see a way to use any of the unittest asserts in a static way. Does this feel too obtuse? : 

    class AssertLogger(unittest.TestCase):
        def runTest():
            pass

    def just_a_function():
        a = AssertLogger()
        a.assertEqual(4,5)

raises:

    AssertionError: '4' != '5'

ptnapoleon:
I'm -1 on that. 

My preference would be to create a subclass of TestCase, and move the functions in there. Then take all the other subclassses where you're going to use these functions, and have them subclass the new class, instead of TestCase

EnigmaCurry:
yea, keeping it in an instance of TestCase definitely simplifies it.

nutbunnies:
Python Noob Response:  Wouldn't that cause some odd coupling.  It would need to subclass Tester which subclasses TestCase so that it could be used as the parent class in TestUpgradeThroughVersions.   It seems very awkward, but might simplify the method signatures as it could pull some of that information from self.

knifewine:
Why not just have the verify methods take a Tester object as an argument and call the assertions via that object? It's pretty minor coupling imo, and stating it in the docstring should be plenty to keep others from getting confused. A little coupling makes the world go round...

ptnapoleon:
Let's just use the nose methods that Jim suggested.

nutbunnies:
How does it look?  I've changed to the nose.tools implementation of assert_equal.

ptnapoleon:
Asserts look good, I'll start a full review now

ptnapoleon:
I'm +1, this is very thorough. Can you squash these before merging though?

PR #440, opened 2015-08-04
Add test for internode_compression: 'none'

yukim:
CASSANDRA-9884 still needs fix NPE when internode compresssion is off.
Patch to add this scenario.


ptnapoleon:
+1

PR #442, opened 2015-08-06
Test retrieving dates in the far future and far past with cqlsh

pauloricardomg:
Test for https://issues.apache.org/jira/browse/CASSANDRA-10000 (not committed yet).


ptnapoleon:
+1. Merge when C*-10k is ready.

PR #448, opened 2015-08-08
Trying to fix decommissioned_wiped_node_can_join_test on Jenkins

stef1927:
decommissioned_wiped_node_can_join_test : replace SIGTERM with SIGKILL after decommision to make sure process does not linger


ptnapoleon:
+1

PR #449, opened 2015-08-10
Fix for replace_address and concurrent_schema_changes failing tests.

stef1927:
I noticed that the test I've added for CASSANDRA-9871 is not running on Jenkins whilst it did run on my box. The reason is that the version of ccm I had installed was not up-to-date. After updating it I got the same failure and fixed it along with a few more tests in replace_address and one test in concurrent_schema_changes that were failing with the same reason. I also sped up a couple of tests in replace_address, no need to wait for 5 minutes for a binary notification that will never be there.

Finally, I attempted one more time to fix decommissioned_wiped_node_can_join_test() although this is just a guess as this test was fine on my box even after updating ccm.


ptnapoleon:
+1

PR #450, opened 2015-08-10
added compression disable/enable test - CASSANDRA-8384

nutbunnies:
2 questions with this pull request:
- should it verify that the actual sstables on disk are compressed?
- I had a small usability nit -- the enabled key only comes back from the compression metadata call if it's disabled -- so a client that was checking status has an odd check -- should this be brought up on the jira ticket?

### DIFF
--- a/compression_test.py
+++ b/compression_test.py
@@ -7,13 +7,13 @@ class TestCompression(TestHelper):
 
     def _get_compression_type(self, file):
         types = {
-            '00 10': 'NONE',
-            '78 9c': 'DEFLATE'
+            '0010': 'NONE',
+            '789c': 'DEFLATE'
         }
 
         with open(file, 'rb') as fh:
-            file_start = fh.read(4)
-            return types.get(''.join(["%02x " % ord(x) for x in file_start[:2]]).strip(), 'UNKNOWN')
+            file_start = fh.read(2)
+            return types.get(file_start.encode('hex'), 'UNKNOWN')
 
     @since("3.0")
     def disable_compression_cql_test(self):
--- a/compression_test.py
+++ b/compression_test.py
@@ -1,8 +1,19 @@
-from dtest import Tester
+from scrub_test import TestHelper
 from tools import since
+import os
 
 
-class TestCompression(Tester):
+class TestCompression(TestHelper):
+
+    def _get_compression_type(self, file):
+        types = {
+            '00 10': 'NONE',
+            '78 9c': 'DEFLATE'
+        }
+
+        with open(file, 'rb') as fh:
+            file_start = fh.read(4)
+            return types.get(''.join(["%02x " % ord(x) for x in file_start[:2]]).strip(), 'UNKNOWN')
 
     @since("3.0")
     def disable_compression_cql_test(self):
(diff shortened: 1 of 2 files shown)


EnigmaCurry:
I think a compressed table will have a header to the file that indicates it's compressed, so sounds like a quick win to me. I would think that setting should be returned normally too.

EnigmaCurry:
I think that's an easy fix for the devs to return the setting, that sounds familiar (new settings not showing up in DESCRIBE)

nutbunnies:
Is that the file type header check you were thinking of?

EnigmaCurry:
Nice, looks right

compression_test.py
```diff
@@ -0,0 +1,128 @@
+from scrub_test import TestHelper
+from tools import since
+import os
+
+
+class TestCompression(TestHelper):
+
+    def _get_compression_type(self, file):
```
mambocab: This could be a top-level function, since it doesn't actually use `self` in the body. This isn't an integral part of the codebase's style, it's just a personal thing of mine, so only change this if you agree that it's better.

compression_test.py
```diff
@@ -0,0 +1,128 @@
+from scrub_test import TestHelper
+from tools import since
+import os
```
mambocab: If you make any changes before merging, could you isort these?

compression_test.py
```diff
@@ -0,0 +1,128 @@
...
+        [node] = cluster.nodelist()
+
+        session = self.patient_cql_connection(node)
+        self.create_ks(session, 'ks', 1)
+        session.execute("create table disabled_compression_table (id uuid PRIMARY KEY ) WITH compression = {'enabled': false};")
+        meta = session.cluster.metadata.keyspaces['ks'].tables['disabled_compression_table']
+        self.assertEqual('false', meta.options['compression']['enabled'])
+
+        for n in range(0, 100):
+            session.execute("insert into disabled_compression_table (id) values (uuid());")
```
mambocab: I always forget about the CQL `uuid()` function! Thanks for the reminder.

compression_test.py
```diff
@@ -0,0 +1,128 @@
...
+                    'chunk_length_in_kb': 256,
+                    'crc_check_chance': 0.25
+                };
+            """)
+        meta = session.cluster.metadata.keyspaces['ks'].tables['start_enabled_compression_table']
+        self.assertEqual('org.apache.cassandra.io.compress.SnappyCompressor', meta.options['compression']['class'])
+        self.assertEqual('256', meta.options['compression']['chunk_length_in_kb'])
+        self.assertEqual('0.25', meta.options['compression']['crc_check_chance'])
+        session.execute("alter table start_enabled_compression_table with compression = {'enabled': false};")
+        meta = session.cluster.metadata.keyspaces['ks'].tables['start_enabled_compression_table']
```
mambocab: Unless there's something I don't know about `session.cluster.metadata...`, this definition is redundant.

ptnapoleon: You're missing the alter table statement in between the two definitions. That should affect the metadata. But @nutbunnies, you'll probably need to call one of the metadata sync methods, to update the driver's understanding of the metadata. The async isn't going to be fast enough here.

mambocab: Ah -- I was expecting these objects to override `__getitem__` and always be current, but no such luck I suppose.

mambocab:
Looks good! +1.

PR #458, opened 2015-08-12
fix for CASSANDRA-10051

pauloricardomg:
this test assumes the commit log will be recovered on next startup, which was not happening on windows because the node was being shutdown before the next sync period, so I basically changed it to use the batch-sync commit log, which fixed the test
(this test is also dependent on https://github.com/pcmanus/ccm/pull/351 to work on windows)


ptnapoleon:
+1

PR #459, opened 2015-08-12
Added test_commitlog_replay_on_startup

pauloricardomg:
During #CASSANDRA-10051 I needed to investigate a difference in behavior of commit log replay between Windows and Linux, and didn't find any dtest for commit log replay, which is a critical feature of cassandra.

This dtest basically inserts a single row, shuts down the node without flushing, and verify if the inserted row is present in next cluster startup due to commit log replay.

### DIFF
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -1,4 +1,3 @@
-
 import glob
 import os
 import stat
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -1,3 +1,4 @@
+
 import glob
 import os
 import stat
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Do none of the commitlog tests really cover this?

commitlog_test.py
```diff
@@ -1,3 +1,4 @@
+
```
ptnapoleon: Why the new line?

ptnapoleon:
+1

commitlog_test.py
pauloricardomg: mostly to separate testing steps.. but not a big deal, you can remove if you want :)

pauloricardomg:
surprisingly I didn't find any test that cover this.. at least explicitly. maybe some other test covers it indirectly, or maybe I just missed it.

ptnapoleon:
Well, that's upsetting.

PR #460, opened 2015-08-13
Skip hanging tests

mambocab:
These tests hung on a recent CassCI run (I believe [this one](http://cassci.datastax.com/view/trunk/job/trunk_dtest/lastCompletedBuild/testReport/)). Here are the relevant tests in the test nodes' stdout; note no more tests are run after these tests:

https://gist.github.com/mambocab/2ae8539003bee6b4eb7d#file-test_stdout-005-txt-L204

https://gist.github.com/mambocab/2ae8539003bee6b4eb7d#file-test_stdout-008-txt-L2

So, this PR skips them hard.

There's also a bit of import reorganization in this PR :+1:

### DIFF
--- a/incremental_repair_test.py
+++ b/incremental_repair_test.py
@@ -1,14 +1,12 @@
 import os
 import time
 from re import findall
-from unittest import skipIf
+from unittest import skip
 
 from cassandra import ConsistencyLevel
-from flaky import flaky
 from nose.plugins.attrib import attr
 
 from assertions import assert_almost_equal, assert_one
-from ccmlib.common import is_win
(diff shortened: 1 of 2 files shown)


ptnapoleon:
+1. Do we want to to consider moving that long repair test to CVH?

mambocab:
You'd know if it's suited for CVH better than I would. Seems that it's definitely not well-suited (at least as implemented) for dtests -- Michael said it's been a problem as long as it's been in the codebase.

mambocab:
Thanks for having a look; merging.

PR #461, opened 2015-08-13
fetch riptano/master and diff against that

mambocab:
Dunno why I didn't make a PR for this this a week or two ago when I did it, but:

We've been having some weird problems with Travis, where the diffing fails to find the right master branch. It's hard to tell when it's a problem, but this should fix it.

It also has the nice side effect of letting people run Travis on their forks and check this stuff via CI before making a PR. For example, here is a build of a branch on my fork that I deliberately inserted whitespace errors into:

https://travis-ci.org/mambocab/cassandra-dtest/builds/75453490


ptnapoleon:
+1

mambocab:
Thanks for the review; merging.

PR #462, opened 2015-08-13
Further fix cqlsh_copy_tests on Windows

josh-mckenzie:
We need to use NamedTemporaryFile(delete=False) on Windows to prevent
access violations. Also cleaned up CqlshCopyTester.tearDown logic on
Windows and for calling super.tearDown in general.


ptnapoleon:
+1

PR #463, opened 2015-08-13
Make output of assertCsvResultEqual more verbose

josh-mckenzie:


josh-mckenzie:
This came in handy while debugging a test that was failing locally. Figured I'd push it upstream if it fits.

ptnapoleon:
+1

PR #464, opened 2015-08-14
Tests for CASSANDRA-9917

pauloricardomg:
Validation tests for gc_grace_seconds on logged batches and materialized views

### DIFF
--- a/batch_test.py
+++ b/batch_test.py
@@ -1,9 +1,10 @@
 import time
 
 from assertions import assert_invalid, assert_unavailable
-from dtest import Tester
+from dtest import Tester, debug
 from cassandra import ConsistencyLevel, Timeout
 from cassandra.query import SimpleStatement
+from tools import since
 
 class TestBatch(Tester):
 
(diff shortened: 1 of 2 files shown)


batch_test.py
```diff
@@ -50,6 +50,40 @@ def logged_batch_accepts_regular_mutations_test(self):
...
+        """)
+        node1 = self.cluster.nodelist()[0]
+        warning = node1.grep_log("Table 'ks.users' has a configured gc_grace_seconds of 0, "
+                                 "below the recommended minimum of 10800 for use in atomic batches. "
+                                 "This might lead to partially applied batches in the case of failures.")
+        debug(warning)
+        self.assertEquals(1, len(warning), "Cannot find the gc_grace_seconds warning message.")
+
+    @since('3.0')
+    def unlogged_batch_gcgs_below_treshold_should_not_print_warning_test(self):
```
ptnapoleon: Can you fix the typo in these test names? Should be `threshold` not `treshold`

materialized_views_test.py
```diff
@@ -73,6 +73,32 @@ def create_test(self):
...
+        # Now let's drop MV
+        session.execute("DROP MATERIALIZED VIEW ks.users_by_state;")
+
+        # Now we should be able to set the gc_grace_seconds of the base table to 0
+        session.execute("ALTER TABLE users WITH gc_grace_seconds = 0")
+
+        # Now we shouldn't be able to create a new MV on this table
+        assert_invalid(session, "CREATE MATERIALIZED VIEW users_by_state AS "
+                                "SELECT * FROM users WHERE STATE IS NOT NULL AND username IS NOT NULL "
+                                "PRIMARY KEY (state, username)", "Base table 'users' must have gcGraceSeconds"
```
mambocab: Small readability nit: it's hard to tell just by glancing at this that there are multiple string arguments. I might insert line breaks in this `assert_invalid` call something like this:

```python
assert_invalid(session,
               "multi-line "
               "string arg",
               "another multi-line "
               "string arg")
```

So that the string arguments are separated by newlines as well as commas. I might even use named arguments:

```python
assert_invalid(session,
               query="multi-line "
               "string arg",
               matching="another multi-line "
               "string arg")
```

or 

```python
assert_invalid(session,
               query="multi-line "
                     "string arg",
               matching="another multi-line "
                        "string arg")
```

Do what you think is best for readability. In my opinion, you should at least have line breaks after the commas, especially between multiline string literals. If you disagree, no worries.

pauloricardomg:
fixed suggestions. thanks @ptnapoleon and @mambocab

ptnapoleon:
+1

pauloricardomg:
CASSANDRA-9917 wasn't merged yet, won't this cause 3.0 tests to fail on cassci?

ptnapoleon:
You are correct, I was wondering why this hadn't been merged yet.

PR #465, opened 2015-08-14
Add retry method for flaky/slow nodes

tjake:
This helps with tests that fail due to Read/Write timeouts.  We could add an exponential backoff but this will probably help 90%


ptnapoleon:
+1, but can you fix the travis whitespace complaint

PR #466, opened 2015-08-14
Import sleep in replace_address_test

mambocab:
This has been failing on CassCI because it used sleep without importing it. But when I ran it locally, it passed? Spooky. Anyway, this should fix it, and contains the obligatory isort changes.


ptnapoleon:
+1

PR #467, opened 2015-08-14
Ignore whitespace differences on pkey_requirement_test

josh-mckenzie:
Fixes a test failure on Windows from a LF vs. CRLF difference


ptnapoleon:
+1, fix the travis complaint please

josh-mckenzie:
I really ought to be running flake8 locally first. Sorry about that - force pushed w/fix.

PR #468, opened 2015-08-14
fixes occasional failures in simple_rebuild_test

mambocab:
This removes the test's dependence on the outcome of a race. Occasionally, the call to `nodetool rebuild` in the main thread would start before the one in the secondary thread, making the call in the secondary thread fail instead of the one in the main thread. The failure is correct -- a `nodetool rebuild` call should fail if another is running. However, the test only expected failure in the main thread. It led to failures like this:

http://cassci.datastax.com/view/trunk/job/trunk_dtest/470/testReport/junit/rebuild_test/TestRebuild/simple_rebuild_test/

where the command in the thread clearly (correctly) failed, but the test didn't pass. It led to 4 failures over the course of the past few days:

http://cassci.datastax.com/view/trunk/job/trunk_dtest/lastCompletedBuild/testReport/rebuild_test/TestRebuild/simple_rebuild_test/history/

Eliminating the race condition itself is...  probably impossible? Or at least difficult. So, instead, the new logic checks that exactly one of the calls throws the error, rather than requiring one or the other to.

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -73,9 +73,16 @@ class TestRebuild(Tester):
         session.execute("ALTER KEYSPACE ks WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
         session.execute('USE ks')
 
+        self.rebuild_errors = 0
+
         # rebuild dc2 from dc1
         def rebuild():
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

mambocab:
Thanks. Merging.

PR #471, opened 2015-08-18
8099 backwards compat

ptnapoleon:
Merging Tyler's branch from C-9893. I will review this, alongside @mambocab and @knifewine

PR #473, opened 2015-08-18
Backwards compat

ptnapoleon:
Moved PR #471 here so that we can edit the branch more easily. Will set up a cassci job for these tests.

### DIFF
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -1,8 +1,7 @@
 import os
 import time
 
-from dtest import Tester, DEBUG
-
+from dtest import DEBUG, Tester
 
 QUERY_UPGRADED = os.environ.get('QUERY_UPGRADED', 'true').lower() in ('yes', 'true')
 QUERY_OLD = os.environ.get('QUERY_OLD', 'true').lower() in ('yes', 'true')
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -83,13 +83,13 @@ class UpgradeTester(Tester):
 
         # start them again
         if UPGRADE_MODE != "none":
-            node1.set_install_dir(version=self.original_install_dir)
+            node1.set_install_dir(install_dir=self.original_install_dir)
             node1.set_log_level("DEBUG" if DEBUG else "INFO")
             node1.set_configuration_options(values={'internode_compression': 'none'})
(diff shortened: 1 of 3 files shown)


ptnapoleon:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/trunk_dtest_backwards_compat/

mambocab:
I'll deal with trivial fixes like string formatting, unless you're already taking care of that @ptnapoleon

ptnapoleon:
I was not yet, feel free

upgrade_tests/upgrade_base.py
```diff
@@ -0,0 +1,113 @@
...
+        node1 = cluster.nodelist()[0]
+        time.sleep(0.2)
+
+        session = self.patient_cql_connection(node1, protocol_version=protocol_version)
+        if create_keyspace:
+            self.create_ks(session, 'ks', rf)
+
+        return session
+
+    def do_upgrade(self, session):
```
mambocab: If I understand correctly, this method doesn't really do an upgrade -- it makes `session.cluster` a mixed cluster.

I don't think it makes much sense for this to return a list of a session for each node. Leads to code like [this](https://github.com/riptano/cassandra-dtest/pull/473/files#diff-a5ca0889b7206a6067d45983280e993eR198) -- takes some digging to figure out why `do_upgrade` returns something iterable.

So, either the concerns of "upgrading to a mixed cluster" and "getting a session for each node" should be separated, or the method should be renamed to something that reflects its behavior and return value. I favor the former, as I can't think of a name that meets those criteria right now.

mambocab:
There two competing concerns here. 

- Many of these tests are duplicated, with small changes, from existing tests, which introduces a considerable maintenance problem. We should be able to eliminate most or all of the duplication with some inheritance trickery, but doing so will take a lot of manual merge work, especially since the new tests are copied from out-of-date versions of the existing tests.
- [It'd be good to have these running ASAP](https://issues.apache.org/jira/browse/CASSANDRA-9893?focusedCommentId=14700943&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14700943).

So, my plan is to get these tests running as structured, but have an ongoing task to refactor them.

ptnapoleon:
Understood and agreed. +1 to merge.

mambocab:
Thanks for the review; merging.

PR #482, opened 2015-08-19
Exclude 'Unable to delete' on 22 upgrades on Windows

josh-mckenzie:
'Unable to delete' from SSTableDeletingTask is expected on Windows on older versions during upgrades.


mambocab:
Looks fine to me -- All the errors this will catch are logged in the same kind of error:

https://github.com/apache/cassandra/search?utf8=%E2%9C%93&q=%22unable+to+delete%22

Merge away!

PR #483, opened 2015-08-19
make linesep comparison os agnostic in hintedhandoff_test.py'

josh-mckenzie:
For CASSANDRA-10101


mambocab:
LGTM, thank you

ptnapoleon:
Why not use string formatting?

josh-mckenzie:
Force pushed update using string formatting - fair point.

PR #492, opened 2015-08-21
Rebased version of #409

ptnapoleon:
I rebased and squashed Steve's #409, and am re-opening here.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -21,6 +21,7 @@ from cassandra.cluster import Cluster as PyCluster
 from cassandra.cluster import NoHostAvailable
 from cassandra.policies import WhiteListRoundRobinPolicy, RetryPolicy
 from nose.exc import SkipTest
+from jmxutils import JolokiaAgent, make_mbean
 
 from ccmlib.cluster import Cluster
 from ccmlib.cluster_factory import ClusterFactory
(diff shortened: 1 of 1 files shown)


dtest.py
```diff
@@ -220,6 +221,19 @@ def modify_log(self, cluster):
...
+    def delay(self, node):
+        mbean = make_mbean('metrics', 'CommitLog', name='PendingTasks')
+        with JolokiaAgent(node) as jmx:
+            moved = False
+            before = None
+            while not moved:
+                after = jmx.read_attribute(mbean, 'Value')
+                if before == after:
+                    moved = True
+                before = after
```
ptnapoleon: This seems to do the opposite of the comment, right @mambocab ? Also, it seems incredibly susceptible to races

mambocab: Yeah, should be `before != after`.

I'm not positive, but I think that sometimes a commitlog task spawns 1 or more other commitlog tasks, so This could actually get tripped when the number of tasks _increases_ which is not what we want. (I could be thinking of compaction here, not sure.)

And yeah, seems like there are a lot of races this could lose -- e.g. what if all the pending commitlog tasks are finished before this method starts?

ptnapoleon: I'm not even sure where this is supposed to be used. Why aren't we just checking if PendingTasks == 0? It really depends on when/how this is to be used.

PR #493, opened 2015-08-21
Revert "Backwards compat"

ptnapoleon:
Reverts riptano/cassandra-dtest#473

This broke cassci so badly


ptnapoleon:
@mambocab to fix as part of his refactor.

PR #495, opened 2015-08-24
Set hints_directory to be different for every node

stef1927:
I'm not sure why it doesn't fail on Jenkins, but locally on my box, unless I set the hints_directory to be different for each node, two tests in hintedhandoff_test.py fail.


ptnapoleon:
Do you have the most up to date ccm? There was an intermediate commit before full support for 6230 was written, where this issue was happening

stef1927:
Thanks, I've updated ccm and it works now.

PR #497, opened 2015-08-24
Updated single_file_split_test to iterate over run_sstablesplit output instead of redirecting output to file

pauloricardomg:
[ccm PR363](https://github.com/pcmanus/ccm/pull/363) caused a [regression](http://cassci.datastax.com/view/cassandra-3.0/job/cassandra-3.0_dtest/lastCompletedBuild/testReport/sstablesplit_test/TestSSTableSplit/single_file_split_test/history/) on dtest `single_file_split_test`.

The simple fix is to rely on the new output of `run_sstablesplit`, which already exposes the stdout and stderr instead of trying to redirect the output to a temp file.

### DIFF
--- a/sstablesplit_test.py
+++ b/sstablesplit_test.py
@@ -1,7 +1,5 @@
 from __future__ import division
 
-import subprocess
-import tempfile
 import time
 from math import floor
 from os.path import getsize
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

PR #498, opened 2015-08-24
Document skipping v2 test

mambocab:
Adds documentation to complement 76216531bf75d8951e9e7ae102b6935a5e9f3f68. Requesting review from @iamaleksey.

### DIFF
--- a/write_failures_test.py
+++ b/write_failures_test.py
@@ -98,7 +98,7 @@ class TestWriteFailures(Tester):
     @since('2.2')
     def test_mutation_v3(self):
         """
-            A failed mutation at v3 receives a WriteTimeout
+        A failed mutation at v3 receives a WriteTimeout
         """
         self.expected_expt = WriteTimeout
         self.protocol_version = 3
--- a/write_failures_test.py
+++ b/write_failures_test.py
@@ -86,7 +86,10 @@ class TestWriteFailures(Tester):
     @since('2.2', '2.2.X')
     def test_mutation_v2(self):
         """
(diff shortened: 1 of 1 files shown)


iamaleksey:
The protocol was introduced in Cassandra 2.0, not 2.2. Deprecated in 2.2, removed in 3.0.

iamaleksey:
TBH I'm not sure why its minimal version is 2.2. Only now looked at it. That one, and some surrounding tests, too.

v3 was added in 2.1.

mambocab:
Thanks, Aleksey. @stef1927 you wrote these tests and added the original `since` restrictions -- why are these tests skipped pre-2.2?

stef1927:
Initially the `since` restrictions were (sort of) correct. There was a problem with this commit: 1c44d814dc026b3cc05f815c0b5aff9348996ca7 when we moved the 3.0 `since` tags to 2.2, we mistakenly also changed a couple of 2.0 and 2.1 tags. `WriteFailureException` was introduced in 2.2 but the timeouts should be testable before that.

In addtion to `test_mutation_v2` and `_v3`, `test_thrift` can probably also be tested at 2.0.

ptnapoleon:
@mambocab bump.

mambocab:
Well, looks like I dealt with this here: https://github.com/riptano/cassandra-dtest/pull/510

Are we comfortable letting sleeping dogs lie and not investigating this mysterious success further?

mambocab:
Closing in favor of #597.

PR #499, opened 2015-08-24
Concurrent operations

nutbunnies:


concurrent_schema_changes_test.py
```diff
@@ -117,6 +126,226 @@ def validate_schema_consistent(self, node):
...
+        session.cluster.refresh_schema_metadata()
+        index_meta = session.cluster.metadata.keyspaces["lots_o_indexes"].indexes
+        self.validate_schema_consistent(node1)
+        self.validate_schema_consistent(node2)
+        self.assertEqual(10, len(index_meta))
+        for n in range(0, 5):
+            self.assertIn("ix_base_{0}_c1".format(n), index_meta)
+            self.assertIn("ix_base_{0}_c2".format(n), index_meta)
+
+        debug("waiting for indexes to fill in")
```
mambocab: I like this, and it'd be nice if the other `sleep` calls had `debug` statements, or at least comments, explaining why they're there. I know there are a lot of them, so maybe it could all just go into a class docstring or something.

concurrent_schema_changes_test.py
```diff
@@ -117,6 +126,226 @@ def validate_schema_consistent(self, node):
...
+        for (success, result) in results:
+            self.assertTrue(success, "didn't get success on table create: {}".format(result))
+
+        wait(30)
+
+        session.cluster.refresh_schema_metadata()
+        table_meta = session.cluster.metadata.keyspaces["lots_o_alters"].tables
+        column_ct = 0
+        for table in table_meta.values():
+            column_ct += len(table.columns)
```
mambocab: This could be clearer as `column_ct = sum([len(table.columns) for table in table_meta.values()])`.

concurrent_schema_changes_test.py
```diff
@@ -117,6 +126,226 @@ def validate_schema_consistent(self, node):
...
+
+        node1, node2, node3 = cluster.nodelist()
+        session = self.cql_connection(node1)
+        session.execute("create keyspace lots_o_alters WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};")
+        session.execute("use lots_o_alters")
+        for n in range(0, 10):
+            session.execute("create table base_{0} (id uuid primary key)".format(n))
+        wait(5)
+
+        cmds = []
```
mambocab: In general, it's better Python to use list comprehensions when you're builidng a list up from nothing, so something like

```
cmds = ['alter table base_{0} add c_{1} int'.format(randrange(0, 10), n) for n in range(500))]
```

If the `start` argument to `range` is 0, you can omit it; `range(0, 500) == range(500)`.

mambocab: or, rather

```python
cmds = [('alter table base_{0} add c_{1} int'.format(randrange(0, 10), n), ())
        for n in range(500))]
```

(Note that you can freely use newlines as whitespace in comprehensions if the statement gets too long.)

concurrent_schema_changes_test.py
```diff
@@ -117,6 +126,226 @@ def validate_schema_consistent(self, node):
...
+        wait(10)
+        cmds = []
+        for n in range(0, 20):
+            cmds.append(("create table new_table_{0} (id uuid primary key, c1 int, c2 int, c3 int, c4 int);".format(n), ()))
+            for a in range(1, 8):
+                cmds.append(("alter table alter_me_{0} drop s{1};".format(n, a), ()))
+                cmds.append(("alter table alter_me_{0} add c{1} int;".format(n, a), ()))
+                cmds.append(("create index ix_index_me_{0}_c{1} on index_me_{0} (c{1});".format(n, a), ()))
+
+        shuffle(cmds)
```
mambocab: A shuffle here worries me a bit. I think it assumes that any failures this catches will be caused by trying to execute too many commands, but if errors are caused by the _order_ of the commands, this could lead to non-reproducible failures.

The thing to do might be to just seed the RNG with `random.seed` before the `shuffle`.

mambocab:
Seems reasonable to me.

There's some testing in this space that'll be done in CVH, so maybe some add a class- or module-level docstring defining the scope and purpose of these tests. Maybe something worth discussing with @ptnapoleon.

But the tests look good as-is, and could be improved a bit stylistically as noted in my comments inline.

concurrent_schema_changes_test.py
nutbunnies: Agreed.  Python list comprehension syntax has yet to click with me -- cleaning up this and similar code.

concurrent_schema_changes_test.py
nutbunnies: Good catch.  It's pointless as well, since the cmd[] is executed concurrently the order is non-deterministic to begin with no need shuffle.

Another good point about reproducing errors.  All these schema tests are based on timing which outside of a major issue (like the MV one) getting consistent reproducible failures (even the MV failures are different each time but at-least consistently fail) is extremely difficult.  Suggestions?

mambocab: I don't know of a great solution -- I'd ask in chat if you think it's worth pursuing -- but one ok-but-not-great solution we've used is just running the critical part of the test multiple times: https://github.com/riptano/cassandra-dtest/pull/388

mambocab:
Looks good to me, merge away!

PR #500, opened 2015-08-25
Tests for CASSANDRA-10086

stef1927:
Two new commands are added to clqsh by CASSANDRA-10086: CLS and CLEAR and here are the tests for them.

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -3,6 +3,7 @@ import binascii
 import csv
 import datetime
 import os
+import re
 import subprocess
 import sys
 from decimal import Decimal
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1, but @mambocab what do you think?

cqlsh_tests/cqlsh_tests.py
```diff
@@ -1176,6 +1177,48 @@ def test_describe_round_trip(self):
...
+
+            In addition we can move the cursor upper left:
+            Esc[H or Esc[;H
+
+            The escape character code is 27.
+
+            We don't check for moving the cursor though, we only check that
+            there is no error and that at least one of the possible clear
+            screen sequences is contained in the output, via a regular
+            expression that allows any optional digit between '['' and 'H'.
```
mambocab: Do you mean `between '['' and 'J'`?

mambocab: And if valid clear screen sequences are `Esc\[[012]?J`, why match on all valid digits?

mambocab: > And if valid clear screen sequences are Esc\[[012]?J, why match on all valid digits?

Sorry, I wasn't being very precise here. I don't understand why this test checks for `\d` instead of `[012]` -- I think the test should be changed to accept the smaller set of values, or the comment should be changed to explain why we're checking for `\d` when only `[012]` are used in screen-clearing codes.

mambocab:
I think this is looking good, thanks for the test! I don't understand enough about control codes to understand why we can't also grep for `Esc[H` and `Esc[;H`, but I inserted that check locally and it failed, so ¯\\\_(ツ)\_/¯.

The other things I don't understand are addressed inline. Once my comments are addressed let me know and I'll see if I understand better.

stef1927:
The `\d` rather than `[012]` was just to keep the regex more readable and slightly more generic for future proofing. I changed it to `[012]` now.

The reason for not checking on cursor move (`[H`) is that what we are interested in is clearing the screen. It just happens that the Linux clear command returns `\x1b[H\x1b[J`. However I don't know about other platforms and other terminals. The test in his original form was confirmed to work on Windows and Mac OS as well but I did not test myself so I don't know the exact sequence returned and even if I did it could well change depending on the terminal and OS version. I do know however what the current ANSI clear screen escape sequences are, so I limited the check to that.

mambocab:
That reasoning sounds good to me. Merge away!

PR #501, opened 2015-08-25
Test for CASSANDRA-9232

stef1927:
A test to check that we can describe tables whose name is a non-reserved CQL keyword. It requires the patch from CASSANDRA-9232.


ptnapoleon:
+1

PR #502, opened 2015-08-25
compaction conf dictionary vals for cqlsh describe

mambocab:
This changes the order of the values in the `compaction` configuration dictionary. Locally, this fixes this error:

http://cassci.datastax.com/view/trunk/job/trunk_dtest/516/testReport/cqlsh_tests/TestCqlsh/test_describe/


ptnapoleon:
And this passes on every version we test against?

ptnapoleon:
If so, +1

mambocab:
Yep, green across the board. Thanks for the review, merging.

PR #504, opened 2015-08-25
For CASSANDRA-10148, handle Unavailable in addition to Timeout since …

aweisberg:
…both can legitimately occur

### DIFF
--- a/batch_test.py
+++ b/batch_test.py
@@ -191,7 +191,7 @@ class TestBatch(Tester):
                         received_responses, e.received_responses,)
                 assert e.received_responses == received_responses, msg
         except Unavailable as e:
-            if not received_responses is None:
+            if received_responses is not None:
                 msg = "Expecting alive_replicas to be %s, got: %s" % (
                         received_responses, e.alive_replicas,)
                 assert received_responses == e.alive_replicas, msg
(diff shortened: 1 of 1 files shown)


batch_test.py
```diff
@@ -190,6 +190,11 @@ def assert_timedout(self, session, query, cl, acknowledged_by=None,
                 msg = "Expecting received_responses to be %s, got: %s" % (
                         received_responses, e.received_responses,)
                 assert e.received_responses == received_responses, msg
+        except Unavailable as e:
+            if received_responses is not None:
+                msg = "Expecting alive_replicas to be %s, got: %s" % (
```
ptnapoleon: Please change the string formatting to be `"{}".format()` instead of `"%s" % s`

batch_test.py
```diff
@@ -190,6 +190,11 @@ def assert_timedout(self, session, query, cl, acknowledged_by=None,
...
+                        received_responses, e.alive_replicas,)
+                assert received_responses == e.alive_replicas, msg
```
ptnapoleon: I'd prefer if you used the better string formatting method, and the unittest asserts here, but I see you're just matching the existing code. I'll make the changes to both after merge.

ptnapoleon:
+1

stef1927:
I think it's because of the flaky retry policy that was introduced in 1659b2f7ef3e1989ca46b9853a0ea640ec949822 that unavailable exceptions can also occur. An alternative would have been to use `retry_policy=FallthroughRetryPolicy()` when creating the statement.

PR #505, opened 2015-08-25
Use better formatting and asserts in batch_test.py

ptnapoleon:
Also some PEP8 and import changes

@mambocab to look over

I checked, all the tests still pass.


mambocab:
+1, merge away

Issue #486, opened 2015-08-21
upgrade tests need to handle CASSANDRA-10146


knifewine:
currently we default to protocol version 1 everywhere.

we need a little bit of code to choose protocol connection versions from those actually available on version V and V+1. It might be useful to abstract these into two strategies such as "reluctant" and "eager" where we use the minimal version that will work, or the maximal version, respectively.

ptnapoleon:
We have code that does this in `dtest.py`'s `cql_connection`. We should extract that out into its own method and re-use it.

PR #506, opened 2015-08-25
changes upgrade tests to focus in a particular protocol version

knifewine:
This is the simplest way to test specific protocol versions, but does
require parting upgrade testing into several jobs to get full upgrade
coverage from 1.2 up through 3.1 and beyond.

The alternative to this method would require a lot more complexity to
maintain connections at the proper protocol version (particularly with
the concurrent testing now taking place with reads/writes).

### DIFF
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -26,7 +26,16 @@ from tools import generate_ssl_stores, new_node
 # other tests will focus on single upgrades from UPGRADE_PATH[n] to UPGRADE_PATH[n+1]
 
 TRUNK_VER = (3, 1)
-DEFAULT_PATH = [(1, 2), (2, 0), (2, 1), (2, 2), (3, 0), TRUNK_VER]
+
+# maps protocol version to c* version(s)
+PROTOCOL_PATHS = {
+    1: [(1, 2), (2, 0), (2, 1)],  # omitting 2.2; it supports v1, but deprecated
+    2: [(2, 0), (2, 1)],          # omitting 2.2; it supports v2, but deprecated
+    3: [(2, 1), (2, 2), (3, 0), TRUNK_VER],
+    4: [(2, 2), (3, 0), TRUNK_VER]
+}
+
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1 to how you implemented this, but are we sure we're okay with not getting full upgrade tests in?

knifewine:
imo the multi version upgrade tests haven't yielded much when compared to the smaller v1->v2 type tests that are more targeted. But I could be misinformed.

So what we're giving up here is the multi version upgrades covering the full released version span (1.2 up through trunk/3.1), but no coverage of v1->v2 type tests will be lost as long as we maintain a test job for each protocol version, if that makes sense.

ptnapoleon:
Okay, I'm convinced. +1

knifewine:
Oops, just added a commit to actually use the protocol_version specified.

ptnapoleon:
Oh, I didn't realize you had been hard coding it to 1. Why set it like this? I feel like if you don't pass in that kwarg, cql_connection will be choosing the correct version regardless

knifewine:
This way we will get a bit more proto version coverage, rather than the more eager proto version decision made in dtest.py (since 3/4 have some overlap for example). That and if you specify PROTOCOL_VERSION in your env you can be sure that version will be used.

knifewine:
Fixes #486

ptnapoleon:
It doesn't seem like this code properly uses protocols other than 3, unless they're set in the env, no?

knifewine:
Yup, that's the idea here. Jenkins jobs will set their own to test other proto versions.

ptnapoleon:
Well won't the tests that use 1.2 or 2.0 break locally then?

knifewine:
PROTOCOL_VERSION drives the path selection, so you are getting a compatible proto and test path when you choose 1,2,3, or 4. Defaulting to 3 was chosen to get the most Cassandra version coverage.

ptnapoleon:
Oh! I see I missed the logic on line 48. Okay, +1!

PR #508, opened 2015-08-26
Refactor _create_session

### LINKED ISSUE #486: upgrade tests need to handle CASSANDRA-10146

ptnapoleon:
Breakout logic to find highest protocol version

@knifewine you closed #486, but I still think this refactoring needed done, and will be useful. Can you review? Thanks

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -346,6 +346,21 @@ class Tester(TestCase):
             if not is_win():
                 os.symlink(basedir, name)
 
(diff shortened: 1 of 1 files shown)


knifewine:
great idea, +1

PR #510, opened 2015-08-26
Unrequire passing tests

mambocab:
This does some whitespace cleanup, but more importantly, removes `@require`s from tests that pass now.


ptnapoleon:
+1

mambocab:

PR #513, opened 2015-08-26
Fix schema_metadata_test regression for secondary indexes

pauloricardomg:
It seems [support to multiple secondary indexes per column](https://datastax-oss.atlassian.net/browse/PYTHON-372) on the most recent python driver release [broke schema_metadata_test](http://cassci.datastax.com/view/cassandra-3.0/job/cassandra-3.0_dtest/lastCompletedBuild/testReport/schema_metadata_test/history/). This PR fixes the existing dtests to assume there might be more than one index per column.


ptnapoleon:
+1

PR #515, opened 2015-08-27
Ignoring errors before upgrade to 1.2 on Windows (CASSANDRA-10037)

pauloricardomg:
Instead of using the error whitelist approach, the idea is to mark the log for errors just before upgrading to 2.2 on Windows, effectivelly ignoring all errors happening on 2.1 (since we're not interested in those errors for the purpose of this test).

This test depends on this [ccm PR](https://github.com/pcmanus/ccm/pull/370) and approval on [this ticket](https://issues.apache.org/jira/browse/CASSANDRA-10037) before being merged.


ptnapoleon:
I'm +1, but we'll see what Josh says on the ticket.

PR #517, opened 2015-08-27
Tuple 2i test

nutbunnies:


secondary_indexes_test.py
```diff
@@ -271,6 +272,95 @@ def __init__(self, *args, **kwargs):
...
+        for n in range(50):
+            self.assertEqual(8, len(session.execute("select * from simple_with_tuple where single_tuple = ({0});".format(n))))
+            self.assertEqual(8, len(session.execute("select * from simple_with_tuple where double_tuple = ({0},{0});".format(n))))
+            self.assertEqual(8, len(session.execute("select * from simple_with_tuple where triple_tuple = ({0},{0},{0});".format(n))))
+            self.assertEqual(8, len(session.execute("select * from simple_with_tuple where nested_one = ({0},({0},{0}));".format(n))))
+
+        # check if indexes work on mutated data
+        for n in range(5):
+            rows = session.execute("select * from simple_with_tuple where single_tuple = ({0});".format(n))
+            [session.execute("update simple_with_tuple set single_tuple = (-999) where id = {0}".format(row.id)) for row in rows]
```
mambocab: This isn't idiomatic Python unfortunately, and I don't find it very readable. I think `map` is evaluated strictly in Py2, but I know it returns a generator in Py3, so unfortunately we can't use that here. I think the thing to do is either use an explicit `for` loop or use prepared statements and `execute_concurrent`.

secondary_indexes_test.py
```diff
@@ -271,6 +272,95 @@ def __init__(self, *args, **kwargs):
...
+            rows = session.execute("select * from simple_with_tuple where nested_one = ({0},({0},{0}));".format(n))
+            [session.execute("update simple_with_tuple set nested_one = (-999,(-999,-999)) where id = {0}".format(row.id)) for row in rows]
+
+        for n in range(5):
+            self.assertEqual(0, len(session.execute("select * from simple_with_tuple where single_tuple = ({0});".format(n))))
+            self.assertEqual(0, len(session.execute("select * from simple_with_tuple where double_tuple = ({0},{0});".format(n))))
+            self.assertEqual(0, len(session.execute("select * from simple_with_tuple where triple_tuple = ({0},{0},{0});".format(n))))
+            self.assertEqual(0, len(session.execute("select * from simple_with_tuple where nested_one = ({0},({0},{0}));".format(n))))
+
+        self.assertEqual(40, len(session.execute("select * from simple_with_tuple where single_tuple = (-999);".format(n))))
```
mambocab: Doesn't look like these actually format the string -- can you remove the `.format` call?

secondary_indexes_test.py
mambocab: make that `execute_concurrent_with_args`.

mambocab:
Test logic looks good to me.

In a lot of cases I ask people to switch to using prepared statements instead of `.format`, but this is quite readable as is, probably more readable than with prepared statements, because you don't have to refer back to a bunch of prepared statements when reading one of the loops. The downside is that there are multiple sites to change if the schema under test changes. So, I don't have strong feelings, but that's what went through my head when I reviewed this.

Other than my little nits, +1, looks good!

PR #520, opened 2015-08-28
Being more cautious with start and stop to avoid UnavailableException on repair_test

pauloricardomg:
`simple_sequential_repair` failed with `UnavaialbleException` on the latest [win32 build](http://cassci.datastax.com/view/win32/job/cassandra-3.0_dtest_win32/35/testReport/repair_test/TestRepair/simple_sequential_repair_test/), so I added a few `wait_other_notice` to make sure all nodes see each other after start/stop.


ptnapoleon:
+1

PR #521, opened 2015-08-28
Make commit log segment size tests more deterministic (CASSANDRA-10098)

pauloricardomg:
This PR makes the commitlog segment size tests more deterministic, as discussed on [CASSANDRA-10098](https://issues.apache.org/jira/browse/CASSANDRA-10098)

Since the main objective of these tests is to verify that the `commitlog_segment_size_in_mb` property is respected, I removed the checks for number of commit log files and total commit log size, since these may greatly vary on different test runs, environments, cassandra versions and compression settings, so there's a lot of variation. There are other tests that test commit log integrity, so I don't think it's crucial to test these attributes on these tests.

Since compressed commit log segment sizes may also vary depending on different factors, I modified the test to check that the compressed commit log segment size is within [segment_size/2, segment_size], so we accept a maximum of 50% compression ratio for compressed segments.

### DIFF
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -7,7 +7,6 @@ import time
 from cassandra import WriteTimeout
 from cassandra.cluster import NoHostAvailable, OperationTimedOut
 
-import ccmlib
 from ccmlib.common import is_win
 from assertions import assert_almost_equal, assert_none, assert_one
 from dtest import Tester, debug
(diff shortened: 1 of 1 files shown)


aboudreault:
LGTM, just a minor lint to fix. yep, I agree that there are other tests for commitlog integrity.

pauloricardomg:
Thanks! fixed lint.

PR #524, opened 2015-09-01
Improve git in require

mambocab:
When `git`, called through `@require` printed whitespace to stderr, the decorator would fail. Because of the way it was called when loading through nose (and because of poor error reporting design on my part), the failure was silent.

I'm not sure I understand what I was going for when I originally wrote this branch-handling code, but my thoughts on it now are in my commit messages; don't let this pass review if you don't agree with them.

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -328,8 +328,8 @@ def cassandra_git_branch():
     p = subprocess.Popen(['git', 'branch'], cwd=CASSANDRA_DIR,
                          stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     out, err = p.communicate()
-    # fail if err was printed and isn't empty OR if git failed
-    if err.split() or (p.returncode != 0):
+    # fail if git failed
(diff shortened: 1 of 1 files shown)


ptnapoleon:
This was breaking things in windows as well, perhaps more obviously since the bucketizer isn't there. I probably should/could have noticed the timing of those two issues. I'm +1

mambocab:
Thanks for the review. I'm going to run with this on cassci and merge if it works.

mambocab:
@mshuler Because of the way I was checking for errors when I shelled out to git, the stderr printed for tracing made `@require` fail. However, failing on stderr being printed was dumb, so I've removed that. When this gets merged, you should be able to add back the git tracing on cassci.

@ptnapoleon could you re-review please? I removed the checking for stderr.

ptnapoleon:
+1

mambocab:
Great, thank you.

PR #526, opened 2015-09-02
Backwards compat modules

mambocab:
This adds in the long-suffering upgrade tests from #471. They do still need to be refactored, but they will be useful to the devs running on trunk.

This also makes the subfolders in the dtests into modules. This will fix the recent problem where some modules of tests didn't run at all because of hidden errors during test module loading. [Here is the list of tests that weren't running](https://gist.github.com/mambocab/5c75a4c7d36b7666bea3); they ran successfully in this CassCI job:

http://cassci.datastax.com/job/scratch_mambocab-dtest_trunk/lastCompletedBuild/testReport/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,3 +57,7 @@ To run doxygen to generate HTML from these test plans, you will need to do the f
         doxygen doxygen/Doxyfile_python
 
 Feel free to submit test plans without the implemented tests. If you are submitting a new test, we would appreciate if it were annotated in this manner. If that is not possible, we will add the markup to your pull request.
+
+## Modules
+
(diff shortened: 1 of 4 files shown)


ptnapoleon:
Is this ready for review, would you like me to review, and did you see the travis complaints?

mambocab:
@ptnapoleon yes, this is ready for review, and you'd be a good person for it.

Just saw the linter problems -- I thought they'd be more extensive. There are few enough that it'll be worth the effort now; I'll fix those up.

ptnapoleon:
This is pre-refactor, just to confirm? So if all we're doing here is modulizing, and making a note in the contributing doc, then +1.

mambocab:
Yes, this is pre-refactor; it

- re-adds the upgrade tests
- makes folders into modules and a note in CONTRIBUTING
- makes a few fixup changes to the upgrade tests, in particular fixing imports, linter errors, and an incorrect use of the ccm cluster-creating api.

ptnapoleon:
+1

PR #527, opened 2015-09-02
added 2 secondary indexes on same map column test - CASSANDRA-9459

nutbunnies:


knifewine:
lgtm, +1 though we might have a little coverage overlap here and with the test_map_indexes method.

nutbunnies:
Agreed but I think the slight overlap is better than adding more version conditional checks to the map test.  I updated the name of the test and pydoc to better convey that it's not directly about maps.

PR #528, opened 2015-09-02
9839 dtests

pauloricardomg:
Tests for [CASSANDRA-9839](https://issues.apache.org/jira/browse/CASSANDRA-9839)

### DIFF
--- a/assertions.py
+++ b/assertions.py
@@ -74,3 +74,20 @@ def assert_row_count(session, table_name, expected):
     assert count == expected, "Expected a row count of {} in table '{}', but got {}".format(
         expected, table_name, count
     )
+
+
+def assert_crc_check_chance_equal(session, table, expected, ks="ks", view=False):
+    """
+    driver still doesn't support top-level crc_check_chance property,
+    so let's fetch directly from system_schema
+    """
+    if view:
+        assert_one(session,
+                   "SELECT crc_check_chance from system_schema.views WHERE keyspace_name = 'ks' AND "
+                   "view_name = '{0}';".format(table),
+                   [expected])
+    else:
+        assert_one(session,
+                   "SELECT crc_check_chance from system_schema.tables WHERE keyspace_name = 'ks' AND "
+                   "table_name = '{0}';".format(table),
+                   [expected])
--- a/compression_test.py
+++ b/compression_test.py
@@ -2,6 +2,7 @@ import os
 
 from scrub_test import TestHelper
 from tools import since
+from assertions import assert_crc_check_chance_equal
 
 
 class TestCompression(TestHelper):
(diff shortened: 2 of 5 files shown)


ptnapoleon:
Can upgrade_crc_check_chance_test.py go into the `upgrade_tests` directory/module? @mambocab, does that seem like a good idea to you?

pauloricardomg:
+1, waiting for @mambocab's verdict :)

upgrade_crc_check_chance_test.py
```diff
@@ -0,0 +1,92 @@
...
+        session = self.patient_exclusive_cql_connection(node1)
+        session.cluster.refresh_schema_metadata()
+        meta = session.cluster.metadata.keyspaces['ks'].tables['cf1']
+
+        self.assertEqual('org.apache.cassandra.io.compress.DeflateCompressor', meta.options['compression']['class'])
+        self.assertEqual('256', meta.options['compression']['chunk_length_in_kb'])
+        # driver still doesn't support top-level crc_check_chance property, so let's fetch directly from system_schema
+        res = session.execute("SELECT crc_check_chance from system_schema.tables "
+                              "WHERE keyspace_name = 'ks' AND table_name  = 'cf1';")
+        assert rows_to_list(res) == [[0.6]], res
```
ptnapoleon: Here could you use self.assertEqual(rows_to_list(res), [[0.6]])? Or even better, the last two lines could become
`assert_one(session, "SELECT crc_check_chance from system_schema.tables "
                              "WHERE keyspace_name = 'ks' AND table_name  = 'cf1';", [0.6])`. You'll need to import assert_one from assertions.py though

upgrade_crc_check_chance_test.py
```diff
@@ -0,0 +1,92 @@
...
+        debug('Upgrading node {node} to {tag}'.format(**format_args))
+        # drain and shutdown
+        node.drain()
+        node.watch_log_for("DRAINED")
+        node.stop(wait_other_notice=False)
+        debug('{node} stopped'.format(**format_args))
+
+        # Ignore errors before upgrade on Windows
+        if is_win():
+            node.mark_log_for_errors()
```
ptnapoleon: This is 2.2 to 3.0, yes? Then there is no need to do this, we should only ignore pre-upgrade errors when coming from 2.1. @josh-mckenzie can confirm

compression_test.py
```diff
@@ -75,6 +96,15 @@ def compression_cql_options_test(self):
...
         self.assertEqual('DEFLATE', self._get_compression_type(os.path.join(sstable_path, sstables['compression_opts_table'][1])))
 
+    def _assert_crc_check_chance_equal(self, session, table, expected):
+        """
+        driver still doesn't support top-level crc_check_chance property,
+        so let's fetch directly from system_schema
+        """
+        res = session.execute("SELECT crc_check_chance from system_schema.tables "
+                              "WHERE keyspace_name = 'ks' AND table_name  = '{table}';".format(**locals()))
+        assert rows_to_list(res) == [[expected]], res
```
ptnapoleon: See comment below on improving the assert

pauloricardomg:
suggestions addresses, thanks!

mambocab:
Ah, sorry I missed this until now. I don't think it's necessary to move the upgrade tests into the upgrade test module -- all the tests currently in there are going to be refactored out. In general I'm +1 on moving tests into modules, but we may not wind up having the upgrade tests in their own module in the long term.

ptnapoleon:
Well if Jim feels that way, then +1. Merge whenever C* is ready for these tests.

pauloricardomg:
updated crc_check_chance upgrade tests to tests with more nodes and added MV validation tests for gc_grace_seconds

assertions.py
```diff
@@ -74,3 +74,20 @@ def assert_row_count(session, table_name, expected):
...
+    """
+    if view:
+        assert_one(session,
+                   "SELECT crc_check_chance from system_schema.views WHERE keyspace_name = 'ks' AND "
+                   "view_name = '{table}';".format(**locals()),
+                   [expected])
+    else:
+        assert_one(session,
+                   "SELECT crc_check_chance from system_schema.tables WHERE keyspace_name = 'ks' AND "
+                   "table_name = '{table}';".format(**locals()),
```
ptnapoleon: Instead of `**locals()`, can you just use `table`? It's more readable.

beobal:
There was an additional small change required to `cqlsh_tests.test_describe` so I've opened #567.

PR #531, opened 2015-09-03
set concurrent schema alter tests to flake with max of 2 - CASSANDRA-…

nutbunnies:
updated concurrent tests to be flaky - @EnigmaCurry

### DIFF
--- a/concurrent_schema_changes_test.py
+++ b/concurrent_schema_changes_test.py
@@ -3,6 +3,7 @@ import os
 import pprint
 import re
 import time
+from flaky import flaky
 from random import randrange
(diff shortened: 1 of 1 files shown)


mambocab:
I'm not sure that `flaky` is what we want here -- `flaky` will run the test multiple times and mark the run as successful if any succeed, to differentiate hard failures from nondeterministic ones. I think I would want to run the test multiple times, then mark it as failed if any of the runs fail, hoping we'll catch what we know is a nondeterministic failure. If I'm wrong about that and that's not our goal, then disregard. The code change itself looks good.

PR #532, opened 2015-09-04
Tests for CASSANDRA-10052

stef1927:


ptnapoleon:
Can you rebase? I am not sure how I missed this originally, sorry.

stef1927:
No problem, rebased.

stef1927:
I added a @require(10052) to restart_node_test as well because after enhancing it I discovered a bug, also fixed by 10052.

dtest.py
```diff
@@ -609,6 +611,13 @@ def get_ip_from_node(self, node):
             node_ip = node.network_interfaces['thrift'][0]
         return node_ip
 
+    def get_port_from_node(self, node):
+        if node.get_base_cassandra_version() >= 2.1:
+            port = node.network_interfaces['binary'][1]
```
ptnapoleon: Are you sure we should be filtering on 2.1 here? We are still connecting via the native protocol to 1.2 and 2.0, afaik. In fact, the _create_session method is always strictly used with the native driver.

pushed_notifications_test.py
```diff
@@ -103,30 +104,112 @@ def move_single_node_test(self):
...
+        therefore we must change the rpc port or else processes won't start.
+        """
+        cluster = self.cluster
+        cluster.populate(2)
+        node1, node2 = cluster.nodelist()
+
+        i = 0  # change 'rpc_address' from '127.0.0.x' to 'localhost' and diversify port numbers
+        for node in cluster.nodelist():
+            node.network_interfaces['thrift'] = ('localhost', node.network_interfaces['thrift'][1] + i)
+            node.network_interfaces['binary'] = ('localhost', node.network_interfaces['thrift'][1] + 1)
```
ptnapoleon: Should we be duplicating the thrift port here? Or should the second line say `node.network_interfaces['binary'][1] + i`

pushed_notifications_test.py
```diff
@@ -103,30 +104,112 @@ def move_single_node_test(self):
...
+        """
+        cluster = self.cluster
+        cluster.populate(3)
+        node1, node2, node3 = cluster.nodelist()
+
+        # change node3 'rpc_address' from '127.0.0.x' to 'localhost', increase port numbers
+        i = 0
+        for node in cluster.nodelist():
+            node.network_interfaces['thrift'] = ('localhost', node.network_interfaces['thrift'][1] + i)
+            node.network_interfaces['binary'] = ('localhost', node.network_interfaces['thrift'][1] + 1)
```
ptnapoleon: Ditto to the comment below about this being a potential copy-paste error.

pushed_notifications_test.py
stef1927: It's correct as it is, we want the binary port to be equal to the thrift port + 1, to give them unique values so we can always bind on them. Then `i` is increased by 2 and the next node gets the next consecutive values.

dtest.py
stef1927: Initially I had it the same as `get_ip_from_node`, using the binary interface whenever it is available. However, I saw `run_cqlsh` in node.py and changed it. I guess it is wrong. Now I reverted it back to use the binary interface as available, same as `get_ip_from_node`.  See latest commit.

ptnapoleon: Okay. I'll have to look into `get_ip_from_node` to see if it's correct, but this should work here.

ptnapoleon:
+1. Merging since the require is already in.

PR #533, opened 2015-09-04
only run upgrade tests on 2.2 and later

mambocab:
Pursuant to some discussion on IRC, these tests need to be run over a few upgrade paths (see [CASSANDRA-10269](https://issues.apache.org/jira/browse/CASSANDRA-10269). Until then, they are hard-coded to upgrade from 2.1, so let's only run them on 2.2 and greater for now.


ptnapoleon:
+1

mambocab:
Thanks!

PR #536, opened 2015-09-08
Ignore 2.1 errors on non-upgraded node

ptnapoleon:
in mixed version cluster.

@josh-mckenzie to review


ptnapoleon:
@mambocab as well, as I just made some changes based on his input

mambocab:
What were the errors? I think this could use a comment, explaining why we need to do this, how it accomplishes that, and an example of the error or errors we want to ignore.

ptnapoleon:
Comment pushed. Also added the version check earlier, in preparation for your refactor.

josh-mckenzie:
Looks good.

mambocab:
+1, thanks!

PR #537, opened 2015-09-08
Specify upgrade paths

mambocab:
This makes the upgrade tests run different paths depending on the version you're running the tests on. In particular,

| C* Version | upgrade path |
| :--- | :--- |
| 2.1 | 2.1->3.0 | 
| 2.2 | 2.2->3.0 | 
| 3.0 | 3.0->trunk | 
| trunk | 2.2->trunk |

as discussed [in CASSANDRA-10269](https://issues.apache.org/jira/browse/CASSANDRA-10269).

This also adds an environment variable that lets users configure what version to upgrade to.

### DIFF
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -21,6 +21,9 @@ OLD_CASSANDRA_DIR = os.environ.get('OLD_CASSANDRA_DIR', None)
 # a 3.0 cluster will use.
 UPGRADE_MODE = os.environ.get('UPGRADE_MODE', 'normal').lower()
 
+# Specify a branch to upgrade to
+UPGRADE_TO = os.environ.get('UPGRADE_TO', None)
+
 
 @since('2.1')
 class UpgradeTester(Tester):
@@ -109,7 +112,9 @@ class UpgradeTester(Tester):
                 node2.mark_log_for_errors()
 
         # choose version to upgrade to
-        if self.original_git_branch == 'trunk' or self.original_version >= '3.0':
+        if UPGRADE_TO:
+            new_branch = UPGRADE_TO
+        elif self.original_git_branch == 'trunk' or self.original_version >= '3.0':
             new_branch = 'git:trunk'
         else:
             new_branch = 'git:cassandra-3.0'
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -1,8 +1,9 @@
 import os
 import time
 
-from ccmlib.common import is_win
-from dtest import DEBUG, Tester
+from ccmlib.common import get_version_from_build, is_win
+from dtest import DEBUG, Tester, debug
+from tools import cassandra_git_branch, since
 
 QUERY_UPGRADED = os.environ.get('QUERY_UPGRADED', 'true').lower() in ('yes', 'true')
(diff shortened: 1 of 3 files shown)


upgrade_tests/upgrade_base.py
```diff
@@ -20,8 +21,19 @@
...
 
+
+@since('2.1')
 class UpgradeTester(Tester):
+    """
+    When run in 'normal' upgrade mode without specifying any version to run,
+    this will test different upgrade paths depending on what version of C* you
+    are testing. When run on 2.1 or 2.2, this will test the upgrade to 3.0.
+    When run on 3.0, this will test the upgrade path to trunk. When run on
+    versions above 3.0, this will test the upgrade path from 2.2 to HEAD.
```
ptnapoleon: 2.2 to HEAD? But 2.2 can't upgrade past 3.0

upgrade_tests/upgrade_base.py
```diff
@@ -44,12 +56,18 @@ def prepare(self, ordered=False, create_keyspace=True, use_cache=False, nodes=2,
...
             if OLD_CASSANDRA_DIR:
                 cluster.set_install_dir(install_dir=OLD_CASSANDRA_DIR)
             else:
-                cluster.set_install_dir(version='git:cassandra-2.1')
-            cluster.start()
+                # upgrade from 3.0 to current install dir if we're running from trunk
+                if self.original_git_branch == 'trunk':
+                    cluster.set_install_dir(version='git:cassandra-3.0')
+            cluster.start(wait_for_binary_proto=True)
+            print 'starting from ' + get_version_from_build(node1.get_install_dir())
```
ptnapoleon: Boooooo. String interpolation, dude.

upgrade_tests/upgrade_base.py
```diff
@@ -88,15 +106,25 @@ def do_upgrade(self, session):
...
 
+        # choose version to upgrade to
+        if UPGRADE_TO:
+            new_branch = UPGRADE_TO
+        elif self.original_git_branch == 'trunk' or self.original_version >= '3.0':
+            new_branch = 'git:trunk'
+        else:
+            new_branch = 'git:cassandra-3.0'
+
+        print 'upgrading to ' + new_branch
```
ptnapoleon: Also, these should be debug(), not print.

upgrade_tests/upgrade_base.py
mambocab: Good thinking. Is it ok to leave this in as long as trunk is targeting 3.0?

upgrade_tests/upgrade_base.py
mambocab: but... but it'll always return a string!

mambocab: ok, fine

mambocab:
@ptnapoleon rebased, addressed all your comments except the one on the docstring. Thoughts?

upgrade_tests/upgrade_base.py
ptnapoleon: But trunk is no longer 3.0. It's 3.2. And while the upgrade is still "possible" from 2.2 to trunk, it is no longer guaranteed to be, and that will break with little/zero warning.

mambocab: Aahhh, yes. Fixed and rebased.

ptnapoleon:
Then +1

PR #538, opened 2015-09-09
Add test for CASSANDRA-10215

beobal:
Only the coordinator should select which index, if any, should be used for a query. Previously, the index selection process would be run every time the index was accessed and on every replica.

### DIFF
--- a/secondary_indexes_test.py
+++ b/secondary_indexes_test.py
@@ -56,8 +56,7 @@ class TestSecondaryIndexes(Tester):
         cluster.populate(3).start()
         node1, node2, node3 = cluster.nodelist()
 
-        conn = self.patient_cql_connection(node1)
-        session = conn
+        session = self.patient_cql_connection(node1)
         session.max_trace_wait = 120
         session.execute("CREATE KEYSPACE ks WITH REPLICATION = {'class': 'SimpleStrategy', 'replication_factor': '1'};")
         session.execute("CREATE TABLE ks.cf (a text PRIMARY KEY, b text);")
@@ -265,7 +264,6 @@ class TestSecondaryIndexes(Tester):
             time.sleep(0.10)
             self.wait_for_schema_agreement(session)
 
-
     @since('3.0')
     def test_only_coordinator_chooses_index_for_query(self):
         """
@@ -276,8 +274,7 @@ class TestSecondaryIndexes(Tester):
         cluster = self.cluster
         cluster.populate(3).start()
         node1, node2, node3 = cluster.nodelist()
-        conn = self.patient_cql_connection(node1)
-        session = conn
+        session = self.patient_cql_connection(node1)
         session.max_trace_wait = 120
         session.execute("CREATE KEYSPACE ks WITH REPLICATION = {'class': 'SimpleStrategy', 'replication_factor': '1'};")
         session.execute("CREATE TABLE ks.cf (a text PRIMARY KEY, b text);")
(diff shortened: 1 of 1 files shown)


secondary_indexes_test.py
```diff
@@ -266,6 +266,70 @@ def wait_for_schema_agreement(self, session):
             self.wait_for_schema_agreement(session)
```
ptnapoleon: There should only be one new line between these two tests.

secondary_indexes_test.py
```diff
@@ -266,6 +266,70 @@ def wait_for_schema_agreement(self, session):
...
+        # node2 is the coordinator for the query
+        check_trace_events(query.trace,
+                           "Index mean cardinalities are b_index:[0-9]*. Scanning with b_index.",
+                           [("127.0.0.1", 0, 0), ("127.0.0.2", 1, 1), ("127.0.0.3", 0, 0)])
+        # check that the index is used on each node, really we only care that the matching
+        # message appears on every node, so the max count is not important
+        check_trace_events(query.trace,
+                           "Executing read on ks.cf using index b_index",
+                           [("127.0.0.1", 1, 200), ("127.0.0.2", 1, 200), ("127.0.0.3", 1, 200)])
+
```
ptnapoleon: There should be two new lines before the start of a new class.

secondary_indexes_test.py
```diff
@@ -266,6 +266,70 @@ def wait_for_schema_agreement(self, session):
...
+        """
+        Checks that the index to use is selected (once) on the coordinator and
+        included in the serialized command sent to the replicas.
+        @jira_ticket CASSANDRA-10215
+        """
+        cluster = self.cluster
+        cluster.populate(3).start()
+        node1, node2, node3 = cluster.nodelist()
+        conn = self.patient_cql_connection(node1)
+        session = conn
```
ptnapoleon: What is the point of defining conn? Why not `session = self.patient_cql_connection(node1)`?

ptnapoleon:
+1 on the test logic. If you could make those whitespace changes to make Travis happy, that would be great.

secondary_indexes_test.py
```diff
@@ -266,6 +266,70 @@ def wait_for_schema_agreement(self, session):
...
+        session.max_trace_wait = 120
+        session.execute("CREATE KEYSPACE ks WITH REPLICATION = {'class': 'SimpleStrategy', 'replication_factor': '1'};")
+        session.execute("CREATE TABLE ks.cf (a text PRIMARY KEY, b text);")
+        session.execute("CREATE INDEX b_index ON ks.cf (b);")
+        num_rows = 100
+        for i in range(num_rows):
+            indexed_value = i % (num_rows / 3)
+            # use the same indexed value three times
+            session.execute("INSERT INTO ks.cf (a, b) VALUES ('%d', '%d');" % (i, indexed_value))
```
mambocab: Could you please, here and elsewhere, use `.format` instead of `%` for string formatting? Thanks.

beobal: Sorry yes, bad habits die hard. I've fixed this instance & will try to remember in future.

secondary_indexes_test.py
beobal: No clue what that was for, it was just a copypasta from `test_low_cardinality_indexes`. I've fixed it in both places.

beobal:
squashed and merged

mambocab:
:+1: thank you!

ptnapoleon:
For easy searching, this was merged at 3929e9f2954aeb9e12d4bae623e175e1b388cc31

PR #529, opened 2015-09-03
Skip failing tests for CASSANDRA-10086 due to TERM variable not set

stef1927:
Sample failure [here] (http://cassci.datastax.com/job/stef1927-8630-3.0-dtest/lastBuild/testReport/cqlsh_tests/TestCqlsh/test_clear/).

It seems the TERM environment variable is not set, i.e. we don't have a terminal on Jenkins. Unless someone knows how to address this, I propose to skip them.

PR #539, opened 2015-09-09
adds TERM env variable to cqlsh clear tests

mambocab:
As discussed on PR #529, this fixes the cqlsh tests for CLEAR and CLS on Linux by setting the TERM environment variable.


ptnapoleon:
+1

PR #540, opened 2015-09-09
Never use the thrift port.

ptnapoleon:
@mambocab to review

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -612,11 +612,15 @@ class Tester(TestCase):
         return node_ip
 
     def get_port_from_node(self, node):
-        if node.network_interfaces['binary']:
-            port = node.network_interfaces['binary'][1]
-        else:
-            port = node.network_interfaces['thrift'][1]
-        return port
+        """
+        Return the port that this node is listening on.
+        We only use this to connect the native driver,
+        so we only care about the binary port.
+        """
+        try:
+            return node.network_interfaces['binary'][1]
+        except Exception as e:
+            raise RuntimeError("No network interface defined on this node object. {}".format(node.network_interfaces))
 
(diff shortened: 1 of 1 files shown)


mambocab:
Are we confident that it won't be None as long as the node is running? If not, we should add a check for that.

ptnapoleon:
What do we want to do if it is None?

ptnapoleon:
Also, looking at the ccm code, I am very sure it will never be None. Someone would need to create a Node object, and explicitly pass in None for the binary interface, which should screw up launching the actual C* process.

mambocab:
If you're sufficiently confident about that, then use this as-is. If we wanted fail fast, I'd raise a `RuntimeError` with a message about the problem.

ptnapoleon:
It should just fail out with a KeyError, but I will improve the message, then ping you here.

ptnapoleon:
Thoughts?

mambocab:
Catch a `KeyError` instead of a base `Exception` and I'm +1 -- we don't want to erroneously catch a random operating system error or something.

ptnapoleon:
It isn't always a KeyError though. I can also make it get a TypeError, and I'm not sure that those two constitute an exhaustive list of potential problems. But I'm pretty (read 90%) sure that since I'm using `except Exception:` and not `except:` I will only catch things propagated by Python, and OS level signals will still work, i.e., I'm not breaking sigkill. Thoughts? I could just catch Key and Type, and call it a day.

mambocab:
Oh, OS stuff doesn't inherit from Exception, TIL. That is good. I'm +1.

ptnapoleon:
Yup. Tyler taught me that.

PR #541, opened 2015-09-09
Use autopep8 to cleanup dtest

ptnapoleon:
Work done by autopep8. @mambocab and myself to review.


ptnapoleon:
It looks good to me.

auth_roles_test.py
```diff
@@ -1168,7 +1168,7 @@ def assert_unauthenticated(self, message, user, password):
             self.cql_connection(node, user=user, password=password)
         host, error = response.exception.errors.popitem()
         pattern = 'Failed to authenticate to %s: code=0100 \[Bad credentials\] message="%s"' % (host, message)
-        assert type(error) == AuthenticationFailed, "Expected AuthenticationFailed, got %s" % error
+        assert isinstance(error, AuthenticationFailed), "Expected AuthenticationFailed, got %s" % error
```
mambocab: Maybe this is out of scope for this PR, but at some point we should change this from a bare `assert` to an `assertIsInstance`.

mambocab:
Yep, looks great.

PR #542, opened 2015-09-09
Remove requires

mambocab:
These two tests have started succeeding:

http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/61/testReport/upgrade_tests.cql_tests/TestCQL/only_pk_test/

http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/61/testReport/nodetool_test/TestNodetool/test_decommission_after_drain_is_invalid/

So I've un-required them. CASSANDRA-8741 has been closed, so that test definitely is ready to be unskipped, and CASSANDRA-6717 is in progress, so it makes sense to me that the other test would start succeeding now.

### DIFF
--- a/nodetool_test.py
+++ b/nodetool_test.py
@@ -1,11 +1,9 @@
 from ccmlib.node import NodetoolError
 from dtest import Tester
-from tools import require
 
 
 class TestNodetool(Tester):
 
-    @require("8741")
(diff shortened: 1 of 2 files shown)


ptnapoleon:
+1

PR #545, opened 2015-09-10
test that repros CASSANDRA-10238

knifewine:


knifewine:
@ptnapoleon and/or @mambocab ?

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
         # Given a diverse enough keyset, each node in the second
         # datacenter should get a chance to be a forwarder:
         self.assertEqual(len(forwarders_used), 3)
+
+
+class Cassandra10238Test(Tester):
+    """
+    Test to repro CASSANDRA-10238, wherein consolidating racks without a restart could violate RF contract.
```
ptnapoleon: Check out the contributing.md file, for how to annotate this for doxygen.

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
...
+            with open(os.path.join(node.get_conf_dir(), 'cassandra-rackdc.properties'), 'w') as topo_file:
+                topo_file.write("dc=DC1\n")
+                topo_file.write("rack=RAC{}".format(i))
+
+        cluster.start()
+
+        conn = self.patient_cql_connection(node1)
+
+        # small stress write to build ks/table for us
+        node1.stress(['write', 'n=20', '-rate', 'threads=50', '-schema', 'keyspace=testing', '-pop', 'seq=1..20'])
```
ptnapoleon: You're using the 2.1+ stress syntax, but this test also runs on 2.0

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
...
+        conn.execute("""ALTER KEYSPACE testing WITH replication = {'class':'NetworkTopologyStrategy', 'DC1':3}""")
```
ptnapoleon: Can't you also do this via stress?

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
...
+
+        # start with separate racks
+        for i, node in enumerate(cluster.nodelist()):
+            with open(os.path.join(node.get_conf_dir(), 'cassandra-rackdc.properties'), 'w') as topo_file:
+                topo_file.write("dc=DC1\n")
+                topo_file.write("rack=RAC{}".format(i))
+
+        cluster.start()
+
+        conn = self.patient_cql_connection(node1)
```
ptnapoleon: Why not name this `session`?

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
...
+        """
+        for node in nodes:
+            cmd = "getendpoints {} {} 11".format(ks, table)
+            out, err = node.nodetool(cmd)
+
+            if len(err.strip()) > 0:
+                debug(err)
+                raise RuntimeError("Error running 'nodetool {}'".format(cmd))
+
+            out = out.rstrip()  # remove extra \n from output
```
mambocab: I think I'd like a little more filtering than this, just to make sure we don't have problems with misinterpreted error output or anything like that. These should be IP addresses, right? Looks like writing a little wrapper to validate them should be easy enough: http://stackoverflow.com/a/319298

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
...
+
+            self.assertEqual(len(out.split('\n')), expected_count, "wrong number of endpoints found ({}), should be: {}".format(len(out.split('\n')), expected_count))
+
+    def wait_for_all_nodes_on_rack(self, nodes, rack_name):
```
mambocab: This is really useful; you should consider migrating it to ccm to encourage us to do more rack-aware testing :)

replication_test.py
```diff
@@ -254,3 +255,87 @@ def network_topology_test(self):
...
+class Cassandra10238Test(Tester):
+    """
+    Test to repro CASSANDRA-10238, wherein consolidating racks without a restart could violate RF contract.
+    """
+    def check_endpoint_count(self, ks, table, nodes, expected_count):
+        """
+        Check a dummy key (of a stress generated table) expecting it to have expected_count endpoints on each node.
+        """
+        for node in nodes:
+            cmd = "getendpoints {} {} 11".format(ks, table)
```
mambocab: Is the key `11` always going to be in there? I think it is, but only because you used `-pop seq=1..20` before calling this. Since it depends on state set up beforehand, I think the key should be a parameter to this method.

replication_test.py
mambocab: I think this is clear, and I don't see the need to. Would using stress be better or clearer?

Since this is isn't a multi-line string, the triple-quotes aren't necessary.

mambocab:
done with review!

replication_test.py
knifewine: it's actually just a dummy value, doesn't need to be present to repro the bug. so the -pop seq isn't strictly needed.

replication_test.py
knifewine: I thought we decided a while back that connection was better nomenclature? I don't have an especially strong opinion either way.

ptnapoleon: Did we? @mambocab do you remember that? I thought we settled on `session`, as that's what is recommended by the drivers.

replication_test.py
ptnapoleon: It feels cleaner in theory, but really the difference is academic, so no issues either way.

knifewine:
pushed new commits to use 'connection', and it turns out stress isn't needed at all so I went back to a really simple ks/table setup instead.

replication_test.py
mambocab: https://github.com/riptano/cassandra-dtest/pull/403#issuecomment-122920460

ptnapoleon:
Jim still has one comment unaddressed, and as long as you ran this test and are sure it works, I like it, +1.

Beautiful code as always, Russ.

replication_test.py
```diff
@@ -254,3 +255,86 @@ def network_topology_test(self):
...
+            out, err = node.nodetool(cmd)
+
+            if len(err.strip()) > 0:
+                debug(err)
+                raise RuntimeError("Error running 'nodetool {}'".format(cmd))
+
+            ips_found = re.findall('(\d+\.\d+\.\d+\.\d+)', out)
+
+            self.assertEqual(len(ips_found), expected_count, "wrong number of endpoints found ({}), should be: {}".format(len(ips_found), expected_count))
+
```
knifewine: @mambocab I threw out the "\n" splitting for a regex -- this work for ya?

mambocab: :+1:

PR #550, opened 2015-09-15
Add test for C-10066

ptnapoleon:
@mambocab to review.

PS, if you think this is readable, you are wrong.

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -40,6 +40,34 @@ class TestCqlsh(Tester):
             os.unlink(self.tempfile.name)
         super(TestCqlsh, self).tearDown()
 
+    @since('2.1.9')
+    def test_pep8_compliance(self):
+        """
+        @jira_ticket CASSANDRA-10066
+        Checks that cqlsh is compliant with pep8 with the following command:
+        pep8 --ignore E501,E402,E731 pylib/cqlshlib/*.py bin/cqlsh.py
+        """
+        cluster = self.cluster
+
+        if cluster.version() < '2.2':
+            cqlsh_path = os.path.join(cluster.get_install_dir(), 'bin', 'cqlsh')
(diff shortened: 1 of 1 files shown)


ptnapoleon:
It passes on every version, for now.

cqlsh_tests/cqlsh_tests.py
```diff
@@ -40,6 +40,35 @@ def tearDown(self):
...
+            cqlsh_path = os.path.join(cluster.get_install_dir(), 'bin', 'cqlsh')
+        else:
+            cqlsh_path = os.path.join(cluster.get_install_dir(), 'bin', 'cqlsh.py')
+
+        cqlshlib_path = os.path.join(cluster.get_install_dir(), 'pylib', 'cqlshlib')
+        cqlshlib_paths = os.listdir(cqlshlib_path)
+        cqlshlib_paths = [os.path.join(cqlshlib_path, x) for x in cqlshlib_paths if '.py' in x and '.pyc' not in x]
+
+        cmds = ['pep8', '--ignore', 'E501,E402,E731', cqlsh_path]
+        cmds.extend(cqlshlib_paths)
```
ptnapoleon: If you're wondering why this is two lines, so am I. ['a', 'b'].extend(['c']) returns None. But `x=['a','b'];x.extend(['c'])` works.

mambocab:
This seems reasonable to me. I'm not convinced that the dtests are the best place for this in principle, but we don't have another test suite that runs on dev branches, so this works.

cqlsh_tests/cqlsh_tests.py
mambocab: Yeah, `extend` and `append` are unfortunate. Since you know both of these are lists, you should be able to use `+`, though:

```
In [1]: [1, 2, 3] + [4, 5, 6]
Out[1]: [1, 2, 3, 4, 5, 6]
```

PR #554, opened 2015-09-16
Update cqlsh_test and schema_metadata_test for CASSANDRA-10216

beobal:


cqlsh_tests/cqlsh_tests.py
```diff
@@ -653,7 +653,8 @@ def test_describe(self):
         self.execute(cql='CREATE INDEX myindex ON test.users (age)')
         self.execute(cql='DESCRIBE INDEX test.myindex', expected_output=self.get_index_output('myindex', 'test', 'users', 'age'))
 
-        # Alter table (rename of column with idx is currently not allowed)
+        # Alter table (rename of column with idx is currently not allowed, so we drop it first)
+        self.execute(cql='DROP INDEX test.test_val_idx')
```
mambocab: If renaming columns with an index isn't allowd, did this part of the test ever work?

beobal: yes, previously you could drop an indexed column and the index would also be dropped. Now though, the relationship between columns and indexes is more fluid, so instead we reject the drop column if any index says it's using it. 
The addition to the pre-existing comment is a little bit unclear I guess;  what it really meant before was `rename of column with idx is currently not allowed so this test doesn't attempt that`
Perhaps it should now read `#Alter table. Rename or drop of a column with idx is not allowed, so the index must be dropped first.

mambocab: I see -- yeah, this comment would be clearer if you specified that this is new behavior and mentioned that you're dropping the index (I know it's in the test, but being specific is good here, I think).

beobal: done

mambocab:
Great -- I've run the tests in the changed files locally, and it looks good. Thanks!

PR #555, opened 2015-09-16
checks for default protocol_version value

mambocab:
The upgrade code should handle the protocol_version being None.


ptnapoleon:
Lgtm +1

PR #556, opened 2015-09-16
Add complex_mv_statements_test

aboudreault:
https://issues.apache.org/jira/browse/CASSANDRA-9664

to review, do not merge until the patch is committed.

### DIFF
--- a/materialized_views_test.py
+++ b/materialized_views_test.py
@@ -943,6 +943,131 @@ class TestMaterializedViews(Tester):
 
         assert_none(session2, "SELECT * FROM ks.t_by_v WHERE v2 = 'a'", cl=ConsistencyLevel.QUORUM)
 
+    def complex_mv_select_statements_test(self):
+        """
+        Test complex MV select statements
+        @jira_ticket CASSANDRA-9664
+        """
+
+        cluster = self.cluster
+        cluster.populate(3).start()
+        node1 = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+
+        debug("Creating keyspace")
+        session.execute("CREATE KEYSPACE mvtest WITH replication = "
+                        "{'class': 'SimpleStrategy', 'replication_factor': '3'}")
+        session.execute('USE mvtest')
+
+        mv_primary_keys = ["((a, b), c)",
+                           "((b, a), c)",
+                           "(a, b, c)",
+                           "(c, b, a)",
+                           "((c, a), b)"]
+
+        for mv_primary_key in mv_primary_keys:
+
+            session.execute("CREATE TABLE test (a int, b int, c int, d int, PRIMARY KEY (a, b, c))")
+
+            insert_stmt = session.prepare("INSERT INTO test (a, b, c, d) VALUES (?, ?, ?, ?)")
+            update_stmt = session.prepare("UPDATE test SET d = ? WHERE a = ? AND b = ? AND c = ?")
+            delete_stmt1 = session.prepare("DELETE FROM test WHERE a = ? AND b = ? AND c = ?")
+            delete_stmt2 = session.prepare("DELETE FROM test WHERE a = ?")
+
+            session.cluster.control_connection.wait_for_schema_agreement()
+
+            rows = [(0, 0, 0, 0),
+                    (0, 0, 1, 0),
+                    (0, 1, 0, 0),
+                    (0, 1, 1, 0),
+                    (1, 0, 0, 0),
+                    (1, 0, 1, 0),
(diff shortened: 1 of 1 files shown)


materialized_views_test.py
```diff
@@ -943,6 +943,133 @@ def really_complex_repair_test(self):
...
+        cluster = self.cluster
+        cluster.populate(3).start()
+        node1 = cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+
+        debug("Creating keyspace")
+        session.execute(
+            ("CREATE KEYSPACE mvtest WITH replication = "
+             "{'class': 'SimpleStrategy', 'replication_factor': '3'}")
+        )
```
mambocab: This line should be short enough to put onto a single line if you like. If you don't, I think I'd prefer it be simplified to something like

```python
session.execute("CREATE KEYSPACE mvtest WITH replication = "
                "{'class': 'SimpleStrategy', 'replication_factor': '3'}")
```

aboudreault: I agree

materialized_views_test.py
```diff
@@ -943,6 +943,133 @@ def really_complex_repair_test(self):
...
+                    "0, 1, 0, 0",
+                    "0, 1, 1, 0",
+                    "1, 0, 0, 0",
+                    "1, 0, 1, 0",
+                    "1, 1, -1, 0",
+                    "1, 1, 0, 0",
+                    "1, 1, 1, 0"]
+
+            for row in rows:
+                session.execute("INSERT INTO test (a, b, c, d) VALUES ({values})".format(values=row))
```
mambocab: Any reason not to use a prepared statement and have the members of `rows` be lists of integers?

aboudreault: There is probably no benefit to do that, but it makes sense to use int there. Will do.

materialized_views_test.py
```diff
@@ -943,6 +943,133 @@ def really_complex_repair_test(self):
...
+
+            debug("Testing MV primary key: {}".format(mv_primary_key))
+
+            session.execute(("CREATE MATERIALIZED VIEW mv AS SELECT * FROM test WHERE "
```
mambocab: I don't believe this needs the extra parens around the string:

```
$ cat demo.py
from __future__ import print_function


print("does this parse? "
      "I sure hope so.")

$ python demo.py
does this parse? I sure hope so.
```

materialized_views_test.py
```diff
@@ -943,6 +943,133 @@ def really_complex_repair_test(self):
...
+            session.execute("INSERT INTO test (a, b, c, d) VALUES ({values})".format(values="1, 2, 1, 0"))
+            assert_all(
+                session, "SELECT a, b, c, d FROM mv",
+                [[1, 0, 1, 0], [1, 1, 1, 0], [1, 2, 1, 0]],
+                ignore_order=True,
+                cl=ConsistencyLevel.QUORUM
+            )
+
+            # update rows that does not match the filter
+            session.execute("UPDATE test SET d = {d} WHERE a = {a} AND b = {b} AND c = {c}".format(
```
mambocab: I think I'd like to see prepared statements here too.

materialized_views_test.py
```diff
@@ -943,6 +943,133 @@ def really_complex_repair_test(self):
...
+        )
+        session.execute('USE mvtest')
+
+        mv_primary_keys = ["((a, b), c)",
+                           "((b, a), c)",
+                           "(a, b, c)",
+                           "(c, b, a)",
+                           "((c, a), b)"]
+
+        for mv_primary_key in mv_primary_keys:
```
mambocab: Any particular reason to use a loop instead of a parameterized test? I don't have strong feelings either way, but I think it's worth thinking about both options.

aboudreault: Any exampe to point out?

mambocab:
My only remaining nit is that it's not clear to me what you mean by "filter" in the comments -- did I miss something?

aboudreault:
The **filter** is the WHERE clause specified during the MV creation. ie
```
CREATE MATERIALIZED VIEW mv AS SELECT * FROM test WHERE 
                            a = 1 AND b IS NOT NULL AND c = 1 [...]
```
The second line is the filter.

aboudreault:
Addressed all comments, thanks for the review!

materialized_views_test.py
mambocab: For the record: a loop is best here because it avoids starting a new cluster for each value.

mambocab:
Looks good post-fixes :+1: Thanks!

PR #559, opened 2015-09-17
removes require from skipped 10238 tests

mambocab:
These dtests pass on CassCI, and CASSANDRA-10238 has been resolved. @stef1927 Are these tests ready to be un-skipped?

### DIFF
--- a/replication_test.py
+++ b/replication_test.py
@@ -4,7 +4,7 @@ import re
 import time
 
 from dtest import Tester, debug, PRINT_DEBUG
-from tools import no_vnodes, require, since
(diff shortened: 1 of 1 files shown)


mambocab:
Here's the CassCI link: http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/69/testReport/replication_test/SnitchConfigurationUpdateTest/

stef1927:
Yes we can un-skip these tests so I am going to merge now. The fact that the ticket has been reopened should not affect these tests.

mambocab:
Thanks!

PR #560, opened 2015-09-17
changed is_upgraded checks that were used to determine if node was >=…

nutbunnies:
…2.2 in version to new helper

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -1,25 +1,23 @@
 # coding: utf-8
 
+import math
 import random
 import struct
 import time
-import math
 from collections import OrderedDict
-from uuid import uuid4, UUID
+from uuid import UUID, uuid4
 
-from thrift_bindings.v22.ttypes import CfDef, Mutation, ColumnOrSuperColumn, Column
-from thrift_bindings.v22.ttypes import ConsistencyLevel as ThriftConsistencyLevel
-
-from dtest import freshCluster, debug
-from assertions import assert_invalid, assert_one, assert_none, assert_all
-from thrift_tests import get_thrift_client
-from tools import since, require, rows_to_list
+from assertions import assert_all, assert_invalid, assert_none, assert_one
 from cassandra import ConsistencyLevel, InvalidRequest
 from cassandra.concurrent import execute_concurrent_with_args
 from cassandra.protocol import ProtocolException, SyntaxException
 from cassandra.query import SimpleStatement
 from cassandra.util import sortedset
-
+from dtest import debug, freshCluster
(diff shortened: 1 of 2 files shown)


nutbunnies:
@mambocab hour later the upgrade_tests.cql_tests.TestCQL test all pass against cassandra-3.0 now

upgrade_tests/upgrade_base.py
```diff
@@ -61,7 +61,6 @@ def get_default_upgrade_path(job_version):
     assert [start_version, upgrade_version].count(None) >= 1, err
     return UpgradePath(start_version, upgrade_version)
 
-
```
mambocab: Don't remove this line; per PEP8 there should be 2 lines of whitespace before a top-level class definition. I guess this didn't trip up Travis because it's a one-line change.

upgrade_tests/cql_tests.py
```diff
@@ -2339,13 +2337,22 @@ def composite_index_collections_test(self):
...
         cursor = self.prepare(ordered=True, use_cache=True)
 
-        cursor.execute("""
-            CREATE TABLE test (
-                k int PRIMARY KEY,
-                v1 int,
-                v2 int,
-            ) WITH CACHING = ALL;
-        """)
+        if self.get_node_version(False) >= '2.1':
```
mambocab: This is a confusing way to write `min([n.version for node in nodes])`. Maybe add another layer to this API, like a `node_versions()` method that returns a list. In situations like this, you could just use `min(self.node_versions()) >= '2.1'`, and `get_node_version` could build on `node_versions`.

mambocab:
This is excellent, thanks. I think `self.get_node_version(False)` could be clearer, but other than that, thanks, +1.

upgrade_tests/cql_tests.py
nutbunnies: what do you think of adding two new methods `get_node_versions` to get the list and `node_version_above` to remove the logic from if statement -- makes it more readable and slightly DRY'er

mambocab: I think I'd rather have the comparison logic in the `if` statement than write/maintain separate methods for `>`, `<`, `>=`, and `<=`. I don't feel super strongly, but that's how we manage it elsewhere, e.g. [here](https://github.com/riptano/cassandra-dtest/blob/0082fe953e0d7c3b66c7489424fa91f26bb10440/compaction_test.py#L424). If you think that's the way to go, I'd prefer the names be `_greater_than`, `_greater_equal`, etc., to match the unittest assertion names.

mambocab: Ah, I didn't put together we only would need the one comparison method. Nice! `_above` is fine, since there's just the one.

PR #562, opened 2015-09-18
Fix details of mv schema table

beobal:
While `concurrent_schema_changes_test` seems a little flaky in general, this particular test is failing consistently since the schema table was modified in CASSANDRA-9921.


mambocab:
I believe those tests are known to be flaky -- it's a C* bug that doesn't reproduce reliably:

https://issues.apache.org/jira/browse/CASSANDRA-10250

Changes look good to me. Thanks, merging.

PR #564, opened 2015-09-18
changed collection datatype check to only use the class name and not the

nutbunnies:
the package of the collections recently changed and broke this check -- it really only needs to verify the class name is the same


ptnapoleon:
Lgtm

PR #565, opened 2015-09-18
changed crc test to check for expected failure instead of verifying b…

nutbunnies:
as far as can tell this test verified the broken behavior of a node starting with corrupted commit logs -- updated to check for expected node shutdown and exception


mambocab:
Great, thanks -- looks like I missed this one when I fixed the original PR. +1.

PR #566, opened 2015-09-18
remove @require from CASSANDRA-9775 tests

mambocab:
CASSANDRA-9775 has been resolved. Since `test_with_equal_results_to_page_size` was hanging the build, I've removed the hard skip, but not the `@require`, so it will run on [the branch for `@require`d tests](http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/). If it's stable there for a few days, I'll make another PR to remove the `@require` there too.


ptnapoleon:
+1

PR #570, opened 2015-09-24
Add test of role name case preservation (CASSANDRA-10394)

beobal:


auth_roles_test.py
```diff
@@ -652,6 +652,33 @@ def create_user_as_alias_for_create_role_test(self):
         cassandra.execute("CREATE USER super_user WITH PASSWORD '12345' SUPERUSER")
         assert_one(cassandra, "LIST ROLES OF super_user", ["super_user", True, True, {}])
 
+    def role_name_test(self):
+        self.prepare()
```
ptnapoleon: Add a block quote here please, that looks like
```
"""
@jira_ticket CASSANDRA-10394
"""
```

ptnapoleon:
+1, after nit

beobal:
Merged manually in 6002e88

PR #571, opened 2015-09-24
Fix SnitchConfigurationUpdateTest on Windows

stef1927:
This will fix the line separator so that it works on Windows as well.


ptnapoleon:
Awesome, tyvm. +1

PR #572, opened 2015-09-28
exposed connect_timeout python driver parameter and set default to 10s

pauloricardomg:
There are number of [flakey tests](http://cassci.datastax.com/view/trunk/job/trunk_dtest/625/testReport/ttl_test/history/) due to driver connection timeout errors: `('Unable to connect to any servers', {'127.0.0.1': OperationTimedOut('errors=Timed out creating connection (5 seconds), last_host=None',)})`

I believe this happens when the jenkins slave is somehow overloaded, so I added a connect_timeout option to dtest.*cql_connection factory methods to set that parameter on cluster instantiation. We're currently using the driver default of 5s for connection timeout, but I think we can increase that to 10s and see if that makes these tests less flakey without affecting other tests.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -363,28 +363,35 @@ class Tester(TestCase):
         return protocol_version
 
     def cql_connection(self, node, keyspace=None, user=None,
-                       password=None, compression=True, protocol_version=None, port=None, ssl_opts=None):
+                       password=None, compression=True, protocol_version=None,
+                       port=None, ssl_opts=None, connect_timeout=None):
 
         return self._create_session(node, keyspace, user, password, compression,
-                                    protocol_version, port=port, ssl_opts=ssl_opts)
+                                    protocol_version, port=port, ssl_opts=ssl_opts,
+                                    connect_timeout=connect_timeout)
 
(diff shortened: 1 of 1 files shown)


mambocab:
In my opinion, this should have a

```
if connect_timeout is None:
    connect_timeout = 10
```

block in the `_create_session` method, and all the other methods that call it should have a default value of `None` -- this way, we set the default in only one place. @ptnapoleon?

I don't understand the problems on CassCI well enough to say if this is likely to fix them. I think it could be worth a try, but feedback from @ptnapoleon and @mshuler would be good.

pauloricardomg:
+1 to the default of none everywhere else, except on `_create_session`.  I was not comfortable with defining the default of 10 everywhere else, but didn't know how to make that better, so that looks much more elegant.

I'm not sure if that'll fix either, but for sure it won't hurt, so at least we can have some hope :P

ptnapoleon:
I agree with Jim's suggested edit. I should also note, we did add a RetryPolicy to help prevent this, but this also can't hurt.

ptnapoleon:
Already implemented in another PR, by Stefania

PR #573, opened 2015-09-29
unrequire passing, non-hanging paging test

mambocab:
Follow-up to #566 -- this has been run on cassci for a while and it doesn't look like it fails or hangs:

http://cassci.datastax.com/job/trunk_dtest-skipped-with-require/82/testReport/paging_test/TestPagingSize/test_with_equal_results_to_page_size/history/

This test should be ready to run with normal jobs now.


ptnapoleon:
+1. Go for it

PR #575, opened 2015-10-01
Skip clear/cls tests on Windows

josh-mckenzie:


josh-mckenzie:
Scratch that - I assumed the functionality didn't work but it actually does.

ptnapoleon:
Is the functionality testable though?

josh-mckenzie:
Checking into it now. My assumption is there's just a different chr code we need to compare against on Windows to confirm it's working.

PR #576, opened 2015-10-01
revise test_clear and test_cls on Windows

josh-mckenzie:


ptnapoleon:
I'm +1, but we'll let @mambocab weigh in

mambocab:
Re: just passing if there are no errors: if it's good enough for Josh, it's good enough for me.

Re `sys.platform == 'win32'`: we check if the test is running under Windows in a couple different ways in the codebase. Sometimes we just check to see if `sys.platform == 'win32'`, as in the PR, but other times we use [`ccmlib.common.is_win`](https://github.com/pcmanus/ccm/blob/master/ccmlib/common.py#L255), which instead checks `sys.platform in ("cygwin", "win32")` . Do we know how this test behaves under cygwin?

josh-mckenzie:
I don't have cygwin installed anymore because of how badly it can (and does) screw up permissions on files in Windows. I'd personally be comfortable with us dropping any attempts at formal cygwin support, but that being said, standardizing on a single way to do something in the code-base is optimal. I'll update the check to use common.is_win.

josh-mckenzie:
Force-pushed an update to use is_win()

mambocab:
Awesome, thanks.

The linter warning from Travis isn't related to this PR. Merging.

PR #577, opened 2015-10-02
Enhancements to cqlsh copy tests for CASSANDRA-9304

stef1927:


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -16,12 +16,18 @@
...
 from tools import rows_to_list, since
 
 DEFAULT_FLOAT_PRECISION = 5  # magic number copied from cqlsh script
 DEFAULT_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'  # based on cqlsh script; timezone stripped
 
+PARTITIONERS = dict()
+PARTITIONERS["murmur3"] = "org.apache.cassandra.dht.Murmur3Partitioner"
+PARTITIONERS["random"] = "org.apache.cassandra.dht.RandomPartitioner"
+PARTITIONERS["byte"] = "org.apache.cassandra.dht.ByteOrderedPartitioner"
+PARTITIONERS["order"] = "org.apache.cassandra.dht.OrderPreservingPartitioner"
```
mambocab: Using a dictionary literal is better style here:

```python
PARTITIONERS = {
    "murmur3": "org.apache.cassandra.dht.Murmur3Partitioner",
    "random": "org.apache.cassandra.dht.RandomPartitioner",
    "byte": "org.apache.cassandra.dht.ByteOrderedPartitioner",
    "order": "org.apache.cassandra.dht.OrderPreservingPartitioner"
}
```

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -30,6 +36,11 @@ class CqlshCopyTest(Tester):
     Tests the COPY TO and COPY FROM features in cqlsh.
     @jira_ticket CASSANDRA-3906
     """
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
+        self.tempfile = None
```
mambocab: I don't see a need to initialize `self.tempfile` here. If you're running into problems in `tearDown`, then catch an `AttributeError` when you dereference it.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -46,10 +57,16 @@ def tearDown(self):
 
         super(CqlshCopyTest, self).tearDown()
 
-    def prepare(self):
+    def prepare(self, nodes=1, partitioner="murmur3"):
         if not self.cluster.nodelist():
-            self.cluster.populate(1).start(wait_for_binary_proto=True)
-        self.node1, = self.cluster.nodelist()
+            p = PARTITIONERS.get(partitioner)
```
mambocab: If `PARTITIONERS` doesn't have key `partitioner`, `p` will be `None` here. I don't think that'll play well with `set_partitioner`. Is that intentional? You could fail earlier in that case by using `PARTITIONERS[partitioner]`.

mambocab:
Other than my little nits, these look great, thank you!

stef1927:
Thank you for your comments, they should be fixed now!

mambocab:
Excellent, thank you! Merging.

PR #580, opened 2015-10-05
Check if Windows on assert for existence of NotSslRecordException

jkni:
This PR adds an OS check around an assert checking for the existence of a log entry, as this log entry does not appear on Windows.

I've tested on Windows and Linux, as well as applied PEP 8 fixes from autopep8.

This issue is tracked in [CASSANDRA-10388](https://issues.apache.org/jira/browse/CASSANDRA-10388).

### DIFF
--- a/native_transport_ssl_test.py
+++ b/native_transport_ssl_test.py
@@ -3,6 +3,7 @@ from dtest import Tester
 from cassandra import ConsistencyLevel
 from cassandra.cluster import NoHostAvailable
 from tools import generate_ssl_stores, putget
+from ccmlib.common import is_win
 
 
 class NativeTransportSSL(Tester):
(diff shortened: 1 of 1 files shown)


mambocab:
Any idea why? I just want to double-check this isn't evidence of a bug. If we're certain this isn't a bug, could you add a comment explaining why this isn't logged on windows?

Do you know if there are any equivalent checks we can make on Windows? or is the assertion above it, making sure that the connection fails, sufficient?

native_transport_ssl_test.py
```diff
@@ -28,8 +29,9 @@ def connect_to_ssl_test(self):
         except NoHostAvailable:
             pass
 
-        assert len(node1.grep_log("^io.netty.handler.ssl.NotSslRecordException.*")) > 0, \
-            "Missing SSL handshake exception while connecting with non-SSL enabled client"
+        if not is_win():
+            assert len(node1.grep_log("^io.netty.handler.ssl.NotSslRecordException.*")) > 0, \
```
ptnapoleon: We should also just clean this assertion up while we're here, by using the unittest asserts

mambocab: +1, but that's our problem, not Joel's. I can fix it up after this is finalized an merged?

jkni:
I'm digging in now to be extremely confident this isn't a bug on our side, but it looks like it happens in Netty's SSL handling on Windows. I'll add a comment with what I find.

There isn't an equivalent check unless Netty throws something else that we log at TRACE, which I'm checking now. From a test semantics standpoint, the assertion above it making sure the connection fails is essentially equivalent (slightly less detailed, but that may be a sacrifice we make here).

Let's leave this open until I update.

Thanks.

jkni:
Okay - so, in the meantime, I've done some digging into this cause. It looks like the missing log lines was actually due to differences in Windows/Linux logging in the fallout of the debug.log/system.log change in C*.  Yuki has contributed fixes for this.

With the latest CCM and no changes to dtests or C*, the tests pass with the assert left in.

Thanks for the pushback on this; I wasn't being persistent enough in root-causing. Nice work.

PR #581, opened 2015-10-05
Clean up ssl tests, possibly begin using `nose.tools` assertions

mambocab:
Depends on #580, @ptnapoleon to review.

Since this is a straight-up refactor (so, not urgent, style discussions are appropriate and worth blocking merge), I thought it might be a good place to start the `self.assertThing` vs `nose.tools.assert_thing` discussion. Philip and I had talked about how it would be nice to use the `nose.tools` after Andrew ran into problems writing custom assert functions that weren't methods. Example in action [on this line](https://github.com/riptano/cassandra-dtest/compare/master...mambocab:clean-up-ssl-tests?expand=1#diff-256d3a62cfb3c3d440953958032f15f1R35). 

I propose we start using them for new code and changed tests, and mark it as such in CONTRIBUTING. I like them better because they allow us to more easily write custom asserts. They also take up less horizontal space, which is good for readability in deeply nested code. More generally, I find them more attractive and readable, which adds up when working in the dtests all day.

The downside that I can think of is that they do introduce a dependency on `nose` in the test code itself, while the `self.assertX` methods just depend on `unittest`. That said, if we ever move off `nose`, we could roll our own replacement for this part of `nose.tools` easily; it's just a few lines of metaprogramming.

### DIFF
--- a/native_transport_ssl_test.py
+++ b/native_transport_ssl_test.py
@@ -19,11 +19,12 @@ class NativeTransportSSL(Tester):
         cluster = self._populateCluster(enableSSL=True)
         node1 = cluster.nodelist()[0]
 
-        # try to connect without ssl options
-        try:
-            cluster.start()
-            session = self.patient_cql_connection(node1)
-            assert False, "Should not be able to connect to SSL socket without SSL enabled client"
+        cluster.start()
+
+        try:  # hack around assertRaise's lack of msg parameter
+            # try to connect without ssl options
+            self.patient_cql_connection(node1)
+            self.fail('Should not be able to connect to SSL socket without SSL enabled client')
         except NoHostAvailable:
             pass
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Well I'm +1 for this code, and I support the idea of switching to nose.tools, because I don't see use moving off nose anytime soon, and it wouldn't be that hard to replace this dependency if we ever did. But we should still get the input of the rest of @riptano/cassandra_test_engineering

mambocab:
Made a few changes -- in particular, the change that used `self.assertGreater` (which I had changed to `nose.tools.assert_greater`) isn't necessary anymore, so the `self.assertX` vs. `nose.tools.assert_x` debate isn't relevant here anymore.

I don't know why I said this was a refactor; it's not. Previously, if the cluster failed to start by raising a `NoHostAvailable` exception, the test would pass, even though the behavior we cared about was when attempting to connect. Not sure that matters, but this test should be a little more precise.

@ptnapoleon could you please rereview?

ptnapoleon:
+1, though are you sure start() should be succeeding?

mambocab:
I was sure, but dangit, Philip, now I have to think about it :P 

(My reasoning was that we're not waiting on the binary protocol, so it should just be starting the process and returning, but I'm not 100% sure that's the behavior. That said, there's a call to `start` that's expected to succeed [in this test](https://github.com/riptano/cassandra-dtest/blob/master/native_transport_ssl_test.py#L65).)

ptnapoleon:
Okay, makes sense given the described alleged behavior of that test and this one.

mambocab:
Thanks for the review.

FTR, the `native_transport_ssl_test` tests pass locally on 2.1+. Merging.

PR #582, opened 2015-10-06
short circuit self.data access

mambocab:
self.data only exists after calling all_datatypes_prepare, so when looking it up on self, check for it before doing so.

This should deal with the recent failures in the copy tests on CassCI:

http://cassci.datastax.com/view/trunk/job/cassandra-3.0_dtest/lastCompletedBuild/testReport/cqlsh_tests.cqlsh_copy_tests/CqlshCopyTest/test_null_as_null_indicator/

http://cassci.datastax.com/view/trunk/job/cassandra-3.0_dtest/lastCompletedBuild/testReport/cqlsh_tests.cqlsh_copy_tests/CqlshCopyTest/test_undefined_as_null_indicator/

I wasn't able to reproduce the original failures locally, but this change should be safe. @nutbunnies You touched this last, right? Could you review?

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -179,7 +179,7 @@ class CqlshCopyTest(Tester):
         encoding_name = 'utf-8'  # codecs.lookup(locale.getpreferredencoding()).name
 
         # this seems gross but if the blob isn't set to type:bytearray is won't compare correctly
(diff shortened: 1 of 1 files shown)


mambocab:
Oh, and this is for [CASSANDRA-10454](https://issues.apache.org/jira/browse/CASSANDRA-10454).

nutbunnies:
LGTM

mambocab:
Thanks, merging.

PR #583, opened 2015-10-06
Fix batch_test.py and compression_test.py after logging changes (CASSANDRA-10241)

pauloricardomg:
Fixes CASSANDRA-10453 and CASSANDRA-10451

### DIFF
--- a/batch_test.py
+++ b/batch_test.py
@@ -73,7 +73,7 @@ class TestBatch(Tester):
                                  "involved in an atomic batch might cause batchlog entries to expire "
                                  "before being replayed.")
         debug(warning)
-        self.assertEquals(1, len(warning), "Cannot find the gc_grace_seconds warning message.")
+        self.assertTrue(len(warning) > 0, "Cannot find the gc_grace_seconds warning message.")
 
     @since('3.0')
     def logged_batch_gcgs_below_threshold_multi_table_test(self):
@@ -100,7 +100,7 @@ class TestBatch(Tester):
                                  "involved in an atomic batch might cause batchlog entries to expire "
                                  "before being replayed.")
         debug(warning)
-        self.assertEquals(1, len(warning), "Cannot find the gc_grace_seconds warning message.")
+        self.assertTrue(len(warning) > 0, "Cannot find the gc_grace_seconds warning message.")
 
     @since('3.0')
     def unlogged_batch_gcgs_below_threshold_should_not_print_warning_test(self):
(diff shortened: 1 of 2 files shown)


batch_test.py
```diff
@@ -100,7 +100,7 @@ def logged_batch_gcgs_below_threshold_multi_table_test(self):
                                  "involved in an atomic batch might cause batchlog entries to expire "
                                  "before being replayed.")
         debug(warning)
-        self.assertEquals(1, len(warning), "Cannot find the gc_grace_seconds warning message.")
+        self.assertTrue(warning, "Cannot find the gc_grace_seconds warning message.")
```
mambocab: Using `assert{True,False}(x)` where `x` is a list is not very readable, I feel; it works, but checking that `len(x) == 0` is clearer, given the type.

mambocab:
Other than my little nit, which you may ignore if you think I'm wrong, +1.

On Jira, you said

> Basically ccm/dtest merges system.log and debug.log, so info/warn/error statements are duplicated. Modified tests to only assert logs are present, and not count them.

This is fine for these tests -- it's not going to be a problem if more than one warning is logged. However, for other tests, we may want to make sure something gets logged some exact number of times. Will the dtest/ccm keep us from doing that when we need to? @ptnapoleon You're overseeing the ccm changes for this, right? Could you have a look, please?

ptnapoleon:
Well we did make some changes to split the logs out, I thought.

pauloricardomg:
well, I didn't update ccm on my machine, so that's maybe why I was able to reproduce the test while @mambocab couldn't. is it possible ccm was not updated on cassci?

ptnapoleon:
No, ccm is always installed from scratch each test run. I'll see if I can reproduce locally.

pauloricardomg:
made suggested changes. couldn't reproduce the issue on current dtest master after upgrading ccm, so would be nice to check if ci is updating ccm correctly.

mambocab:
These tests fail, for I believe the same reason:

http://cassci.datastax.com/view/trunk/job/cassandra-3.0_dtest/lastCompletedBuild/testReport/deprecated_repair_test/TestDeprecatedRepairAPI/

https://github.com/riptano/cassandra-dtest/blob/master/deprecated_repair_test.py#L175

Skimming the test logs, I don't see how we could be using an old version of ccm.

pauloricardomg:
:four_leaf_clover:

mambocab:
These tests are passing now. Thanks for your help, Paulo!

PR #586, opened 2015-10-07
Fix default upgrade logic

mambocab:
This fixes the logic that determines the default upgrade path for the tests in `./upgrade_tests/`. Previously, choosing the sensible default required that the `UPGRADE_TO` env variable be set, which is the opposite of what we want.

It also adds a parameter to the function that detects the Cassandra git branch so that you can pass in an argument. This parameter has a default value, and the function behaves identically with no argument specified, so it shouldn't affect other callers.

### DIFF
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -31,7 +31,7 @@ UPGRADE_TO = os.environ.get('UPGRADE_TO', None)
 UpgradePath = namedtuple('UpgradePath', ('starting_version', 'upgrade_version'))
 
 
-def get_default_upgrade_path(job_version):
+def get_default_upgrade_path(job_version, cdir=None):
     """
     Given a version (which should be specified as a LooseVersion,
     StrictVersion, or NormalizedVersion object), return a tuple (start, target)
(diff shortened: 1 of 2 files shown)


upgrade_tests/upgrade_base.py
```diff
@@ -51,7 +52,10 @@ def get_default_upgrade_path(job_version):
         upgrade_version = 'binary:2.2.0'
     elif '3.0' <= job_version < '3.1':
         # We can choose 2.2 because it will run on JDK 1.8
-        start_version = 'binary:2.2.0'
+        if cassandra_git_branch(cdir=cdir) == 'trunk':
+            start_version = 'binary:3.0.0-rc1'
+        else:
+            start_version = 'binary:2.2.0'
```
ptnapoleon: Why 2.2.0? 2.2.2 was just released.

mambocab: Because we're testing, e.g., that (released 2.2) -> (3.0 HEAD) works, and this needs to work for arbitrary 2.2. My thought was that 2.2.0 will be the most-diverged-from version, so we should compare against that by default.

ptnapoleon: I think I'd like to vet this in #dev, but I think you have it backwards. It's my understanding that (prior to tick tock), we only support upgrading from X.Y.latest, rather than arbitrary X.Y.

mambocab:
Based on feedback from Tyler in #dev, I'm using 2.2.1. That may change based on [CASSANDRA-10488](https://issues.apache.org/jira/browse/CASSANDRA-10488), which I'll keep an eye on.

Any other comments?

ptnapoleon:
Nope, +1

mambocab:
Thanks!

PR #589, opened 2015-10-08
Changes for second part of CASSANDRA-7066

stef1927:
In the second installment of CASSANDRA-7066, `sstablelister` was renamed to `sstableutil` and the location of the txn log files was changed. However the dtests were never 'unrequired' and so we forgot to merge the fix for them. This patch addresses these changes, and this includes the fix for sstableverify_test (CASSANDRA-10461). It also 'unrequires' the tests.

### DIFF
--- a/offline_tools_test.py
+++ b/offline_tools_test.py
@@ -255,14 +255,14 @@ class TestOfflineTools(Tester):
     def _get_final_sstables(self, node, ks, table):
         """
         Return the node final sstable data files, excluding the temporary tables.
-        If sstablelister exists (>= 3.0) then we rely on this tool since the table
+        If sstableutil exists (>= 3.0) then we rely on this tool since the table
         file names no longer contain tmp in their names (CASSANDRA-7066).
         """
         # Get all sstable data files
         allsstables = node.get_sstables(ks, table)
 
         # Remove any temporary files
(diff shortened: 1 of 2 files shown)


ptnapoleon:
Fantastic changes, +1. Do these tests pass though?

mambocab:
On trunk, I'm seeing failures on `offline_tools_test.py:TestOfflineTools.sstableofflinerelevel_test` and `offline_tools_test.py:TestOfflineTools.sstableverify_test`:

https://gist.github.com/mambocab/0f453bb8a31a7fe692ec

So, I guess we can leave it to @stef1927 to decide if she wants to fix and merge, or merge and file tickets and fix them after?

mambocab:
`offline_tools_test.py:TestOfflineTools.sstableofflinerelevel_test` passes on 3.0, but `offline_tools_test.py:TestOfflineTools.sstableverify_test` fails.

stef1927:
`sstableverify_test` on 3.0+ should pass when this patch is merged, if that's not the case we can log it to CASSANDRA-10461. `sstableofflinerelevel_test` is unrelated to this patch and it seems to pass locally on my box and in the recent 3.0 jobs on jenkins. If it is flacky we can either open a new ticket or add it to 10461 so I prefer to merge this now.

PR #590, opened 2015-10-08
Fix jmxmetrics_test usage of jmx port

carlyeks:
ccm lib will already add the port, so we don't need to specify it here.


ptnapoleon:
I think ccm didn't always add that port. If we leave in the redundancy, is it causing problems? If so, +1, merge this. If not, no harm in leaving this in for clarity.

carlyeks:
Appending that port causes stress to die with `-port is defined multiple times. Each option/command can be specified at most once.` which is why the metrics are the same before and after the stress run.

mambocab:
Just for record-keeping: this is for [CASSANDRA-10457](https://issues.apache.org/jira/browse/CASSANDRA-10457).

PR #591, opened 2015-10-08
Watch debug.log for compaction logs

yukim:
Fix for CASSANDRA-10463.
This change depends on ccm change here(https://github.com/pcmanus/ccm/pull/386).

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -17,6 +17,12 @@ class TestCompaction(Tester):
         kwargs['cluster_options'] = {'start_rpc': 'true'}
         Tester.__init__(self, *args, **kwargs)
 
+    def setUp(self):
+        Tester.setUp(self)
+        # compaction test for version 2.2.2 and above relies on DEBUG log in debug.log
+        if self.cluster.version() > '2.1':
+            self.cluster.set_log_level("DEBUG")
+
     @since('0', '2.2.X')
     def compaction_delete_test(self):
         """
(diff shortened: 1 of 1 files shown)


compaction_test.py
```diff
@@ -287,11 +293,15 @@ def disable_autocompaction_nodetool_test(self):
             session.execute('insert into to_disable (id, d) values ({0}, \'{1}\')'.format(i, 'hello' * 100))
             if i % 100 == 0:
                 node.flush()
-        self.assertTrue(len(node.grep_log('Compacting.+to_disable')) == 0, 'Found compaction log items for {0}'.format(self.strategy))
+        if node.get_cassandra_version() < '2.2':
+            log_file = 'system.log'
+        else:
+            log_file = 'debug.log'
```
mambocab: This logic is repeated a lot -- you might pull it out into a top-level `debug_log_for_version` function or something.

compaction_test.py
```diff
@@ -17,6 +17,12 @@ def __init__(self, *args, **kwargs):
         kwargs['cluster_options'] = {'start_rpc': 'true'}
         Tester.__init__(self, *args, **kwargs)
 
+    def setUp(self):
+        Tester.setUp(self)
+        # compaction test for version 2.2.2 and above relies on DEBUG log in debug.log
+        if self.cluster.version() > '2.1':
```
mambocab: The version here is 2.1, but your comment says 2.2.2. Is that intentional?

yukim: Greater than 2.1, so 2.2.
CASSANDRA-10241 was introduced in 2.2.2.
I don't think minor version matters but I just wanted to note it.

mambocab: Ah, I see, thanks.

mambocab:
My nit is minor, so +1 whether you factor out the version logic or not. Waiting on the CCM PR before merge.

ptnapoleon:
I merged the CCM PR, but do we want to consider adding functions that grep *both* log files? Rather than either one?

mambocab:
@ptnapoleon What's the case where we don't care which log file it goes to? Using a function that checks both could make us miss bugs where something is logged to the wrong file.

ptnapoleon:
Makes sense. I concede.

mambocab:
Cool, thanks, Yuki and Philip. Merging.

PR #593, opened 2015-10-08
do node version check in limit_multiget_test

mambocab:
This doesn't compare the node version, so it always executes the code meant for 2.2 nodes. I believe this causes a failure here:

http://cassci.datastax.com/view/Upgrades/job/storage_engine_upgrade_dtest-21_HEAD-22_tarball/20/testReport/upgrade_tests.cql_tests/TestCQL/limit_multiget_test/

On the 2.1 -> 2.2 upgrade path. @nutbunnies, can you please review?


nutbunnies:
Doh, my bad.  LTGM +1

mambocab:
Schweet, thanks! Merging.

PR #595, opened 2015-10-12
Add note on installing latest pyhon-driver for Cassandra 3.x

spodkowinski:
Had to spend some time to figure out why tests failed with an `NoHostAvailable: "unconfigured table schema_keyspaces"` with Cassandra 3. Looks like you need to use the latest python driver alpha release to make the tests work. Would be nice to have that actually documented somewhere.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -63,7 +63,10 @@ will often need to modify them in some fashion at some later point:
 * python-driver
 
         cd ~/git/cstar
+        Cassandra 2.x:
         sudo pip install cassandra-driver
+        Cassandra 3.x (requires latest python-driver):
+        sudo pip install --pre cassandra-driver
         For more instructions on how to install the python-driver,
         see http://datastax.github.io/python-driver/installation.html
 


ptnapoleon:
For now actually, the cassandra-test branch of the python-driver is the most up to date. @mambocab should we document that, or leave @spodkowinski's changes?

mambocab:
Good question. My initial reaction is to document using the `cassandra-test` branch, since it was branched off for the dtest use case. I don't know what kinds of guarantees we have w.r.t. that branch sticking around forever/being the most up-to-date version forever, though.

Would it be reasonable to document the `cassandra-test` branch as a first choice and `--pre` as a fallback?

ptnapoleon:
+1 to that. Furthermore, once 3.0 goes GA, the appropriate driver version should go GA, and we shouldn't have this issue again, given the pace of tick-tock.

mambocab:
Added some changes on top of this and filed as #598; discussion should move there.

PR #596, opened 2015-10-12
Combine commitlog test files

jkni:
Around a month ago, we introduced a commit_log_test.py file containing commitlog tests. We already have a commitlog_test.py. This PR combines the new tests introduced in commit_log_test.py into commitlog_test.py.

I've run the tests to confirm no functionality difference was introduced.


mambocab:
I'm +1 on this. @ptnapoleon one last check, do you know of any reason to keep these in separate files?

mambocab:
Also, thanks @jkni!

ptnapoleon:
There is no good reason, so if you're +1 on the code, I'm merging.

PR #599, opened 2015-10-12
Add tests for paging with only static columns containing data

blerer:
DTest changes for CASSANDRA-10381

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -907,6 +907,28 @@ class TestPagingData(BasePagingTester, PageAssertionMixin):
         self.assertEqual(pf.num_results_all(), [400, 200])
         self.assertEqualIgnoreOrder(expected_data, pf.all_data())
 
+    def static_columns_with_empty_non_static_columns_paging_test(self):
+        """
+        @jira_ticket CASSANDRA-10381.
+        """
+
+        session = self.prepare()
+        self.create_ks(session, 'test_paging_static_cols', 2)
+        session.execute("CREATE TABLE test (a int, b int, c int, s int static, PRIMARY KEY (a, b))")
+        session.row_factory = named_tuple_factory
+
+        for i in range(10):
+            session.execute("UPDATE test SET s = %d WHERE a = %d" % (i, i))
+
+        session.default_fetch_size = 2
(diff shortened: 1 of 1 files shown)


paging_test.py
```diff
@@ -907,6 +907,28 @@ def bool_from_str_int(text):
...
+        session.execute("CREATE TABLE test (a int, b int, c int, s int static, PRIMARY KEY (a, b))")
+        session.row_factory = named_tuple_factory
+
+        for i in range(10):
+            session.execute("UPDATE test SET s = %d WHERE a = %d" % (i, i))
+
+        session.default_fetch_size = 2
+        results = list(session.execute("SELECT * FROM test"))
+        self.assertEqual(10, len(results))
+        self.assertEqual([0, 1, 2, 3, 4, 5, 6, 7, 8, 9] * 1, sorted([r.s for r in results]))
```
mambocab: Since you're comparing the list against `[0, 1, 2, 3...`, the `self.assertEqual(10, len(results))` comparisons are redundant.

The `* 1` applied to the lists here and below have no effect.

paging_test.py
```diff
@@ -907,6 +907,28 @@ def bool_from_str_int(text):
...
+        @jira_ticket CASSANDRA-10381.
+        """
+
+        session = self.prepare()
+        self.create_ks(session, 'test_paging_static_cols', 2)
+        session.execute("CREATE TABLE test (a int, b int, c int, s int static, PRIMARY KEY (a, b))")
+        session.row_factory = named_tuple_factory
+
+        for i in range(10):
+            session.execute("UPDATE test SET s = %d WHERE a = %d" % (i, i))
```
mambocab: Could you please switch to using `.format` instead of `%` as described in [CONTRIBUTING](https://github.com/riptano/cassandra-dtest/blob/master/CONTRIBUTING.md)?

mambocab:
Other than a couple small nits I've noted inline, this looks good as far as style goes. Thanks! Let us know when you'd like this merged.

blerer:
CASSANDRA-10381 as been marked as fix. If everything is fine on your side you can push the new DTest

PR #600, opened 2015-10-12
added test checking for ALLOW FILTERING when using multiple predicates

nutbunnies:


secondary_indexes_test.py
```diff
@@ -265,6 +265,37 @@ def wait_for_schema_agreement(self, session):
...
+        session.execute("USE ks;")
+        session.execute("CREATE TABLE tbl (id uuid primary key, c0 text, c1 text, c2 text);")
+        session.execute("CREATE INDEX ix_tbl_c0 ON tbl(c0);")
+        session.execute("CREATE INDEX ix_tbl_c1 ON tbl(c1);")
+        session.execute("INSERT INTO tbl (id, c0, c1, c2) values (uuid(), 'a', 'b', 'c');")
+        session.execute("INSERT INTO tbl (id, c0, c1, c2) values (uuid(), 'a', 'b', 'c');")
+        session.execute("INSERT INTO tbl (id, c0, c1, c2) values (uuid(), 'q', 'b', 'c');")
+        session.execute("INSERT INTO tbl (id, c0, c1, c2) values (uuid(), 'a', 'e', 'f');")
+        session.execute("INSERT INTO tbl (id, c0, c1, c2) values (uuid(), 'a', 'e', 'f');")
+        time.sleep(4)
```
mambocab: Unless I'm missing something, this shouldn't be necessary; the final `INSERT` should block. If it is necessary, could you add a comment explaining why?

mambocab:
Looks great other than the unexplained `sleep`. With that commented out or removed, I'm +1.

mambocab:
Pfft, not commented out, commented on.

nutbunnies:
Removed the un-need sleep.  Also added a test to use nodetool rebuild_index.  Please give the new method a once over. Thanks

secondary_indexes_test.py
```diff
@@ -266,6 +267,69 @@ def wait_for_schema_agreement(self, session):
...
+
+        before_files = os.listdir(index_sstables_dir)
+
+        node1.nodetool("rebuild_index keyspace1 standard1 ix_c0")
+
+        debug("waiting for index to rebuild")
+        time.sleep(10)
+
+        after_files = os.listdir(index_sstables_dir)
+        self.assertNotEqual(before_files, after_files)
```
mambocab: This is comparing the output of `os.listdir`, which doesn't make any guarantees about the order of its contents:

https://docs.python.org/2/library/os.html#os.listdir

Could you change this comparison to `self.assertNotEqual(set(before_files), set(after_files))` so order won't matter?

mambocab:
Other than my ordering comment, ready to go :+1:

PR #601, opened 2015-10-12
Test to reproduce CASSANDRA-10231 that crashes nodes during decommission

jkni:
This PR includes tests added to topology_test that check for a null status entry on nodes that crash during the decommission of the other node. This test is necessary to reproduce [CASSANDRA-10231](https://issues.apache.org/jira/browse/CASSANDRA-10231), which is not covered by any preexisting tests.

Since topology_test was flagged as @no_vnodes at a class level, I moved this down to the scope of the preexisting methods and added the test as a new method that is not decorated.

The commit shows up as Stefania since she did the hard work of first writing this test; that said, I've rebased with nits and fixes, so I'll be responsible for any feedback.

I've run through autopep8 for low-hanging fruit.

Thanks!

### DIFF
--- a/topology_test.py
+++ b/topology_test.py
@@ -2,14 +2,16 @@ from dtest import Tester
 from tools import insert_c1c2, query_c1c2, no_vnodes, new_node, debug, since
 from assertions import assert_almost_equal
 
+import re
 import time
 from ccmlib.node import TimeoutError
 from cassandra import ConsistencyLevel
+from threading import Thread
 
 
-@no_vnodes()
 class TestTopology(Tester):
 
+    @no_vnodes()
     def movement_test(self):
         cluster = self.cluster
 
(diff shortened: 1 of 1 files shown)


topology_test.py
```diff
@@ -180,3 +185,63 @@ def decommissioned_node_cant_rejoin_test(self):
...
+        """
+        cluster = self.cluster
+        cluster.populate(3).start(wait_other_notice=True)
+
+        node1, node2 = cluster.nodelist()[0:2]
+
+        t = DecommissionInParallel(node1)
+        t.start()
+
+        p = re.compile(".N(?:\s*)127\.0\.0\.1(?:.*)null(?:\s*)rack1")
```
ptnapoleon: Can you rename this variable from `p`?

jkni: I've pushed a commit with the rename.

ptnapoleon:
@mambocab any objections? I'm +1

mambocab:
Looks good to me, thanks Stefania and @jkni!

PR #604, opened 2015-10-14
Fixes for CASSANDRA-10461

stef1927:
As discussed on CASSANDRA-10461, to better debug disk space issues such as [this] (http://cassci.datastax.com/job/cassandra-3.0_dtest/248/testReport/offline_tools_test/TestOfflineTools/sstablelevelreset_test/), I've changed `sstablelevelreset_test` to first look at the exit code and `stderr`, and only afterwards validate `stdout`.

I've also improved splitting of `stdout` into lines for `SSTableUtilTest` since it passes locally but fails on Jenkins due to an extra line; so my best guess is a line ending problem.

### DIFF
--- a/offline_tools_test.py
+++ b/offline_tools_test.py
@@ -39,7 +39,7 @@ class TestOfflineTools(Tester):
         cluster.stop(gently=False)
 
         (output, error, rc) = node1.run_sstablelevelreset("keyspace1", "standard1", output=True)
-        debug(error)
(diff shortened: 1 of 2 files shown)


mambocab:
Looks good, +1.

stef1927:
I swapped the lines checking the error code back to the end since the error code is less helpful than the error message, merging now thanks.

PR #606, opened 2015-10-14
Add test for EACH_QUORUM reads

carlyeks:
For CASSANDRA-9602

### DIFF
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -301,6 +301,7 @@ class TestAvailability(TestHelper):
             (ConsistencyLevel.QUORUM, ConsistencyLevel.QUORUM),
             (ConsistencyLevel.LOCAL_QUORUM, ConsistencyLevel.LOCAL_QUORUM),
             (ConsistencyLevel.EACH_QUORUM, ConsistencyLevel.LOCAL_QUORUM),
+            (ConsistencyLevel.LOCAL_QUORUM, ConsistencyLevel.EACH_QUORUM),
             (ConsistencyLevel.ONE, ConsistencyLevel.ONE, None, False),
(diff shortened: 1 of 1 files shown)


mambocab:
Any reason this shouldn't be added to the other multi-dc tests, in particular `test_network_topology_strategy_users` and `test_network_topology_strategy_counters`?

carlyeks:
Actually, these tests need to be separated because they will only run on 3.0+. I'll post another patch that separates out the each_quorum reads tests.

consistency_test.py
```diff
@@ -482,7 +520,7 @@ def run():
 
     def test_simple_strategy_users(self):
         """
-        Test for a single datacenter, users table.
+        Test for a single datacenter, users table, only the each quorum reads.
```
mambocab: Is this supposed to be "except for the each quorum reads"?

mambocab:
Other than my note on the comment, everything looks good; ping me and I can merge. Thanks!

mambocab:
Just to follow up: am I right that that comment is incorrect? Happy to fix it myself.

consistency_test.py
iamaleksey: Looks correct to me.

mambocab: Can you or @carlyeks explain? I guess I'm obtuse today, I don't see it. Looks to me like this test tests everything but each_quorum reads, plus makes sure each quorum calls don't fall over, then [test_simple_strategy_each_quorum_users](https://github.com/riptano/cassandra-dtest/pull/606/files#diff-8e908af48367adae7218567e819d6ec1R553) tests only the each_quorum reads.

PR #608, opened 2015-10-14
add streaming_timeout_in_ms to resumable_bootstrap_test

yukim:
Potential fix to [CASSANDRA-10167](https://issues.apache.org/jira/browse/CASSANDRA-10167).

I couldn't reproduce flapping in my laptop, but potential hang may be caused by streaming waiting to be timed out for 1 hour (default value).
This patch sets `streaming_socket_timeuout_in_ms` so that test won't hang and time out.

Can you try running this in CI env to see this will fix the problem?


mambocab:
Can do, merging. Would you like this run on any special jobs, or do you just want to wait and see what happens on the normal CassCI jobs?

yukim:
Thanks.
I think running on  normal CI job would be fine.
Change shouldn't do any harm on regular CI run.
Let's see if this will fix flapping.

PR #611, opened 2015-10-15
Test for Cassandra 7069

ptnapoleon:
I wrote this test forever ago, but didn't merge it because of https://issues.apache.org/jira/browse/CASSANDRA-9484 Well now it isn't failing anymore on 2.2 or 3.0, so we should merge this in.

@mambocab to review.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -8,6 +8,7 @@ import re
 from dtest import Tester, debug
 from tools import new_node, query_c1c2, since, KillOnBootstrap, InterruptBootstrap
 from assertions import assert_almost_equal
+from assertions import assert_almost_equal, assert_one
 from ccmlib.node import NodeError
 from cassandra import ConsistencyLevel
 from cassandra.concurrent import execute_concurrent_with_args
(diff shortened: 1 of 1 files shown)


bootstrap_test.py
```diff
@@ -428,3 +426,53 @@ def failed_bootstap_wiped_node_can_join_test(self):
...
+
+        Start a one node cluster and run a stress write workload.
+        Start up a second node, and wait for the first node to detect it has joined the cluster.
+        While the second node is boostrapping, start a third node. This should fail.
+
+        @jira_ticket CASSANDRA-7069
+        """
+
+        bootstrap_error = "Other bootstrapping/leaving/moving nodes detected,"
+        " cannot bootstrap while cassandra.consistent.rangemovement is true"
```
mambocab: You should check this, but I don't think this does what it looks like it does. I think at the end of this, `bootstrap_error == "Other bootstrapping/leaving/moving nodes detected,"`, and you create a `str` that doesn't get assigned to anything. I think you want

```python
bootstrap_error = ("Other bootstrapping/leaving/moving nodes detected,"
                   " cannot bootstrap while cassandra.consistent.rangemovement is true")
```

bootstrap_test.py
```diff
@@ -428,3 +426,53 @@ def failed_bootstap_wiped_node_can_join_test(self):
...
+        node2 = new_node(cluster)
+        node2.start(wait_other_notice=True)
+
+        node3 = new_node(cluster, remote_debug_port='2003')
+        try:
+            process = node3.start(verbose=True)
+            stdout, stderr = process.communicate()
+            if stderr.find(bootstrap_error):
+                debug(stderr)
+                assert True
```
mambocab: `assert True` is a noop.

bootstrap_test.py
```diff
@@ -428,3 +426,53 @@ def failed_bootstap_wiped_node_can_join_test(self):
...
+            if node3.is_running():
```
mambocab: You could replace this whole block with `self.assertFalse(node3.is_running(), msg=...)`.

bootstrap_test.py
```diff
@@ -428,3 +426,53 @@ def failed_bootstap_wiped_node_can_join_test(self):
...
+                assert False, "Node should not have started."
+        except NodeError:
+            assert True
```
mambocab: Another noop. Do we want to fail if there's no `NodeError`? If so, we should use `self.assertRaises`, otherwise just a `pass` here is clearer.

bootstrap_test.py
```diff
@@ -428,3 +426,53 @@ def failed_bootstap_wiped_node_can_join_test(self):
...
+
+        node2.watch_log_for("Starting listening for CQL clients")
+
+        session = self.patient_cql_connection(node2)
+        assert_one(session, "SELECT count(*) from keyspace1.standard1", [500000], cl=ConsistencyLevel.ALL)
```
mambocab: I'd rather have this in a `for _ in range(5)` loop for ease of future editing, but I don't feel strongly. It'd also be nice to have a comment explaining the repetition.

ptnapoleon:
@mambocab, can you review the changed version? I don't seem to ever get the NodeError I was catching before. Though I'm slightly afraid it might happen on cassci. So while you review this, I'll go run these there. I am not sure what do if the Exception is raised, as I don't think I'll be able to get to the stderr to verify the correct error was thrown. Hopefully it's put in system.log

mambocab:
Code looks good. I don't know what the `NodeError` was for in the first place either, so ¯\\_(ツ)_/¯

ptnapoleon:
Well if the node crashes inside the call to start, it should throw a NodeError.

ptnapoleon:
Passed on cassci

PR #612, opened 2015-10-15
tag related jira tickets in consistency_test.py

mambocab:
Figured I'd clean stuff up while I'm poking around in here. I'm not sure that `short_read_test` and `short_read_reversed_test` are primarily related to CASSANDRA-9460, but it is noted in the test itself.


ptnapoleon:
+1

PR #613, opened 2015-10-16
Ignore output lines that are not table file names in sstableutil_test

stef1927:
This fixes sstableutil_test.py so that it works regardless of extra output lines such as `WARN  14:58:01 Only 37512 MB free across all data volumes. Consider adding more capacity to your cluster or removing obsolete snapshots`, see CASSANDRA-10461 for more details.


mambocab:
Looks good and works locally. +1, thanks!

PR #614, opened 2015-10-16
Add test for CASSANDRA-10504

ptnapoleon:
@mambocab to review

### DIFF
--- a/jmx_test.py
+++ b/jmx_test.py
@@ -1,5 +1,7 @@
+import re
 import sys
 import unittest
+import time
 
 from dtest import Tester, debug
 from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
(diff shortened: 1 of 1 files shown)


jmx_test.py
```diff
@@ -124,3 +126,51 @@ def table_metric_mbeans_test(self):
 
             sstables = jmx.read_attribute(sstable_count, "Value")
             self.assertGreaterEqual(int(sstables), 1)
+
+    def test_compactionstats(self):
+        """
+        @jira_ticket CASSANDAR-10504
+        @jira_ticket CASSANDRA-10427
+
+        Test that nodetool compactionstats shows changes to the 'completed' and 'progress' fields.
```
ptnapoleon: I should probably update this description, yeah?

jmx_test.py
```diff
@@ -124,3 +126,51 @@ def table_metric_mbeans_test(self):
...
+
+        cluster = self.cluster
+        cluster.populate(1)
+        node = cluster.nodelist()[0]
+        cluster.set_configuration_options({'concurrent_compactors': 1, 'memtable_cleanup_threshold': 0.01})
+        remove_perf_disable_shared_mem(node)
+        cluster.start(wait_for_binary_proto=True)
+
+        node.nodetool('disableautocompaction')
+        node.stress(['write', 'n=1M'])
```
mambocab: If you want this to actually do 1M operations and not more, I believe you should add a `-rate` argument so it stops when it's done with a single pass, rather than running more to get consistent results.

jmx_test.py
```diff
@@ -124,3 +126,51 @@ def table_metric_mbeans_test(self):
...
+
+            # We want to make sure that the progress is increasing,
+            # and that values other than zero are displayed.
+            self.assertGreater(updated_progress, progress)
+            self.assertGreater(progress, 0)
+            self.assertGreater(updated_progress, 0)
+
+            # Block until the major compaction is complete
+            # Otherwise nodetool will throw an exception
+            while len(jmx.read_attribute(compaction_manager, 'CompactionSummary')) > 0:
```
mambocab: Is there any danger that this will never stop?

jmx_test.py
ptnapoleon: I disagree with this interpretation of the stress behavior.

jmx_test.py
ptnapoleon: Yes, fixing

jmx_test.py
```diff
@@ -124,3 +126,62 @@ def table_metric_mbeans_test(self):
...
+            self.assertGreater(updated_progress, 0)
+
+            # Block until the major compaction is complete
+            # Otherwise nodetool will throw an exception
+            # Give a timeout, in case compaction is broken
+            # and never ends.
+            start = time.time()
+            now = time.time()
+            debug("Waiting for compaction to finish:")
+            while (len(jmx.read_attribute(compaction_manager, 'CompactionSummary')) > 0) and (now - start < 600):
```
mambocab: Much better. It'd be nice not to use `600` as a magic number here; maybe pull out a `max_query_timeout` or something somewhere?

mambocab: You could also call `time.time` in the conditional itself and avoid having that state you need to keep updating.

mambocab:
I've made [a comment on Jira](https://issues.apache.org/jira/browse/CASSANDRA-10504?focusedCommentId=14961151&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14961151):

> Is it possible that the "hanging at 0 progress" problem is caused by compaction finishing too fast? i.e. before you get here:
> https://github.com/riptano/cassandra-dtest/pull/614/files#diff-a5036552850755a7ef3b1b262c2f2727R174
> I don't see any other particular possible problems with the test itself.

ptnapoleon:
I'm not seeing any problems now that I properly set disableautocompaction, so I'll run this a few more times, but then update on Jira and here.

ptnapoleon:
@mambocab Okay, I think this is ready for final review. I've addressed all of your comments so far

mambocab:
LGTM!

PR #615, opened 2015-10-16
use 2.2.2 by default for upgrades

mambocab:
As discussed on #cassandra-dev. @ptnapoleon look good?


ptnapoleon:
Change looks good, but I'd like @thobbs or @blerer to confirm first.

thobbs:
+1 from me.

mambocab:
Thanks, all. Merging.

PR #616, opened 2015-10-16
run upgrade suite 2x with different cluster topos

mambocab:
This makes each upgrade test run twice, each with a different topology. It does this basically by treating the test methods as methods from mixins and adding a bunch of extra classes that are actually run as the tests.

Requesting review from @thobbs and from @ptnapoleon. @ptnapoleon, sorry to ask this of you, but could you double-check that I've made 2 subclasses of each test class in `upgrade_test`? I don't think I missed any, but I want to make sure none of them fell through the cracks. (There may also be a better solution here, open to suggestions.)

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -5187,7 +5187,7 @@ class TestCQL(UpgradeTester):
 
 
 class TestCQLNodes3RF3(TestCQL):
-    NODES, RF, __test__ = 3, 3, True
+    NODES, RF, __test__, CL = 3, 3, True, ConsistencyLevel.ALL
 
 
 class TestCQLNodes2RF1(TestCQL):
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -5184,3 +5184,11 @@ class TestCQL(UpgradeTester):
         for is_upgraded, cursor in self.do_upgrade(cursor):
             debug("Querying %s node" % ("upgraded" if is_upgraded else "old",))
             assert_all(cursor, "SELECT k FROM ks.test WHERE v = 0", [[0]])
+
+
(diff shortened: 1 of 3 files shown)


thobbs:
This looks pretty good to me so far.  If we're not already using ``CL.ALL``, we should set that at the class level for the RF=3 versions.

mambocab:
@thobbs anything wrong with setting the default CL to ALL on all the sessions produced by `do_upgrade`? That seems like the cleanest way to do it, and when RF=1, it shouldn't change behavior at all.

ptnapoleon:
You did not miss any classes. Can you link a cassci run of this branch?

mambocab:
Links here:

http://cassci.datastax.com/job/mambocab_scratch-cassandra-3.0-dtest_extra-upgrade-classes/
http://cassci.datastax.com/job/mambocab_scratch-storage_engine_upgrades-more_upgrade_classes/

One simulates a normal 3.0 run, the other simulates the `upgrade_tests` test run where it tests a single upgrade path. Just started the run, so it'll be a bit before they're done.

ptnapoleon:
This does appear to work, though some/many of the tests fail.

mambocab:
I've pushed another pair of commits making the changes I believe that @thobbs suggested. @thobbs can you confirm that this is what you had in mind? Running the tests again on CassCI; see the same links as before.

Regarding the failures, I reckon the thing to do is to merge and file tickets?

thobbs:
>  @thobbs can you confirm that this is what you had in mind?

Yes, that looks good to me.

ptnapoleon:
Ask in #dev, but I don't object to "merge and file" outright, though I don't exactly support it.

mambocab:
The CassCI jobs seem to be doing ok, so I'll merge and file tickets once the tests have run again.

PR #602, opened 2015-10-13
Catch windowserror git

mambocab:
Previously this code assumed that `git` would be available in the environment running `nosetests`. This is not always true; this was brought to my attention by a user in a desktop Windows environment who uses a `git` shell for some work but `cmd.exe` for `nosetests`. This PR changes the function that shells out to `git` so it returns `None` if you can't shell out.

PR #619, opened 2015-10-19
fix nonsensical behavior in gitless env

mambocab:
This should address the problem @blerer mentioned here:

https://github.com/riptano/cassandra-dtest/pull/602#issuecomment-149192018

I've run this in a `git`-less environement on my machine and it seems to work there. @blerer can you confirm? Sorry for merging #602 before this was ready.

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -304,9 +304,10 @@ def require(require_pattern, broken_in=None):
         return tagging_decorator
     require_pattern = str(require_pattern)
     git_branch = ''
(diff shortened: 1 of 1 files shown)


mambocab:
Pinging @blerer. Does this help with the issues you run into in gitless environments?

blerer:
I get the proper error message. Thanks.

mambocab:
Great, I'll merge. Thanks!

PR #620, opened 2015-10-20
Schema alter test

nutbunnies:


ptnapoleon:
Looks like we missed review on this. Is this up to date, ready for review/merge? You can remove both of those `since` annotations, fyi.

nutbunnies:
@ptnapoleon : yep this should be ready to go -- removed since decorators

schema_test.py
```diff
@@ -1,13 +1,51 @@
...
+            session.execute(stmt1, [n, 'aaa', 'bbb'])
+            if n % 100 == 0:
+                node1.flush()
+
+        session.execute("alter table tbl_o_churn add c2 text")
+        session.execute("alter table tbl_o_churn drop c0")
+        stmt2 = session.prepare("insert into tbl_o_churn (id, c1, c2) values (?, ?, ?);")
+
+        for n in range(1000):
+            session.execute(stmt2, [n + 1000, 'ccc', 'ddd'])
```
ptnapoleon: You could use execute_concurrent_with_args to do these 100 at a time, and speed the test up.

ptnapoleon: Why not use n here, and then make the for loop go from 1001, to 2000?

schema_test.py
```diff
@@ -1,13 +1,51 @@
...
+            if n % 100 == 0:
+                node1.flush()
+
+        for n in range(1000):
+            rows = session.execute("select * from tbl_o_churn where id = %s", (n, ))
+            self.assertEqual(rows[0].c1, 'bbb')
+            self.assertIsNone(rows[0].c2)
+            self.assertFalse(hasattr(rows[0], 'c0'))
+
+            rows = session.execute("select * from tbl_o_churn where id = %s", (n + 1000, ))
```
ptnapoleon: Both this execute and the previous should use `where id = {}".format()` rather than the `%s` form of string interpolation

nutbunnies: `%s` is for the driver statement formatting.  Isn't this preferred over inlining the var with the `.format()` method since it would prevent injection attacks (not that it matters in this case).

nutbunnies:
changed to concurrent executor, which didn't shave much time so I also reduced the size of working set able to get the same result with half the rows.

ptnapoleon:
My bad on the string test, I misread your code.

ptnapoleon:
+1

PR #623, opened 2015-10-22
10461: fixed sstableutil compaction log watching

stef1927:
Compaction logging was moved to debug level and so this test was broken.


mambocab:
Yep, looks good, thanks! +1.

PR #624, opened 2015-10-22
use UPGRADE_TO_DIR env var for upgrade tests

mambocab:
By @pcmanus' request, this allows users to upgrade to a local directory by specifying a `UPGRADE_TO_DIR` environment variable. Yes, the logic is hacky and lives all over the file, but there's a refactor for these tests in the queue anyway, so I don't feel too bad about it.

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -20,7 +20,7 @@ from thrift_bindings.v22.ttypes import ConsistencyLevel as ThriftConsistencyLeve
 from thrift_bindings.v22.ttypes import (CfDef, Column, ColumnOrSuperColumn, Mutation)
 from thrift_tests import get_thrift_client
 from tools import require, rows_to_list, since
-from upgrade_base import UpgradeTester, UPGRADE_TO
+from upgrade_base import UpgradeTester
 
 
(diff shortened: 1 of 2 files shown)


mambocab:
Pushed a commit that uses a corresponding `OLD_CASSANDRA_VERSION` env var. Still not great, but works.

ptnapoleon:
@knifewine to review, maybe?

mambocab:
Now that I think about it, @nutbunnies is familiar with some of this upgrade code. I don't believe @knifewine is. So, @nutbunnies to review, maybe?

nutbunnies:
I haven't tried but I do suspect that if UPGRADE_TO_DIR is used over UPGRADE_TO that [sql_tests.py:TestCQL.test_v2_protocol_IN_with_tuples](https://github.com/riptano/cassandra-dtest/blob/master/upgrade_tests/cql_tests.py#L4098) will have issues.

Other than that looks ok.

mambocab:
Your suspicions are correct, thanks. I've pushed a new commit to fix that; can you re-review @nutbunnies?

nutbunnies:
:+1:

mambocab:
Thanks! Merging.

PR #626, opened 2015-10-23
Extend thrift cas test to cover static columnfamilies (CASSANDRA-10576)

beobal:


beobal:
@mambocab ping for review

thrift_tests.py
```diff
@@ -382,25 +384,48 @@ def test_empty_slice(self):
...
+
+            # now that the partition has been updated, repeating the
+            # operation which expects it to be empty should not succeed
+            cas_result = cas([], first_columns, column_family)
+            assert not cas_result.success
+            # When we CAS for non-existence, current_values is the first live column of the row
+            assert dict((c.name, c.value) for c in cas_result.current_values) == {first_columns[0].name: first_columns[0].value}, cas_result
+
+            # CL.SERIAL for reads
+            # assert client.get('key1', ColumnPath(column_family, column=first_columns[0].name), ConsistencyLevel.SERIAL).column.value == first_columns[0].value
```
mambocab: This should either be uncommented or removed, unless you have some reason to do otherwise.

mambocab:
Looks good, sorry to have let this sit so long. I've run this on supported versions on Linux and Windows, works for me locally. If you have time before you want this merged, could you change the bare `assert`s to `self.assert{False,True,etc}`? and if you don't have time, could you file a quick JIRA for that change and assign it to me?

beobal:
I've added that commented out assertion back (it was temporarily problematic due to CASSANDRA-10586). I'm afraid I've punted on changing the asserts for now, partly because I want to get this in for 3.0 without changing the entire class. The other reason is that there's a bunch of assert helper functions defined which obviously can't be converted to `self.assertX`, so I prefer to leave that to you to decide how to work it. I didn't open a new Jira for it as @ptnapoleon mentioned in https://github.com/riptano/cassandra-dtest/pull/643 that he'd already created such a ticket (but I couldn't find it I'm afraid)

beobal:
squashed & merged offline

mambocab:
Perfect, thanks. And no worries about the asserts -- the cleanup is nice to have while there's already churn, but not expected from folks outside TE.

PR #630, opened 2015-10-28
Better checking of system.IndexInfo when rebuiling indexes

beobal:
See CASSANDRA-10601


mambocab:
Looks good, and passes locally for me on all versions we test in CI. Thanks a lot, merging.

PR #632, opened 2015-10-28
changed index column check since columns no longer exists

nutbunnies:


knifewine:
+1 if it passes, lgtm

PR #635, opened 2015-10-29
added debug and set correct jvm version on dynamic upgrade tests

nutbunnies:


mambocab:
Code looks good; what error does switching JDKs fix?

nutbunnies:
https://cassci.datastax.com/view/Upgrades/job/cassandra_upgrade_proto_v2/lastCompletedBuild/testReport/

It "should" fix all the tests ^ that involve c*2.0.  They were failing to compile since jdk8 isn't compatible with < c*2.1.

mambocab:
Got it, thanks for the explanation :+1:

PR #637, opened 2015-10-29
Tests for CASSANDRA-9912

pauloricardomg:
Also added a test for related issue CASSANDRA-9034.


mambocab:
Looks good, and runs on my machine with the failures described in CASSANDRA-9912. Merging, thanks for the tests!

PR #638, opened 2015-10-30
Minor fixes for CASSANDRA-8653

beobal:
* Remove requires annotation
* Bump connection timeout when checking permissions. During the upgrade
  to 2.2 test, this can timeout with the default value.


ptnapoleon:
Can you remove the import of `require` as well now that it's unused?

ptnapoleon:
+1, after the import fix to make travis happy

PR #639, opened 2015-11-03
Make alter system_auth test less flaky

beobal:


auth_test.py
```diff
@@ -36,6 +34,11 @@ def system_auth_ks_is_alterable_test(self):
 
         self.assertEquals(3, session.cluster.metadata.keyspaces['system_auth'].replication_strategy.replication_factor)
 
+        debug("Repairing system_auth keyspace")
+        for node in self.cluster.nodelist():
+            node.repair(['-pr', 'system_auth'])
```
mambocab: I believe you can also call `self.cluster.nodetool('-pr system_auth')` to achieve the same effect; `Cluster.nodetool` uses the same logic:

https://github.com/pcmanus/ccm/blob/master/ccmlib/cluster.py#L422

auth_test.py
```diff
@@ -607,14 +610,21 @@ def prepare(self, nodes=1, permissions_validity=0):
...
-            self.cluster.nodelist()[0].watch_log_for('Created default superuser')
-        else:
-            # can' just watch for log - the line will appear in just one of the nodes' logs
-            # only one test uses more than 1 node, though, so some sleep is fine.
-            time.sleep(15)
+        time.sleep(10)
+        attempts = 0
+        while not self.find_in_any_log(self.cluster.nodelist(), 'Created default superuser') and attempts < 10:
+            attempts += 1
+            time.sleep(1)
```
mambocab: In the previous logic, `watch_log_for` would block until it found `'Created default superuser'`, or raise a `TimeoutError` if it didn't find it. If I understand the new logic, this will: block until the message is found, then, if the message is found, do nothing, else, do nothing. Is that change intentional?

Style-wise, I think this loop would be more readable and idiomatic as

```python
for _ in range(10):
    if self.find_in_any_log(self.cluster.nodelist(), 'Created default superuser'):
        break
    time.sleep(1)
```

auth_test.py
mambocab: More importantly, what's the purpose of this change? Seems to me that it changes what behavior is being tested; is that what we want to do?

mambocab:
Done with round 1 of review. Thanks for this, it'll be really exciting to see these flaking less.

auth_test.py
ptnapoleon: Wouldn't he need to do `self.cluster.nodetool('repair -pr system_auth')`?

auth_test.py
```diff
@@ -607,14 +610,21 @@ def prepare(self, nodes=1, permissions_validity=0):
...
+
+    def find_in_any_log(self, nodes, pattern):
```
ptnapoleon: Is this really not implemented anywhere else? That's.... unpleasant. Can you make this a helper function and put it somewhere more general?

auth_test.py
ptnapoleon: Yes, it does seem like we no longer fail if we don't find it.

auth_test.py
mambocab: Whoops, you're right -- I left out `repair`.

beobal: tl;dr is that I don't believe this does change what we're actually testing. That is, this is supposed to ensure that modifications to the `system_auth` ks are durable across restarts, which is still the case. What the repair does is to remove the chance of any false negatives caused by CASSANDRA-10655. 

FTR, the way that that causes spurious failures here is that after increasing RF and restarting, the first read (performed when we try to connect to node1) against the roles table is by `PasswordAuthenticator`, which selects only the encrypted password. As this is done at QUORUM, it triggers read repair (as now all 3 nodes should have that row). As per #10655, because this only selects a subset of the row's columns, the values for the ommitted columns are skipped, so when node1 resolves the 3 responses, it determines that `is_superuser` and `can_login` should have empty values & sends out repair mutations to that effect. The very next thing it does is read the roles table again, this time from `CassandraRoleManager` to verify that the user we just authenticated as has the login privilege. As the columns were incorrectly set with the null values, this throws an error from the `ROW_TO_ROLE` function in the role manager, which is logged and causes the test to fail during tear down. The point is that it's the read repair that's broken here, not the altering of the auth keyspace.

beobal: That said, I was manually iterating the nodes for the repair as I was thinking it would be more efficient to just repair what absolutely needed it, and `cluster.repair()` doesn't permit that. That's not really that important though, given that the cluster is basically empty at this point, so I've fixed it as per the suggestion.

auth_test.py
beobal: In the case of the flaky test (`system_auth_ks_is_alterable_test`), where n=3, the logic was just to wait 15s then proceed. If the default user hadn't been created at that point, the test would immediately fail when it tried to establish an authenticated connection. So, the same would be the case for the other, n=1, tests. That said, the timeout is more useful, so I've reworked it to do that (it's not doing the timing perfectly, just good enough).

auth_test.py
beobal: Done, moved to dtest.Tester

ptnapoleon:
+1

auth_test.py
mambocab: Gotcha, thanks for the explanation.

mambocab:
Yep, these look great. Merging.

PR #640, opened 2015-11-04
thrift_hsha: Allow a few seconds to go by while testing for closed sockets

carlyeks:
When running this in a tight loop locally, I can reproduce it very infrequently; if I add the pause here, I haven't seen this ever happen (been running for a few hours).

### DIFF
--- a/thrift_hsha_test.py
+++ b/thrift_hsha_test.py
@@ -68,8 +68,12 @@ class ThriftHSHATest(Tester):
             debug("Closing connections from the client side..")
             for pool in pools:
                 pool.dispose()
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Lgtm. @mambocab ?

mambocab:
This is fine as-is, but it'd be more readable as

```python
time.sleep(3)
stdout = subprocess.Popen(...
```

The downside is that it'd always take 3 seconds to run, instead of sometimes sleeping less or not at all. @ptnapoleon do you have feelings one way or the other? Just a thought; if you prefer it this way go ahead and merge.

carlyeks:
Just to be clear, it will add at least 30 seconds to the run of this test, since this is done inside a loop that runs 10 times.

mambocab:
Ah, thanks @carlyeks, I missed that. In that case, this seems right to me.

PR #641, opened 2015-11-04
workaround for CASSANDRA-10632

mambocab:
The `sstableutil` tests fail on Windows because they compared the output from `sstableutil` and `glob.glob`, which were outputting filenames in different (but still correct, because of Windows tooling's case-insensitivity) casings. In my OpenStack Windows test environment, I'd also get different strings for filenames for the same files because filenames would get abbreviated to their short forms with `~1` when returned by (IIRC) `glob.glob`.

Once we port to Py3, we should be able to can switch to using `os.path.samefile`. For the moment, we'll have to work around it. The solution here is to trim the common prefix from each filename when we get them back from a given tool. The remainder will uniquely identify the file and allow us to make sure the same files are reported by `sstableutil` and `glob.glob`.

This PR also checks if `sstableutil` fails by checking the return code, not checking the `stderr`, because sometimes there are warnings that we don't care about.

In my OpenStack Windows environment, this makes `sstableutil_test.py:SSTableUtilTest.compaction_test` pass, but `sstableutil_test.py:SSTableUtilTest.abortedcompaction_test` still fails when the calculated value of `expected_tempfiles` is an empty list here:

https://github.com/mambocab/cassandra-dtest/blob/10632/sstableutil_test.py#L149

Still, it's an improvement. @josh-mckenzie and @ptnapoleon, can you please review?

### DIFF
--- a/sstableutil_test.py
+++ b/sstableutil_test.py
@@ -12,6 +12,19 @@ KeyspaceName = 'keyspace1'
 TableName = 'standard1'
 
 
+def _strip_common_prefix(strings):
+    strings = list(map(os.path.normcase, strings))
+    common_prefix = os.path.commonprefix(strings)
+    rvs = []
+    for s in strings:
+        stripped = s.lstrip(common_prefix)
+        if s:
+            assert stripped
+        rvs.append(stripped)
+
+    return rvs
+
+
 @since('3.0')
 class SSTableUtilTest(Tester):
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Okay, I understand what is going on, and approve. +1

mambocab:
ping @josh-mckenzie -- do you see any issues with this workaround?

josh-mckenzie:
lgtm

mambocab:
Thanks, all. Merging.

PR #642, opened 2015-11-04
For CASSANDRA-10492, fix test bug in paging_test.py where assert_read…

aweisberg:
…_timeout_or_failure doesn't handle OperationTimedOut

### DIFF
--- a/upgrade_tests/paging_test.py
+++ b/upgrade_tests/paging_test.py
@@ -1696,7 +1696,9 @@ class TestPagingWithDeletions(BasePagingTester, PageAssertionMixin):
         self.allow_log_errors = True
         self.cluster.set_configuration_options(
             values={'tombstone_failure_threshold': 500,
-                    'read_request_timeout_in_ms': 1000}
+                    'read_request_timeout_in_ms': 1000,
+                    'request_timeout_in_ms': 1000,
+                    'range_request_timeout_in_ms': 1000}
         )
         cursor = self.prepare()
         nodes = self.cluster.nodelist()


mambocab:
If I read the driver docs correctly, `OperationTimedOut`s happen on the driver side, while `ReadTimeout`s propagate C* errors. If we want to treat those the same, I'm +1 on the behavioral change.

For the exceptions we `pass` on in that function, could you please move them all to a single `except` block like so:

```python
    except (ReadTimeout, ReadFailure, OperationTimedOut):
        pass
```

Might as well get some style cleanup in while we're churning. Thanks!

aweisberg:
Well now I'm not so sure. The timeout for the read operation is 1 second, but the default timeout in the driver for OperationTimedOut is 10 seconds. Doesn't seem like it should every throw OperationTimedOut.

aweisberg:
Turns out the issue was that the timeout server side was not as small as the test thought so it was hitting the timeout in the driver. I updated the server side configuration to the expected 1 second timeout and now it throws the expected exceptions.

upgrade_tests/paging_test.py
```diff
@@ -24,7 +24,6 @@ def assert_read_timeout_or_failure(session, query):
     except ReadFailure as e:
         pass
 
-
```
mambocab: Can you add this line back, please?

mambocab:
Ran them locally, looks good. Can you add back that whitespace and ping me for merge?

mambocab:
Thanks, merging.

PR #643, opened 2015-11-05
Add read repair test for CASSANDRA-10655

beobal:


read_repair_test.py
```diff
@@ -3,35 +3,103 @@
...
+        debug("Changing RF from 1 to 3")
+        session.execute("""ALTER KEYSPACE alter_rf_test
+                           WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3};""")
+        cl_all_stmt = SimpleStatement("SELECT a FROM alter_rf_test.t1 WHERE k=1",
+                                      consistency_level=ConsistencyLevel.ALL)
+        debug("Executing SELECT on non-initial replica to trigger read repair " + non_replicas[0].name)
+        read_repair_session = self.patient_exclusive_cql_connection(non_replicas[0])
+        res = read_repair_session.execute(cl_all_stmt)
+        # result of the CL ALL query contains only the selected column
+        assert rows_to_list(res) == [[1]], res
```
ptnapoleon: This result assert, and the later one could be cleaned up by using the functions in assertions.py, but I don't actually remember off-hand the syntax for those. I opened a ticket for me to document those, and then clean up everywhere they should have been used. So no need to change these now, but hopefully next time you write a test, they'll be easily available to you.

ptnapoleon:
Looks good, thanks for all the comments, +1

PR #644, opened 2015-11-05
Improve/add MV tests around bootstrapping/adding datacenters

jkni:
This adds/fixes tests for [CASSANDRA-10621](https://issues.apache.org/jira/browse/CASSANDRA-10621) and [CASSANDRA-10634](https://issues.apache.org/jira/browse/CASSANDRA-10634).

A few changes are made:
* in the tests for bootstrapping nodes and adding datacenters, we modify the test so that the values for the partition key are different between the base table and the view table. This prevents nodes from always being paired to themselves in view replica calculation
* we add a test for Cassandra write survey mode
* we make sure to test adding datacenters under NetworkTopologyStrategy and SimpleStrategy

### DIFF
--- a/materialized_views_test.py
+++ b/materialized_views_test.py
@@ -324,15 +324,15 @@ class TestMaterializedViews(Tester):
 
         debug("Writing 1k to base")
         for i in xrange(1000):
-            session.execute("INSERT INTO t (id, v) VALUES ({v}, {v})".format(v=i))
+            session.execute("INSERT INTO t (id, v) VALUES ({id}, {v})".format(id=i, v=-i))
 
         debug("Reading 1k from view")
         for i in xrange(1000):
-            assert_one(session, "SELECT * FROM t_by_v WHERE v = {}".format(i), [i, i])
+            assert_one(session, "SELECT * FROM t_by_v WHERE v = {}".format(-i), [-i, i])
 
         debug("Reading 1k from base")
         for i in xrange(1000):
-            assert_one(session, "SELECT * FROM t WHERE id = {}".format(i), [i, i])
+            assert_one(session, "SELECT * FROM t WHERE id = {}".format(i), [i, -i])
 
         debug("Bootstrapping new node in another dc")
         node4 = new_node(self.cluster, data_center='dc2')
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1 from me. @mambocab ?

materialized_views_test.py
```diff
@@ -312,10 +312,10 @@ def clustering_column_test(self):
         result = list(session.execute("SELECT * FROM ks.users_by_state_birth_year WHERE state='TX' AND birth_year=1968"))
         self.assertEqual(len(result), 1, "Expecting {} users, got {}".format(1, len(result)))
 
-    def add_dc_after_mv_test(self):
+    def __add_dc_after_mv_test(self, rf):
```
mambocab: Tiny Style Nit: a single leading underscore would be better here in case we want this parameterized test available on a subclass for test suite organization. The double leading underscore mangles the name of the method to protect it in subclasses, but I don't see a need to do that.

materialized_views_test.py
```diff
@@ -377,13 +384,37 @@ def add_node_after_mv_test(self):
...
+                         "WHERE v IS NOT NULL AND id IS NOT NULL PRIMARY KEY (v, id)"))
+
+        for i in xrange(1000):
+            session.execute("INSERT INTO t (id, v) VALUES ({id}, {v})".format(id=i, v=-i))
+
+        for i in xrange(1000):
+            assert_one(session, "SELECT * FROM t_by_v WHERE v = {}".format(-i), [-i, i])
+
+        node4 = new_node(self.cluster)
+        node4.start(wait_for_binary_proto=True, jvm_args=["-Dcassandra.write_survey=true"])
```
mambocab: I don't know anything about write survey mode, so, just to confirm: is it all right that we're adding a single node in that mode while the others aren't? If so, ignore me.

mambocab:
My tiny style nit is extremely tiny, feel free to ignore it if you don't care.

It's not clear from the name that `add_survey_node_after_mv_test` tests write survey mode. Should the test be renamed? Should there be a test for adding a node that isn't in write survey mode?

Could you please add `@jira_ticket` doxygen tags where applicable?

materialized_views_test.py
ptnapoleon: Yes, it is okay.

materialized_views_test.py
jkni: Good to know! I'll push a version with fix.

materialized_views_test.py
jkni: Just to doubly confirm: yes, it's fine. That's what we want here.

jkni:
I've pushed a commit with all the concerns addressed.

Thank you both for the close review; although my opinion here is far from the only one that matters, I like fixing those kinds of issues upfront.

mambocab:
So do we, thanks for addressing the feedback. I'm happy, Travis is happy, merging.

PR #648, opened 2015-11-06
Skip static columns paging test 10476

mambocab:
As per a comment from Sylvain on CASSANDRA-10476:

> We do expect 2.2.3 -> 3.0 upgrade to succeed. We don't expect
> static_columns_paging_test to pass however since that test don't even
> pass for non-upgrade test on 2.2.3, it's a bug of 2.2.3 that has been
> fixed in 2.2.4 (CASSANDRA-10381). If we can make so that the test
> doesn't run on 2.2.3 (but does run once we'll update to 2.2.4), that
> would be perfect.

@pcmanus is this what you had in mind? This will skip the entire test if any node is version 2.2.3 or lower. Another option is to still run the test, but only query nodes running higher versions. I'm not sure what configurations we expect errors in.

### DIFF
--- a/upgrade_tests/paging_test.py
+++ b/upgrade_tests/paging_test.py
@@ -1,6 +1,7 @@
 import itertools
 import time
 import uuid
+from unittest import SkipTest
 
 from cassandra import ConsistencyLevel as CL
 from cassandra import InvalidRequest, ReadFailure, ReadTimeout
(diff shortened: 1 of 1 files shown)


pcmanus:
Looks good on principle. I haven't tested it myself, but if you have, +1.

mambocab:
Great, thank you. Merging.

PR #650, opened 2015-11-06
Compaction throughput is an upper limit, don't fail tests if throughp…

krummas:
…ut is actually slower than the limit


ptnapoleon:
+1. For reference, explanation for removal of assert is in here: https://issues.apache.org/jira/browse/CASSANDRA-10667

PR #651, opened 2015-11-06
CASSANDRA-10645: Rework sstableverify tests to work cross-platform 

jkni:
This PR fixes CASSANDRA-10645.

It addresses a number of issues:
* move to cross-platform line-splitting
* fix filesystem casing issues in output across Python/Java
* more reliably corrupt sstable file so that the sstable isn't sometimes empty
* reduce number of stress writes to reduce test duration

The test now reliably passes on Windows 2.2/3.0 and Linux 2.2/3.0.

### DIFF
--- a/offline_tools_test.py
+++ b/offline_tools_test.py
@@ -1,4 +1,5 @@
 import os
+import random
 import re
 import subprocess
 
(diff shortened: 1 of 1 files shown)


offline_tools_test.py
```diff
@@ -192,16 +193,21 @@ def sstableverify_test(self):
...
         (out, error, rc) = node1.run_sstableverify("keyspace1", "standard1", output=True)
 
         self.assertEqual(rc, 0, msg=str(rc))
 
-        outlines = out.split("\n")
+        # Process sstableverify output to normalize paths in string to Python casing
+        outlines = map(lambda line: re.sub("(?<=path=').*(?=')",
+                                           lambda match: os.path.normcase(match.group(0)),
+                                           line),
+                       out.splitlines())
```
ptnapoleon: Discussed off-line, but this is clever, but a little hard to parse, so a longer comment would be nice.

ptnapoleon:
Other than commenting that mapping of lambdas a little better, I'm +1

PR #652, opened 2015-11-09
Tests for CASSANDRA-9302

stef1927:
Some extra tests required by CASSANDRA-9302.


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -954,8 +1053,34 @@ def _test_bulk_round_trip(self, nodes, partitioner, num_records):
...
+                                   profile=os.path.join(os.path.dirname(os.path.realpath(__file__)), 'blogposts.yaml'),
+                                   stress_table='stresscql.blogposts', page_timeout=60)
 
+    @require('9302')
     @freshCluster()
-    def test_bulk_round_trip(self):
-        self._test_bulk_round_trip(nodes=3, partitioner="murmur3", num_records=100000)
+    def test_bulk_round_trip_with_timeouts(self):
+        """
+        Test bulk import with very short read and write timeout values, this should exercise the
```
ptnapoleon: Can you add `@jira_ticket CASSANDRA-9302` here in this docstring and for all of the modified tests for 9302?

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -954,8 +1053,34 @@ def _test_bulk_round_trip(self, nodes, partitioner, num_records):
...
+        """
+        self._test_bulk_round_trip(nodes=3, partitioner="murmur3", num_operations=100000)
+
+    @require('9302')
+    @freshCluster()
+    def test_bulk_round_trip_blogposts(self):
+        """
+        Test bulk import with a user profile that inserts 10 rows per operation
+        """
+        self._test_bulk_round_trip(nodes=3, partitioner="murmur3", num_operations=10000,
```
ptnapoleon: Since stress takes options like `n=10K`, would it be possible to set `num_operations` here and elsewhere to use that format, instead of being integers? It is more readable.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -107,9 +117,77 @@ def all_datatypes_prepare(self):
...
             )''')
 
+        class Datetime(datetime.datetime):
+            def __str__(self):
+                return self.strftime(DEFAULT_TIME_FORMAT)
+
+            def __repr__(self):
+                return self.strftime(DEFAULT_TIME_FORMAT)
+
+        def str_q(s):
```
ptnapoleon: Could you comment this function or give it a more descriptive name, please?

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -107,9 +117,77 @@ def all_datatypes_prepare(self):
...
+            iteritems = frozenset.__iter__
+
+            def __repr__(self):
+                return '{{{}}}'.format(', '.join(['{}: {}'.format(str_q(t[0]), str_q(t[1])) for t in sorted(self)]))
+
+        class ImmutableSet(SortedSet):
+            def __repr__(self):
+                return '{{{}}}'.format(', '.join([str_q(t) for t in sorted(self._items)]))
+
+        class CommonEqualityMixin(object):
```
mambocab: You could probably use [`collections.namedtuple`](https://docs.python.org/2/library/collections.html#collections.namedtuple) instead of this and its subclasses. It's basically a class factory that provides nice `__repr__`s and equality functions similar to what you've implemented:

```python
>>> from collections import namedtuple
>>> Name = namedtuple('Name', ['firstname', 'lastname'])
>>> n0, n1, n2 = Name('Jim', 'Witschey'), Name('Jim', 'Witschey'), Name('Stefania', 'Alborghetti')
>>> n2
Name(firstname='Stefania', lastname='Alborghetti')
>>> n2._asdict()
OrderedDict([('firstname', 'Stefania'), ('lastname', 'Alborghetti')])
>>> n0 == n1
True
>>> n0 == n2
False
>>> 
```

mambocab: I had to do some digging to confirm for myself that this will work with `register_user_type`; I thought I'd share my findings. Here's an example of a `namedtuple` being used this way in the driver tests:

https://github.com/datastax/python-driver/blob/f22306972c3648072d0244432e1ba660f5514fe8/tests/integration/standard/test_udts.py#L76

stef1927:
I've addressed your comments in the latest commit, thanks. I've introduced named tuples for user types but we still need to keep a custom `__repr__` since the output format must match the csv format, including quotes.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -815,7 +905,7 @@ def test_wrong_number_of_columns(self):
                                         return_output=True)
 
         self.assertFalse(self.session.execute("SELECT * FROM testcolumns"))
-        self.assertIn('Aborting import', err)
+        self.assertTrue(len(err) > 0, 'Expected an error message')
```
mambocab: I should have asked earlier -- is is this change here to accommodate a behavior change in C*? If there's a change between C* versions, it'd be better to catch the behavior more precisely, even if we have to check different values for different versions.

mambocab:
I'm +1 on the changes now. @ptnapoleon?

By the way, when I'm +1 on your PRs, I've been letting them sit until you decide to merge them since you have permissions to do so. Is that what you want us to do, or would you rather have us merge your PRs once we review them and have no requested changes?

ptnapoleon:
I'm +1

mambocab:
Sorry, I wasn't clear -- if there's behavior we can check for more precisely in the code I commented on, the test should do that, then go ahead and merge. Thanks!

stef1927:
The error messages have changed a bit. I've added some more `@require` and re-introduced the checks on the error messages. Please see [latest commit] (https://github.com/stef1927/cassandra-dtest/commit/983f00033202b1005e3d74ee1ff8fb3b34cfa660). CASSANDRA-9302 goes to version 2.1, so we don't need any version checks.

Once you are OK with the changes, you can merge or you can just give me a clean +1, whichever you find easier! :)

mambocab:
Great, these look good to me. Thanks, merging.

PR #653, opened 2015-11-10
Add test for CASSANDRA-10390

beobal:


user_types_test.py
```diff
@@ -770,3 +770,25 @@ def test_user_type_isolation(self):
...
+        Ensure that a UDT with quoted field names can be accessed using the python driver
+        @jira_ticket CASSANDRA-10390
+        """
+
+        cluster = self.cluster
+        cluster.populate(1).start()
+        session = self.patient_cql_connection(cluster.nodelist()[0])
+        self.create_ks(session, 'ks', 1)
+
+        session.execute("""CREATE TYPE ks.test_type ("test.field" int)""")
```
mambocab: It's not worth messing with for this PR, but in the future, you can make string literals containing `"` by quoting them with `'` rather than `"""`.

mambocab:
The code is good, but why maintain this in the dtests? This looks like it's exclusively testing driver behavior.

user_types_test.py
beobal: Ok can do, but what about the case when the string literal contains both `"` and `'` characters? i.e. if `test.field` was of type `text`, the insert stmt would be:

`INSERT INTO ks.test_table (pk, test_field) VALUES (0, {"test.field": 'some text'})`

In that case, aren't you going to need `"""`, and if so, wouldn't it be more consistent to use that throughout for strings with escape-able chars?

mambocab: Using `'''` or `"""` is a good way to avoid escaping quotes when a string contains both `'` and `"`, yes. As for being consistent, I don't value that kind of consistency enough to make the common case use triple-quotes. It's a 4-character difference. 4 whole characters, Sam!

(I actually don't value quote consistency very highly in general for this codebase. I always use `'` because I'm a Python nerd, but preferring it over `"` isn't worth slowing down a code review or muddying up the Git history. I only mention it here because I find single-quoted strings more readable,  and because I don't know offhand if you've used Python enough to know about single-quoting with `'`.)

beobal: :) I didn't mean to suggest using triple quoting everywhere, think of all those wasted chars! 
Anyway, you've exposed my Pythonic ignorance here, I had no idea triple quoting with `'` was even a thing. For shame! I do generally use `'` where it's consistent with the surrounding code, but I ended up changing everything to `"` here (or so I thought, but now I notice I'd left a single-quoted string on line 783) because of the triple quoting. 
It wasn't my intention to argue about this point btw, the question about triple quoting where required was a genuine one.  I'm totally cool with following the conventions, and I'll continue to do that on future PRs.

beobal:
That's a good point, I guess I was too hasty in adding the test simply because it presented via cqlsh. A test was added to the driver along with the fix, so I've no problem at all about dropping this PR.

user_types_test.py
mambocab: Sounds good, thanks! Python quoting conventions are, more or less, "use `'` except where necessary to avoid escaping quotes in strings". Also, for whatever reason, `"""` is preferred over `'''` for docstrings, though I don't think one or the other is preferred for multiline strings in the code. In any event, this codebase has never been particularly consistent with respect to quotes. While that's ok, I appreciate any effort you put towards using Python conventions. Thanks!

mambocab:
Ok, cool -- I'll close since this primarily tests the driver. Thanks!

PR #655, opened 2015-11-11
ignore low diskspace warning

krummas:


mambocab:
Change looks fine, but I have a followup question: Is there a reason we're checking `stderr` here, rather than checking the return code, or even doing both?

PR #656, opened 2015-11-12
PEP8 Fixes

lmr:
I was taking a look at ccm's code with a checking tool I wrote (similar to pyflakes) that helped me prepare some PEP8, pylint and indentation fixes.

I did read the contributing guidelines and I hope to have followed them accordingly.

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -1819,8 +1819,9 @@ class CqlLoginTest(Tester):
             if expected_error in err_line:
                 break
         else:
-            self.fail("Did not find expected error '%s' in cqlsh stderr output: %s" %
-                      expected_error, '\n'.join(err_lines))
+            self.fail("Did not find expected error '{}' in "
+                      "cqlsh stderr output: {}".format(expected_error,
+                                                       '\n'.join(err_lines)))
 
     def test_login_allows_bad_pass_and_continued_use(self):
         self.create_ks(self.session, 'ks1', 1)
--- a/replication_test.py
+++ b/replication_test.py
@@ -288,25 +288,25 @@ class SnitchConfigurationUpdateTest(Tester):
         for i, node in enumerate(nodes):
             wait_expire = time.time() + 120
             while time.time() < wait_expire:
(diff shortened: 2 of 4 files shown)


largecolumn_test.py
```diff
@@ -4,6 +4,7 @@
 
 @since('2.2')
 class TestLargeColumn(Tester):
+
```
ptnapoleon: Checking the most up to date versions of PEP 257 and PEP8
https://www.python.org/dev/peps/pep-0257/
https://www.python.org/dev/peps/pep-0008/

I see no mention of the need to put a line between the class header and the docstring. Also, neither the pep8 linter nor pyflakes make any complaint here. Given that, we would prefer to keep the existing style.

delete_insert_test.py
```diff
@@ -9,6 +9,7 @@
 
 
 class DeleteInsertTest(Tester):
+
```
ptnapoleon: See comment on largecolumn_test

scrub_test.py
```diff
@@ -296,6 +298,7 @@ def test_scrub_collections_table(self):
 
 
 class TestScrub(TestHelper):
+
```
ptnapoleon: See comment on largecolumn_test.py

scrub_test.py
```diff
@@ -155,6 +156,7 @@ def increase_sstable_generations(self, sstables):
 
 @since('2.2')
 class TestScrubIndexes(TestHelper):
+
```
ptnapoleon: See comment on largecolumn_test.py

ptnapoleon:
+1, after requested changes are made. Thanks!

largecolumn_test.py
lmr: Sure. I usually follow that convention since `autopep8` suggests that change. I can prepare an updated version of this tomorrow, I think. Thanks!

ptnapoleon:
Bump. Could you make those updates so that I can merge? Thanks!

lmr:
@ptnapoleon I'm sorry, I've been absorbed in some other tasks, I'll send an updated version real quick.

ptnapoleon:
No problem at all. Thanks!

PR #657, opened 2015-11-12
jmx_test exception handling

mambocab:
2 commits here:

- one that does `s/"/'/` for Python nerd purposes, and
- one that cleans up the exception-handling logic in `jmx_test.py`.

The main reason for this PR is because I have a question about the existing logic here:

https://github.com/riptano/cassandra-dtest/compare/master...mambocab:jmx_test-exception-handling?expand=1#diff-a5036552850755a7ef3b1b262c2f2727L77

This catches the exception, checks to see if its message meets some criteron, then happily continues running the test. In the other call to `nodetool netstats`, it fails the test regardless of whether the exception meets our criterion or not:

https://github.com/riptano/cassandra-dtest/compare/master...mambocab:jmx_test-exception-handling?expand=1#diff-a5036552850755a7ef3b1b262c2f2727L88

Is both of these correct? The first seems wrong to me.

EDIT: Is both of these correct? Oh good.

### DIFF
--- a/jmx_test.py
+++ b/jmx_test.py
@@ -3,6 +3,9 @@ import sys
 import unittest
 import time
 
+import ccmlib.common
+from ccmlib.node import NodetoolError
+
 from dtest import Tester, debug
 from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
 from tools import since
@@ -68,24 +71,30 @@ class TestJMX(Tester):
         node1.stress(['write', 'n=5M', '-schema', 'replication(factor=3)'])
         node1.flush()
         node1.stop(gently=False)
-        try:
-            node1.nodetool('netstats')
-        except Exception as e:
-            if 'ConcurrentModificationException' in str(e):
-                self.fail('Netstats failed due to CASSANDRA-6577')
-            else:
-                debug(str(e))
 
-        node1.start(wait_for_binary_proto=True)
-
-        try:
+        with self.assertRaisesRegexp(NodetoolError, "ConnectException: 'Connection refused'."):
             node1.nodetool('netstats')
-        except Exception as e:
-            if 'java.lang.reflect.UndeclaredThrowableException' in str(e):
-                debug(str(e))
-                self.fail('Netstats failed with UndeclaredThrowableException (CASSANDRA-8122)')
-            else:
-                self.fail(str(e))
+
+        # don't wait; we're testing for when nodetool is called on a node mid-startup
+        node1.start(wait_for_binary_proto=False)
+
+        # until the binary interface is available, try `nodetool netstats`
+        binary_interface = node1.network_interfaces['binary']
+        time_out_at = time.time() + 30
+        running = False
+        while (not running and time.time() <= time_out_at):
+            running = ccmlib.common.check_socket_listening(binary_interface, timeout=0.5)
+            try:
+                node1.nodetool('netstats')
+            except Exception as e:
+                self.assertNotIn('java.lang.reflect.UndeclaredThrowableException', str(e),
+                                 'Netstats failed with UndeclaredThrowableException (CASSANDRA-8122)')
+                if not isinstance(e, NodetoolError):
+                    raise
+                else:
+                    self.assertIn("ConnectException: 'Connection refused'.", str(e))
+
+        self.assertTrue(running, msg='node1 never started')
 
     @since('2.1')
     def table_metric_mbeans_test(self):
--- a/jmx_test.py
+++ b/jmx_test.py
@@ -69,21 +69,21 @@ class TestJMX(Tester):
         node1.flush()
         node1.stop(gently=False)
         try:
-            node1.nodetool("netstats")
+            node1.nodetool('netstats')
         except Exception as e:
-            if "ConcurrentModificationException" in str(e):
-                self.fail("Netstats failed due to CASSANDRA-6577")
+            if 'ConcurrentModificationException' in str(e):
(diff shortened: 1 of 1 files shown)


mambocab:
Notes from a Hangout about this:

- In the first `except` block, the `assert in` is inaccurate -- CASSANDRA-6577 is about a node leaving during decommission, not about a node being stopped. Since we're calling `netstats` on a node that should be stopped, this error wouldn't be a regression on CASSANDRA-6577.
- The check for CASSANDRA-8122 doesn't actually test for that regression. The ticket concerns `nodetool` failing gracefully when it's called on a node that's still starting up. However, because of the `wait_for_binary_proto` in the previous `start` call, this test will never call `nodetool` on a test that's still starting up.

mambocab:
@ptnapoleon I believe I've addressed what we talked about. One thing I'm not sure of: I'm using `while not node1.is_running()` to see if the node is running properly. Now that I think about it harder, I believe that's just incorrect, we just care about it listening for the binary protocol. What's the best non-blocking way to check for that? Do I need to make another PR to ccm?

jmx_test.py
```diff
@@ -68,24 +70,19 @@ def netstats_test(self):
...
-                debug(str(e))
-                self.fail("Netstats failed with UndeclaredThrowableException (CASSANDRA-8122)")
-            else:
-                self.fail(str(e))
+        with self.assertRaisesRegexp(NodetoolError, "ConnectException: 'Connection refused'."):
+            node1.nodetool('netstats')
+
+        # don't wait; we're testing for when nodetool is called on a node mid-startup
+        node1.start(wait_for_binary_proto=False)
+        while not node1.is_running():
```
mambocab: What I want here is a non-blocking way to see if node1 is listening on the binary port. Current thinking, based on the implementation of `wait_for_binary_interface`: I want to do something like 

```python
mark = node1.mark_log()
while ((not node1.watch_log_for('Starting listening for CQL clients', from_mark=mark, timeout=0))
       and (not common.check_socket_listening(self.network_interfaces['binary'], timeout=0)):
```

This seems sort of unreasonably long, so I'm probably missing something.

ptnapoleon: Why not just grep the log for the "Starting listening" message?

mambocab: The `grep_log` method doesn't take a mark:

https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L301

ptnapoleon: You could refactor it to do so? I see no other methods that do what we want here, which feels silly.

jmx_test.py
```diff
@@ -68,24 +71,28 @@ def netstats_test(self):
...
-                self.fail(str(e))
+        with self.assertRaisesRegexp(NodetoolError, "ConnectException: 'Connection refused'."):
+            node1.nodetool('netstats')
+
+        # don't wait; we're testing for when nodetool is called on a node mid-startup
+        node1.start(wait_for_binary_proto=False)
+
+        # until the binary interface is available, try `nodetool netstats`
+        binary_interface = node1.network_interfaces['binary']
+        time_out_at = time.time() + 30
```
ptnapoleon: Bikeshed time, I think `timeout_at` is preferable.

mambocab:
I've pushed the changes we discussed. To continue the "wait, are we wasting our time" theme: does C* have to be running at all to test for the 8122 regression?

jmx_test.py
mambocab: As far as I'm concerned, "timeout" is a noun, "time out" is the verb-particle conjunction.

jmx_test.py
```diff
@@ -68,24 +71,28 @@ def netstats_test(self):
...
+        running = False
+        while (not running and time.time() <= time_out_at):
+            running = ccmlib.common.check_socket_listening(binary_interface, timeout=0.5)
+            try:
+                node1.nodetool('netstats')
+            except Exception as e:
+                self.assertNotIn('java.lang.reflect.UndeclaredThrowableException', str(e),
+                                 'Netstats failed with UndeclaredThrowableException (CASSANDRA-8122)')
+                if not isinstance(e, NodetoolError):
```
ptnapoleon: Are we sure this is right? Is this accidentally ever catching more than ConnectException?

jmx_test.py
mambocab: pfft, construction.

jmx_test.py
mambocab: It totally could. Fixing now.

jmx_test.py
ptnapoleon: A quick check of dictionaries shows that you are more correct than I. I yield.

ptnapoleon:
+1, but squash these commits first

mambocab:
I'll squash. Before merge, though, do you have any thoughts on my "does C* have to be running at all to test for the 8122 regression?" question? Maybe something to ask Carl/Marcus about.

ptnapoleon:
Sorry I missed that comment. I think so, maybe? In the ticket it was in the `STARTING` state when the exception occurred.

mambocab:
Fair enough. Does the squash look good to you?

ptnapoleon:
I don't see the point of keeping that second commit, given you entirely remove it immediately afterwards. But +1 either way

mambocab:
Re-squashed. I've also added some detail to the commit message to explain the changes; could you look those over please?

ptnapoleon:
+1

mambocab:
Great, thank you.

PR #658, opened 2015-11-12
Remove all pre 2.1 logic

ptnapoleon:
So I will clearly set up and run a CI job with my changes, but looking for review from @mambocab and @EnigmaCurry.

### DIFF
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -90,8 +90,7 @@ class TestCommitLog(Tester):
         conf = {'commitlog_segment_size_in_mb': segment_size_in_mb}
         if compressed:
             conf['commitlog_compression'] = [{'class_name': 'LZ4Compressor'}]
-        if self.cluster.version() >= "2.1":
-            conf['memtable_heap_space_in_mb'] = 512
+        conf['memtable_heap_space_in_mb'] = 512
         self.prepare(configuration=conf, create_test_keyspace=False)
 
         segment_size = segment_size_in_mb * 1024 * 1024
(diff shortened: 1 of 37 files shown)


EnigmaCurry:
This is awesome.

ptnapoleon:
For those unaware, as 2.0 and earlier are EOL, and we aren't testing them anymore, there's no need to clutter the codebase with any pre-2.1 logic.

knifewine:
Nice.

mambocab:
I've only given this a quick skim so far, but very exciting stuff. Looks like there are a bunch of `since('2.1')`s hanging around too. Any reason to keep them around?

ptnapoleon:
Nope, I'll get those too

commitlog_test.py
```diff
@@ -478,11 +471,7 @@ def get_header_crc(header):
             # read the header and find the crc location
             with open(os.path.join(cl_dir, cl), 'r') as f:
                 f.seek(0)
-                version = struct.unpack('>i', f.read(4))[0]
                 crc_pos = 12
-                # if version < 5:
```
mambocab: If this is removed, we may also be able to take out this:

https://github.com/riptano/cassandra-dtest/blob/two-oh-cleanup/commitlog_test.py#L388

Maybe check with Blake, I don't know what the logic is.

mambocab: btw this isn't a blocker for merging this, just something to clean up if we get the chance/it's easy.

mambocab:
Review by visual inspection looks good so far (as of 5e7ddb9).

ptnapoleon:
The run with 5e7ddb9 succeeded on cassci. Waiting on the removal of the since code now.

ptnapoleon:
Okay the results look great: http://cassci.datastax.com/job/ptnapoleon-2.0-cleanup-dtest/

Will squash/merge after a +1

mambocab:
+1 post-squash as well.

ptnapoleon:
\o/. Thanks

PR #660, opened 2015-11-13
don't use bare except trying to strip timeszones

mambocab:
I believe `ValueError`s and `TypeError`s are the only ones thrown when the functions called here are provided incorrect arguments. I ran the cqlsh_copy_tests and nothing blew up, which is promising. (In any event, we need to not use a bare `except:` here, so if this is an incomplete list of exceptions to check for, it'll still be a starting point for doing this correctly.)


ptnapoleon:
+1

mambocab:
Thanks, merging.

PR #662, opened 2015-11-16
added test for cassandra-10711; empty batch causing exception

nutbunnies:


batch_test.py
```diff
@@ -7,11 +7,24 @@
...
+        """
+        @jira_ticket CASSANDRA-10711
+        """
+        session = self.prepare()
+        session.execute("""
+            BEGIN BATCH
+            APPLY BATCH;
+        """)
+        for node in self.cluster.nodelist():
+            self.assertEquals(0, len(node.grep_log_for_errors()))
```
ptnapoleon: This happens automatically in node teardown, and thus is redundant, no?

nutbunnies: True, but I have an aversion to a test without any assertions.  Could be replaced with a comment I suppose.

mambocab: I prefer the assertion to a comment, personally.

mambocab:
I'm +1.

PR #663, opened 2015-11-17
Increase connect timeout

stef1927:
We are getting so many timeouts when connecting, see for example [this job here](http://cassci.datastax.com/job/cassandra-3.1_dtest/lastCompletedBuild/testReport) .

I would like to see if increasing the connect_timeout helps.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -397,7 +397,7 @@ class Tester(TestCase):
 
         cluster = PyCluster([node_ip], auth_provider=auth_provider, compression=compression,
                             protocol_version=protocol_version, load_balancing_policy=load_balancing_policy, default_retry_policy=FlakyRetryPolicy(),
-                            port=port, ssl_options=ssl_opts)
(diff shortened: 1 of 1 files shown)


mambocab:
@ptnapoleon You had more experience with the bad parts of getting the tests working on Windows. I think this change is fine, including changing the multiplier on Windows. As I read it, this change

- bumps the driver timeouts from 5 to 10 seconds
- bumps the retry logic's timeout from 10 to 30 seconds
- bumps the Windows retry timeout from 50 to 60 seconds.

so, @ptnapoleon, do you think I read that right, and if so, do you agree this is a reasonable change?

ptnapoleon:
Thirty seconds for the retry logic seems excessive, but I suppose now that the driver timeout is 10, that's only 3 retries, which seems reasonable. I'm +1, it's worth seeing if this helps.

mambocab:
Thanks, Stefania and Philip. Merging.

PR #664, opened 2015-11-17
check that we dont anticompact after certain repair modes

krummas:


repair_test.py
```diff
@@ -39,6 +39,60 @@ def check_rows_on_node(self, node_to_check, rows, found=None, missings=None, res
             for node in stopped_nodes:
                 node.start(wait_other_notice=True)
 
+    @since('2.2.1')
+    def no_anticompaction_after_dclocal_repair_test(self):
+        cluster = self.cluster
+        debug("Starting cluster..")
+        cluster.populate([2, 2]).start()
```
mambocab: Is there any reason not to use `start(wait_for_binary_proto=True)` here? I think I'd prefer that.

krummas: no reason, fixed here: https://github.com/riptano/cassandra-dtest/pull/668 - thanks for noticing

PR #665, opened 2015-11-17
Fix  strip common prefix

mambocab:
Fixes the problem described [in this Jira comment from Joel](https://issues.apache.org/jira/browse/CASSANDRA-10632?focusedCommentId=15007942&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15007942) and adds some assertions that the prefix-stripping works correctly.

I used the file in [this Gist](https://gist.github.com/865f97bde4534bd3f89c) to smoke test the changed function:

```python
In [1]: from sstableutil_test import _strip_common_prefix ; from filenames import filenames                                                                                   

In [2]: _strip_common_prefix(filenames)
Out[2]: 
['1-big-CRC.db',
 '1-big-Data.db',
 '1-big-Digest.crc32',
 '1-big-Filter.db',
 '1-big-Index.db',
 '1-big-Statistics.db',
 '1-big-Summary.db',
 '1-big-TOC.txt',
 '2-big-CRC.db',
 '2-big-Data.db',
 '2-big-Digest.crc32',
 '2-big-Filter.db',
 '2-big-Index.db',
 '2-big-Statistics.db',
 '2-big-Summary.db',
 '2-big-TOC.txt',
 '3-big-CRC.db',
 '3-big-Data.db',
 '3-big-Digest.crc32',
 '3-big-Filter.db',
 '3-big-Index.db',
 '3-big-Statistics.db',
 '3-big-Summary.db',
 '3-big-TOC.txt',
 '4-big-CRC.db',
 '4-big-Data.db',
 '4-big-Digest.crc32',
 '4-big-Filter.db',
 '4-big-Index.db',
 '4-big-Statistics.db',
 '4-big-Summary.db',
 '4-big-TOC.txt',
 '5-big-CRC.db',
 '5-big-Data.db',
 '5-big-Digest.crc32',
 '5-big-Filter.db',
 '5-big-Index.db',
 '5-big-Statistics.db',
 '5-big-Summary.db',
 '5-big-TOC.txt']

In [3]: len(filenames), len(_strip_common_prefix(filenames))
Out[3]: (40, 40)
```

This looks correct to me. @jkni can you please review?

### DIFF
--- a/sstableutil_test.py
+++ b/sstableutil_test.py
@@ -12,18 +12,19 @@ KeyspaceName = 'keyspace1'
 TableName = 'standard1'
 
 
+def _remove_prefix(string, prefix):
+    assert string.startswith(prefix), '{p} not a prefix of {s}'.format(p=prefix, s=string)
+    suffix = string[len(prefix):]
+    assert string.endswith(suffix), '{u} not a prefix of {t}'.format(u=suffix, t=string)
+    assert len(suffix) + len(prefix) == len(string)
+    return suffix
+
+
 def _strip_common_prefix(strings):
     strings = list(map(os.path.normcase, strings))
     common_prefix = os.path.commonprefix(strings)
 
-    rvs = []
-    for s in strings:
-        stripped = s[len(common_prefix):]
-        assert len(stripped) + len(common_prefix) == len(s)
-        assert s.endswith(stripped)
-        rvs.append(stripped)
-
-    return rvs
+    return [_remove_prefix(string, common_prefix) for string in strings]
 
 
 @since('3.0')
--- a/sstableutil_test.py
+++ b/sstableutil_test.py
@@ -145,6 +145,8 @@ class SSTableUtilTest(Tester):
 
         if len(expected_tmpfiles) == 0:
             expected_tmpfiles = sorted(list(set(allfiles) - set(finalfiles)))
+        if common.is_win():
+            expected_tmpfiles = _strip_common_prefix(expected_tmpfiles)
 
         debug("Comparing tmp files...")
         tmpfiles = self._invoke_sstableutil(ks, table, type='tmp')
--- a/sstableutil_test.py
+++ b/sstableutil_test.py
@@ -143,14 +143,13 @@ class SSTableUtilTest(Tester):
 
         self.assertEqual(expected_finalfiles, finalfiles)
 
-        if len(expected_tmpfiles) == 0:
-            expected_tmpfiles = sorted(list(set(allfiles) - set(finalfiles)))
-        if common.is_win():
-            expected_tmpfiles = _strip_common_prefix(expected_tmpfiles)
-
         debug("Comparing tmp files...")
         tmpfiles = self._invoke_sstableutil(ks, table, type='tmp')
-        self.assertEqual(expected_tmpfiles, tmpfiles)
+
+        common_prefix = os.path.commonprefix(list(tmpfiles) + list(expected_tmpfiles))
+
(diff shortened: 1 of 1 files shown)


mambocab:
I should note that I don't have a Windows environment up and running, so I haven't run the test suite with these changes on Windows.

sstableutil_test.py
```diff
@@ -143,6 +145,8 @@ def _check_files(self, node, ks, table, expected_finalfiles=[], expected_tmpfile
 
         if len(expected_tmpfiles) == 0:
             expected_tmpfiles = sorted(list(set(allfiles) - set(finalfiles)))
+        if common.is_win():
+            expected_tmpfiles = _strip_common_prefix(expected_tmpfiles)
```
jkni: Expected_tmpfiles doesn't need the common prefix stripped since allfiles and finalfiles have already been stripped.

It is the tmpfiles two lines down that needs the common prefix stripped; that said, _strip_common_prefix won't work here since tmpfiles will strip more than just the path up to directory since there are likely only tmpfiles for one sstable.

mambocab: Ah, I see. I've pushed a possible fix.

I don't know that I understand what the correct behavior is supposed to be -- is it even possible to pass in a value of `expected_tmpfiles` that you can know will be correct beforehand? In any event, I don't see any `_check_files` calls that actually use the `expected_tmpfiles` arg, so this may be moot, as `expected_tmpfiles` only ever be `[]` as the function is currently called.

jkni:
The last commit isn't exactly what I meant.

We still want to set expected_tmpfiles to the intersection of allfiles and finalfiles if not provided.

If stripping prefixes, we want to calculate common_prefix once for allfiles, then use that to strip the prefix for other paths.  Once that is fixes, the assert on line 90 will be broken since the results of invoking sstableutil won't be stripped to match finalfiles.  After that, we want to be able to pass in empty lists for expected_finalfiles and expected_tmpfiles, so we instead check for these args to be None.  I then observed that stripping files seems to be incidental in the original path fixing and is no longer necessary, so I removed it.

With these changes, the test passes consistencly on Windows. I haven't checked on Linux.  Pushed [here](https://github.com/jkni/cassandra-dtest/tree/sstableutil) on my fork.

mambocab:
The prefix-stripping was necessary for my OpenStack Windows environment I was testing in, which shortened the names of files as reported by some tools as described here:

http://superuser.com/q/348079/400161

I'll see what I can do about disabling that next time I have a Windows env set up. I assumed this was just something I'd have to deal with in that environment, but that could be wrong.

I've skimmed your changes and they look good; can you make a PR for a proper review? I'll close this PR.

mambocab:
Actually, I have a few small python 3 compatability changes to make; I'll make the PR.

jkni:
Sounds great - thanks.

On Wed, Nov 18, 2015 at 9:27 AM, Jim Witschey <notifications@github.com>
wrote:

> Actually, I have a few small python 3 compatability changes to make; I'll
> make the PR.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/riptano/cassandra-dtest/pull/665#issuecomment-157748767>
> .
>



-- 

<http://www.datastax.com/>

Joel Knighton
Cassandra Developer | joel.knighton@datastax.com

<https://www.linkedin.com/company/datastax>
<https://www.facebook.com/datastax> <https://twitter.com/datastax>
<https://plus.google.com/+Datastax/about>
<http://feeds.feedburner.com/datastax> <https://github.com/datastax/>

PR #670, opened 2015-11-18
CASSANDRA-9043: Add test case for COPY from COUNTER csv

jasonstack:
CASSANDRA-9043: Add test case for COPY from COUNTER csv

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -468,6 +468,41 @@ class CqlshCopyTest(Tester):
         self.assertItemsEqual(csv_values,
                               [['a', 'b'], ['1', '10'], ['2', '20'], ['3', '30']])
 
+    def test_reading_counter(self):
+        """
+        Test that COPY can read a CSV of COUNTER by:
+
+        - creating a table,
+        - writing a CSV with COUNTER data with header,
+        - importing the contents of the CSV file using COPY with header,
+        - checking that the contents of the table are the written values.
+        @jira_ticket CASSANDRA-9043
+        """
+        self.prepare()
+        self.session.execute("""
+            CREATE TABLE testcounter (
+                a int primary key,
+                b counter
+            )""")
+
+        self.tempfile = NamedTemporaryFile(delete=False)
+
+        data = [[1, 20], [2, 40], [3, 60], [4, 80]]
+
+        with open(self.tempfile.name, 'w') as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=['a', 'b'])
+            writer.writeheader()
+            for a, b in data:
+                writer.writerow({'a': a, 'b': b})
+            csvfile.close
+
+        cmds = "COPY ks.testcounter FROM '{name}'".format(name=self.tempfile.name)
+        cmds += " WITH HEADER = true"
+        self.node1.run_cqlsh(cmds=cmds)
+
+        result = self.session.execute("SELECT * FROM testcounter")
+        self.assertItemsEqual(data, rows_to_list(result))
+
     def test_reading_use_header(self):
         """
         Test that COPY can read a CSV with a header by:


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -468,6 +468,41 @@ def test_writing_use_header(self):
...
+        data = [[1, 20], [2, 40], [3, 60], [4, 80]]
+
+        with open(self.tempfile.name, 'w') as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=['a', 'b'])
+            writer.writeheader()
+            for a, b in data:
+                writer.writerow({'a': a, 'b': b})
+            csvfile.close
+
+        cmds = "COPY ks.testheader FROM '{name}'".format(name=self.tempfile.name)
```
mambocab: This isn't the same table name from the other operations.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -468,6 +468,41 @@ def test_writing_use_header(self):
...
+        cmds += " WITH HEADER = true"
+        self.node1.run_cqlsh(cmds=cmds)
+
+        result = self.session.execute("SELECT * FROM testcounter")
+        self.assertItemsEqual([tuple(d) for d in data],
+                              [tuple(r) for r in rows_to_list(result)])
```
mambocab: Converting to tuples here shouldn't be necessary.

mambocab: Actually, the list comprehensions shouldn't be necessary either; `assertItemsEqual(data, rows_to_list(result))` should be fine here.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -468,6 +468,41 @@ def test_writing_use_header(self):
...
+            for a, b in data:
+                writer.writerow({'a': a, 'b': b})
+            csvfile.close
+
+        cmds = "COPY ks.testheader FROM '{name}'".format(name=self.tempfile.name)
+        cmds += " WITH HEADER = true"
+        self.node1.run_cqlsh(cmds=cmds)
+
+        result = self.session.execute("SELECT * FROM testcounter")
+        self.assertItemsEqual([tuple(d) for d in data],
```
ptnapoleon: `self.assertItemsEqual(data, rows_to_list(result))` also works, so why cast to tuple here?

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -468,6 +468,41 @@ def test_writing_use_header(self):
...
 
+    def test_reading_counter(self):
+        """
+        Test that COPY can read a CSV of COUNTER by:
+
+        - creating a table,
+        - writing a CSV with COUNTER data with header,
+        - importing the contents of the CSV file using COPY with header,
+        - checking that the contents of the table are the written values.
+        """
```
mambocab: Could you please add this doxygen tag to the docstring here: `@jira_ticket CASSANDRA-9043`

Here's another example of it in use:

https://github.com/riptano/cassandra-dtest/blob/7e7b09d977eb1e113fd4e26fb96bdba3a250a03a/offline_tools_test.py#L23

mambocab:
Thank you for the test! Patched trunk passes, and existing branches fail (after fixing the table name error in the test itself).

Once you address my comments I'll be +1. Do you anticipate the patch being merged soon? If so, I'd rather just hold off merging these test changes until that happens.

cqlsh_tests/cqlsh_copy_tests.py
jasonstack: thank you. fixed.

cqlsh_tests/cqlsh_copy_tests.py
jasonstack: thank you. fixed.

cqlsh_tests/cqlsh_copy_tests.py
jasonstack: thank you. fixed.

cqlsh_tests/cqlsh_copy_tests.py
jasonstack: thank you. fixed.

jasonstack:
Hi all, I have updated based on the comments. and CASSANDRA-9043 has been merged. Can you check this PR again? thank you

ptnapoleon:
Merged, thanks!

PR #671, opened 2015-11-18
Cleanup repair test

ptnapoleon:
@mambocab to review

### DIFF
--- a/repair_test.py
+++ b/repair_test.py
@@ -41,11 +41,14 @@ class TestRepair(Tester):
 
     @since('2.2.1')
     def no_anticompaction_after_dclocal_repair_test(self):
+        """
+        @jira_ticket CASSANDRA-10422
+        """
         cluster = self.cluster
         debug("Starting cluster..")
         cluster.populate([2, 2]).start(wait_for_binary_proto=True)
-        [node1_1, node2_1, node1_2, node2_2] = cluster.nodelist()
-        node1_1.stress(stress_options=['write', 'n=50000', 'cl=ONE', '-schema', 'replication(factor=2)'])
+        node1_1, node2_1, node1_2, node2_2 = cluster.nodelist()
+        node1_1.stress(stress_options=['write', 'n=50K', 'cl=ONE', '-schema', 'replication(factor=2)'])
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Tests all pass on 3.0, after these non-logic changes

repair_test.py
```diff
@@ -58,12 +61,15 @@ def no_anticompaction_after_dclocal_repair_test(self):
...
+        """
         cluster = self.cluster
         debug("Starting cluster..")
         cluster.populate([2, 2]).start(wait_for_binary_proto=True)
-        [node1_1, node2_1, node1_2, node2_2] = cluster.nodelist()
-        node1_1.stress(stress_options=['write', 'n=50000', 'cl=ONE', '-schema', 'replication(factor=2)'])
-        node1_1.nodetool("repair -hosts 127.0.0.1,127.0.0.2,127.0.0.3,127.0.0.4 keyspace1 standard1")
+        node1_1, node2_1, node1_2, node2_2 = cluster.nodelist()
+        node1_1.stress(stress_options=['write', 'n=50K', 'cl=ONE', '-schema', 'replication(factor=2)'])
+        node1_1.nodetool("repair -hosts 127.0.0.1,127.0.0.2,12  7.0.0.3,127.0.0.4 keyspace1 standard1")
```
mambocab: Intentional? and if the tests pass after this change, what is going on?

mambocab:
+1 other than my comment. Maybe we need assertions to make sure that repair actually happens.

repair_test.py
ptnapoleon: Definitely not intentional. Also, scary that test passes

ptnapoleon:
I assume the 
`self.assertTrue(node.grep_log("Not a global repair"))` makes sure the repair happens

mambocab:
The merge is fine, but I'm still concerned. Why didn't nodetool repair fail on what I assume are malformed arguments? Maybe that assertion doesn't do what we think it does.

ptnapoleon:
After rerunning tests, it does fail with that mistake. So I must have accidentally included that space after checking the tests, but before committing.

PR #673, opened 2015-11-18
Fix  strip common prefix

mambocab:
Consider this a reopened version of #665.

I've applied @jkni's (excellent) changes and made some changes, mostly for Python 3 compatibility (`map` returns a list in py2, but a generator in py3, so we have to flatten it to a list. I pulled out a method to do just that.).

@jkni can you please review and smoke test on Windows when you get a chance? and @ptnapoleon can you review for readability? I think Joel's changes make this much more readable but I already know the test, so.

### DIFF
--- a/sstableutil_test.py
+++ b/sstableutil_test.py
@@ -11,22 +11,6 @@ import os
 KeyspaceName = 'keyspace1'
 TableName = 'standard1'
 
-
-def _remove_prefix(string, prefix):
-    assert string.startswith(prefix), '{p} not a prefix of {s}'.format(p=prefix, s=string)
-    suffix = string[len(prefix):]
-    assert string.endswith(suffix), '{u} not a prefix of {t}'.format(u=suffix, t=string)
-    assert len(suffix) + len(prefix) == len(string)
-    return suffix
-
-
-def _strip_common_prefix(strings):
-    strings = list(map(os.path.normcase, strings))
-    common_prefix = os.path.commonprefix(strings)
-
-    return [_remove_prefix(string, common_prefix) for string in strings]
-
-
 @since('3.0')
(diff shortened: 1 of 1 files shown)


sstableutil_test.py
```diff
@@ -110,48 +104,36 @@ def _create_data(self, node, ks, table, numrecords):
...
-                expected_finalfiles = _strip_common_prefix(expected_finalfiles)
-            if expected_tmpfiles:
-                expected_tmpfiles = _strip_common_prefix(expected_tmpfiles)
+    def _check_files(self, node, ks, table, expected_finalfiles=None, expected_tmpfiles=None):
+        sstablefiles = _normcase_all(self._get_sstable_files(node, ks, table))
+        allfiles = _normcase_all(self._invoke_sstableutil(ks, table, type='all'))
+        finalfiles = _normcase_all(self._invoke_sstableutil(ks, table, type='final'))
+        tmpfiles = _normcase_all(self._invoke_sstableutil(ks, table, type='tmp'))
+        expected_oplogs = _normcase_all(self._get_sstable_transaction_logs(node, ks, table))
+        tmpfiles_with_oplogs = _normcase_all(self._invoke_sstableutil(ks, table, type='tmp', oplogs=True))
```
ptnapoleon: This is very dense, but I see no way to meaningfully improve it.

ptnapoleon:
Sufficiently readable. I wouldn't complain about comments on those variables explaining what they are, but the names are pretty descriptive. I'll leave functional review to Joel.

jkni:
Both abortedcompaction_test and compaction_test pass on my Windows environment now. +1.

mambocab:
\o/ Thanks, all. Merging.

PR #676, opened 2015-11-19
fixes ssl store file deletion

mambocab:
As previously written, if one of the file deletions failed, the subsequent deletions wouldn't even be attempted.

I've also added a note about a change to make after the move to Python 3.


ptnapoleon:
+1

mambocab:
Thanks, merged.

PR #677, opened 2015-11-19
Unify corrupt sstable output analysis in sstableverify test

jkni:
In [CASSANDRA-10582](https://issues.apache.org/jira/browse/CASSANDRA-10582), the exception thrown on corrupt sstables was unified with that of 3.0.

The test wasn't updated, so it was failing on 2.2 when it looked for the old exception output.

I've removed the 2.2 branch in an if condition; we no longer need to handle it separately.


ptnapoleon:
+1

PR #679, opened 2015-11-21
adjust schema metadata checks

knifewine:
PYTHON-404 seems to have changed this, so tests need to adjust.

Good benefit is no more version checking.


ptnapoleon:
Could you make Travis happy please?

ptnapoleon:
And if you're sure the new logic is correct, I'm +1 after the travis fixes

PR #680, opened 2015-11-23
fix sstablelevelreset test

krummas:


ptnapoleon:
If you're comfortable that the new logic is correct, I'm +1 on the code changes, though there is no ticket mentioned here, so I don't know if the new behavior is correct or not.

krummas:
just a quick hack to make it pass - makes it ignore the small file-warning (and yes, I think it is safe to ignore that)

ptnapoleon:
This test is still failing though, http://cassci.datastax.com/job/cassandra-3.0_dtest/374/testReport/offline_tools_test/TestOfflineTools/sstablelevelreset_test/

PR #681, opened 2015-11-23
Reduce stress load in netstats_test

ptnapoleon:
There is no need to load this much data to effectively test what we're testing. It just makes the test take too long. I'd like to drop it even farther if I could, to 50 or 100k. I also used the cleaner shorthand to notate the load on another test in the file.

@mambocab

### DIFF
--- a/jmx_test.py
+++ b/jmx_test.py
@@ -65,7 +65,7 @@ class TestJMX(Tester):
         cluster.populate(3).start(wait_for_binary_proto=True)
         node1, node2, node3 = cluster.nodelist()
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
It looks like this is actually causing the test to fail. Having more data is making it take longer to replay commitlogs on startup. So we need to choose a value that let's it take long enough to startup to properly test what we want, while also not taking too long. 1M?

ptnapoleon:
For the record I'm still fine with 500K, as we can be sure it takes at least 3 seconds on cassci, which is 5-6 iterations of that loop

mambocab:
Yep, looks good. +1.

PR #684, opened 2015-11-24
ignore disk space warning in sstablelevelreset_test

krummas:


offline_tools_test.py
```diff
@@ -39,7 +39,8 @@ def sstablelevelreset_test(self):
         cluster.stop(gently=False)
 
         (output, error, rc) = node1.run_sstablelevelreset("keyspace1", "standard1", output=True)
-        self.assertTrue(len(error) == 0 or "Max sstable size of" in error, error)
+        if len(error) > 0:
+            self.assertTrue("Max sstable size of" in error or "Consider adding more capacity" in error, error)
```
mambocab: This check will lead to false positives if `stderr` has both an acceptable error and an unacceptable one. Would something like this work?

```python
if len(error) > 0:
    for line in error.splitlines():
        self.assertTrue("Max sstable size of" in error or "Consider adding more capacity" in line,
                        'Found line \n\n"{line}"\n\n in error\n\n{error}'.format(line=line, error=error))

**Comment by mambocab** [comment]
Could you please fix the check to make sure we don't pass on unacceptable errors, then ping me for another review? Thanks.

**Comment by ptnapoleon** [comment]
Bump. I'd like to resolve this so that we can make these tests stop failing

**Comment by krummas** [comment]
oops, missed this, will fix tomorrow

**Comment by krummas** [comment]
force pushed a fix

File: offline_tools_test.py [hunk 2]
```diff
@@ -264,6 +263,12 @@ def sstableexpiredblockers_test(self):
         [(out, error, rc)] = node1.run_sstableexpiredblockers(keyspace="ks", column_family="cf")
         self.assertIn("blocks 2 expired sstables from getting dropped", out)
 
+    def _check_stderr_error(self, error):
+        if len(error) > 0:
+            for line in error.splitlines():
+                self.assertTrue("Max sstable size of" in error or "Consider adding more capacity" in line,
```
mambocab: Shouldn't this be `"Max sstable size of" in line`, not `"Max sstable size of" in error`?

mambocab: Ah, looking at my comment on my last review, I made this mistake too; sorry about that.

krummas:
nice catch, fixed

mambocab:
Looks great, thank you. Merging.

PR #685, opened 2015-11-25
Make sure that the index is built before using it (CASSANDRA-8505)

blerer:
CASSANDRA-8505 is changing the behaviour when an index is building. If an index is not fully built, it will throw an error when it is queried. The patch makes the test wait until the index is built.


ptnapoleon:
+1. We/You can merge once 8505 is committed. Since 8505 is for 2.2 and 3.0 only, will this function break running on C* versions prior to 8505's commit?

blerer:
No, it should work fine with or without 8505.

ptnapoleon:
Do we want to merge now then?

PR #687, opened 2015-11-26
Add GossipingPropertyFileSnitch test for CASSANDRA-9748

pauloricardomg:


snitch_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # for i, node in enumerate(cluster.nodelist()):
+        out, err = node.nodetool('gossipinfo')
+        self.assertEqual(0, len(err), err)
+        debug(out)
+
+        assert "/127.0.0.3" in out
+        assert "INTERNAL_IP:6:127.0.0.1" in out
+        assert "/127.0.0.4" in out
+        assert "INTERNAL_IP:6:127.0.0.2" in out
```
ptnapoleon: `self.assertIn()` and `self.assertNotIn()` both exist and are preferable here.

snitch_test.py
```diff
@@ -0,0 +1,69 @@
...
+                snitch_file.write("dc=dc1" + os.linesep)
+                snitch_file.write("rack=rack1" + os.linesep)
+                snitch_file.write("prefer_local=true" + os.linesep)
+
+        node1.start(wait_for_binary_proto=True)
+        node1.watch_log_for("Starting Messaging Service on /127.0.0.1:7000", timeout=60)
+        node1.watch_log_for("Starting Messaging Service on /127.0.0.3:7000", timeout=60)
+
+        # write some data to node1
+        node1.stress(['write', 'n=10000', '-rate', 'threads=8'])
```
ptnapoleon: Could you make this use `n=10K` instead, please?

snitch_test.py
```diff
@@ -0,0 +1,69 @@
...
+        @jira_ticket CASSANDRA-8084
+
+        Test that it's possible to connect over the broadcast_address when
+        listen_on_broadcast_address=true and that GossipingPropertyFileSnitch
+        reconnect via listen_address when prefer_local=true
+        """
+
+        cluster = self.cluster
+        cluster.populate(2)
+        [node1, node2] = cluster.nodelist()
```
ptnapoleon: This can just be 
```
node1, node2 = cluster.nodelist()
```

snitch_test.py
```diff
@@ -0,0 +1,69 @@
...
+        node1.start(wait_for_binary_proto=True)
+        node1.watch_log_for("Starting Messaging Service on /127.0.0.1:7000", timeout=60)
+        node1.watch_log_for("Starting Messaging Service on /127.0.0.3:7000", timeout=60)
+
+        # write some data to node1
+        node1.stress(['write', 'n=10000', '-rate', 'threads=8'])
+
+        session = self.patient_cql_connection(node1)
+        stress_table = 'keyspace1.standard1'
+        original_rows = list(session.execute("SELECT * FROM %s" % (stress_table,)))
```
ptnapoleon: Could you make the execute call look like `execute("SELECT * FROM {}".format(stress_table))`

snitch_test.py
```diff
@@ -0,0 +1,69 @@
...
+                                filename='debug.log', timeout=60)
+            node2.watch_log_for("Intiated reconnect to an Internal IP /127.0.0.1 for the /127.0.0.3",
+                                filename='debug.log', timeout=60)
+
+        # read data from node2 just to make sure data and connectivity is OK
+        session = self.patient_exclusive_cql_connection(node2)
+        new_rows = list(session.execute("SELECT * FROM %s" % (stress_table,)))
+        self.assertEquals(original_rows, new_rows)
+
+        # for i, node in enumerate(cluster.nodelist()):
```
ptnapoleon: Is this for loop supposed to be commented out?

snitch_test.py
```diff
@@ -0,0 +1,69 @@
...
+        out, err = node.nodetool('gossipinfo')
```
ptnapoleon: Is `node` defined here?

ptnapoleon:
It looks like this should be restricted to running on 2.2+. 
```
out, err = node1.nodetool('gossipinfo')
        self.assertEqual(0, len(err), err)
```
is problematic, as this will guarantee the test fails on OSX, which always prints to stderr when any java command is run.

ptnapoleon:
Ticket was committed. Merging this.

PR #605, opened 2015-10-14
Testing rack swapping

carlyeks:
For CASSANDRA-10242

PR #688, opened 2015-11-30
Fixes to replication_test.py for CASSANDRA-10243

stef1927:
This fixes the replication tests due to the changes for CASSANDRA-10243 and introduces new tests for it. One new test, `test_cannot_restart_with_different_rack` will still fail until CASSANDRA-9474 is also committed (which is imminent so I did not `@require` the test).

### DIFF
--- a/replication_test.py
+++ b/replication_test.py
@@ -257,20 +257,26 @@ class ReplicationTest(Tester):
         self.assertEqual(len(forwarders_used), 3)
 
 
+@require('10243')
+@require('9474')
 class SnitchConfigurationUpdateTest(Tester):
     """
     Test to reproduce CASSANDRA-10238, wherein changing snitch properties to change racks without a restart
     could violate RF contract.
 
     Since CASSANDRA-10243 it is no longer possible to change rack or dc for live nodes so we must specify
-    which nodes should be shutdown in order to have the rack changed. Since CASSANDRA-10242 it is no longer
-    possible to start a node with a different rack unless we specify -Dcassandra.ignore_rack=true.
+    which nodes should be shutdown in order to have the rack changed. Since CASSANDRA-10242 and CASSANDRA-9474
+    it is no longer possible to start a node with a different rack unless we specify -Dcassandra.ignore_rack
+    (CASSANDRA-10242 only in 2.1) or -Dcassandra.override_rackdc (CASSANDRA-9474, 2.1+). These tests already use
+    the latest property name (override_rackdc) and therefore require CASSANDRA-9474 as well.
     """
 
     def __init__(self, *args, **kwargs):
         Tester.__init__(self, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I'm +1, @mambocab ?

replication_test.py
```diff
@@ -259,8 +259,19 @@ def network_topology_test(self):
...
+    Test to reproduce CASSANDRA-10238, wherein changing snitch properties to change racks without a restart
+    could violate RF contract.
+
+    Since CASSANDRA-10243 it is no longer possible to change rack or dc for live nodes so we must specify
+    which nodes should be shutdown in order to have the rack changed.
     """
+
+    def __init__(self, *args, **kwargs):
+        Tester.__init__(self, *args, **kwargs)
+        self.ignore_log_patterns = ["Fatal exception during initialization",
```
mambocab: I believe this would be equivalent to setting `ignore_log_patterns` at the class level; if so, we should do that, rather than mess with `__init__` machinery.

replication_test.py
```diff
@@ -474,13 +512,171 @@ def _test_rf_on_snitch_update(self, nodes, rf, snitch_class_name, snitch_config_
...
+
+        cluster.start(wait_for_binary_proto=True)
+
+        # check racks are as specified
+        self.wait_for_nodes_on_racks(cluster.nodelist(), racks)
+
+        marks = [node.mark_log() for node in cluster.nodelist()]
+
+        debug("Updating snitch file")
+        for i, node in enumerate(cluster.nodelist()):
```
mambocab: I don't think `i` is used in this loop, is that intentional?

replication_test.py
```diff
@@ -474,13 +512,171 @@ def _test_rf_on_snitch_update(self, nodes, rf, snitch_class_name, snitch_config_
...
+
+    def _test_failed_snitch_update(self, nodes, snitch_class_name, snitch_config_file,
+                                   snitch_lines_before, snitch_lines_after, racks, error):
+        cluster = self.cluster
+        cluster.populate(nodes)
+        cluster.set_configuration_options(values={'endpoint_snitch': 'org.apache.cassandra.locator.{}'
+                                          .format(snitch_class_name)})
+
+        # start with initial snitch lines
+        for i, node in enumerate(cluster.nodelist()):
```
mambocab: Same comment as below -- it doesn't look like `i` is used in this loop, is that intentional?

replication_test.py
ptnapoleon: Don't we always do it like this?

mambocab:
This is what I meant to merge instead of #605. Thanks for the tests; I'm going to merge these to get the tests passing, then fix my style nits myself.

replication_test.py
mambocab: I'm not sure, but regardless, it shouldn't be necessary.

ptnapoleon:
Discussion [and code] moving to #693

PR #692, opened 2015-11-30
Revert "Fixes to replication_test.py for CASSANDRA-10243"

ptnapoleon:
Reverts riptano/cassandra-dtest#688


ptnapoleon:
Tests don't pass. If you check the CI for these tests

http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-10243-dtest/3/testReport/

You see they were all skipped, which is why we didn't see the failures before merging.

PR #696, opened 2015-12-01
Tests for CASSANDRA-9474 (v2)

pauloricardomg:


replication_test.py
```diff
@@ -672,3 +672,35 @@ def _test_failed_snitch_update(self, nodes, snitch_class_name, snitch_config_fil
...
             for node, mark in zip(cluster.nodelist(), marks):
                 node.watch_log_for(error, from_mark=mark)
+
+    def test_switch_data_center_startup_fails(self):
+        """
+        @jira_ticket CASSANDRA-9474
+
+        Confirm that switching data centers fails to bring up the node.
+        """
+        expected_error = r"Cannot start node if snitch's data center (.*) differs from previous data center (.*)\." \
```
mambocab: I believe this would be cleaner as

```python
expected_error = (r"Cannot start node if snitch's data center (.*) differs from previous data center (.*)\. "
                  "Please fix the snitch configuration, decommission and rebootstrap this node or use the flag -Dcassandra.ignore_dc=true.")
```

The two strings literals separated by only whitespace are parsed as a single string literal, and surrounding it in parentheses allows you to use line breaks there.

ptnapoleon:
Lgtm

mambocab:
Ran this test on all the branches linked in CASSANDRA-9474, and it passed. As far as I'm concerned, go ahead and merge once that ticket is resovled.

PR #698, opened 2015-12-09
CASSANDRA-10666 - write less data and set compactionthroughput lower

krummas:


ptnapoleon:
This was +1'd on jira.

PR #700, opened 2015-12-09
Known failure decorator

mambocab:
Adds simple decorator that tags a known failing test with some metadata about the failure. Also tags a couple tests for demo purposes. I smoke-tested this with

```
CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure -v
CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure '!known_flaky' -v
CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure=cassandra -v
CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure=systemic -v
```

Which demonstrates the basic functionality of the decorator.

Printing out associated ticket URLs is a little more difficult and will involve writing a small nose plugin, I believe. 

@ptnapoleon to review and give feedback. In particular

- Are these tags the model of known failures we want to present?
  - In particular, is it acceptable that we can only tag a given test with one kind of failure? I'm trying to avoid a huge proliferation of attributes, but we may want to tag some tests as, e.g. flaky because of a systemic issue, but also failing hard because of a C\* bug.
- Will this output be consumable in the way that the failure dashboard needs? Expect that, with the above-mentioned nose plugin, it will print the ticket URL after the name and module of the test.

### DIFF
--- a/concurrent_schema_changes_test.py
+++ b/concurrent_schema_changes_test.py
@@ -10,7 +10,7 @@ from cassandra.concurrent import execute_concurrent
 
 from ccmlib.node import Node
 from dtest import Tester, debug
-from tools import require, since
+from tools import known_failure, require, since
 
 
 def wait(delay=2):
@@ -21,6 +21,9 @@ def wait(delay=2):
 
 
 @require(10699)
+@known_failure(failure_source='cassandra',
+               jira_url='https://issues.apache.org/jira/browse/CASSANDRA-10699',
+               flaky=True)
 class TestConcurrentSchemaChanges(Tester):
 
     def __init__(self, *argv, **kwargs):
--- a/tools.py
+++ b/tools.py
@@ -356,7 +356,8 @@ def known_failure(failure_source, jira_url, flaky=False):
     assert isinstance(flaky, bool)
 
     def wrapper(f):
-        tagged_func = attr(known_failure=failure_source)(f)
+        tagged_func = attr(known_failure=failure_source,
+                           jira_url=jira_url)(f)
         if flaky:
             tagged_func = attr('known_flaky')(tagged_func)
         return tagged_func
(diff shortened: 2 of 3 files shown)


concurrent_schema_changes_test.py
```diff
@@ -21,6 +21,9 @@ def wait(delay=2):
 
 
 @require(10699)
+@known_failure(failure_source='cassandra',
+               jira_url='https://issues.apache.org/jira/browse/CASSANDRA-10699',
```
ptnapoleon: Surely we can change this to just jira, and autogenerate the url prefix?

tools.py
```diff
@@ -326,6 +326,44 @@ def tag_and_skip(decorated):
...
+        $ nosetests -a known_failure=cassandra -a !known_flaky
+
+    Known limitations: a given test may only be tagged once and still work as
+    expected with the attrib plugin machinery; if you decorate a test with
+    known_failure multiple times, the known_failure attribute of that test
+    will have the value applied by the outermost instance of the decorator.
+    """
+    valid_failure_sources = ('cassandra', 'test', 'systemic')
+    assert failure_source in valid_failure_sources
+    assert isinstance(flaky, bool)
```
ptnapoleon: Why use naked asserts here? Especially for the `in`

ptnapoleon:
I like that you can tag a test multiple times, but only the outermost is printed/resolved. That lets us pop issues off a test's stack as we fix them.

I like having flaky as a boolean, attachable to any failure type, that's clever.  I'll run this locally to examine the output, since you didn't leave me any >:(.

mambocab:
Closing in favor of #701

PR #701, opened 2015-12-10
Known failure decorator

mambocab:
Supercedes #700.

> Adds simple decorator that tags a known failing test with some metadata about the failure. Also tags a couple tests for demo purposes. I smoke-tested this with
> 
> ```
> CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure -v
> CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure '!known_flaky' -v
> CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure=cassandra -v
> CASSANDRA_VERSION=git:cassandra-3.0 nosetests --collect-only -a known_failure=systemic -v
> Which demonstrates the basic functionality of the decorator.
> ```
> Printing out associated ticket URLs is a little more difficult and will involve writing a small nose plugin, I believe.
> 
> @ptnapoleon to review and give feedback. In particular
> 
> - Are these tags the model of known failures we want to present?
>   - In particular, is it acceptable that we can only tag a given test with one kind of failure? I'm trying to avoid a huge proliferation of attributes, but we may want to tag some tests as, e.g. flaky because of a systemic issue, but also failing hard because of a C* bug.
> - Will this output be consumable in the way that the failure dashboard needs? Expect that, with the above-mentioned nose plugin, it will print the ticket URL after the name and module of the test.

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -356,7 +356,8 @@ def known_failure(failure_source, jira_url, flaky=False):
     assert isinstance(flaky, bool)
 
     def wrapper(f):
-        tagged_func = attr(known_failure=failure_source)(f)
+        tagged_func = attr(known_failure=failure_source,
+                           jira_url=jira_url)(f)
         if flaky:
             tagged_func = attr('known_flaky')(tagged_func)
         return tagged_func
--- a/tools.py
+++ b/tools.py
@@ -352,10 +352,11 @@ def known_failure(failure_source, jira_url, flaky=False):
     will have the value applied by the outermost instance of the decorator.
     """
     valid_failure_sources = ('cassandra', 'test', 'systemic')
-    assert failure_source in valid_failure_sources
-    assert isinstance(flaky, bool)
 
     def wrapper(f):
+        assert failure_source in valid_failure_sources
+        assert isinstance(flaky, bool)
+
         tagged_func = attr(known_failure=failure_source,
                            jira_url=jira_url)(f)
         if flaky:
--- a/tools.py
+++ b/tools.py
@@ -326,6 +326,43 @@ def require(require_pattern, broken_in=None):
         return tagging_decorator
 
 
+def known_failure(failure_source, jira_url, flaky=False):
+    """
+    Tag a test as a known failure. Associate it with the URL for a JIRA
+    ticket and tag it as flaky or not.
+
+    Valid values for failure_source include: 'cassandra', 'test', and
+    'systemic'.
+
+    To run all known failures, use the functionality provided by the nosetests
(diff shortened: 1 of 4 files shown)


mambocab:
https://github.com/riptano/cassandra-dtest/pull/700#discussion_r47233257

> Surely we can change this to just jira, and autogenerate the url prefix? Since we only need the url in the output.

We can. MVP tho, dude!

tools.py
```diff
@@ -326,6 +326,44 @@ def tag_and_skip(decorated):
...
+        $ nosetests -a known_failure=cassandra -a !known_flaky
+
+    Known limitations: a given test may only be tagged once and still work as
+    expected with the attrib plugin machinery; if you decorate a test with
+    known_failure multiple times, the known_failure attribute of that test
+    will have the value applied by the outermost instance of the decorator.
+    """
+    valid_failure_sources = ('cassandra', 'test', 'systemic')
+    assert failure_source in valid_failure_sources
+    assert isinstance(flaky, bool)
```
mambocab: https://github.com/riptano/cassandra-dtest/pull/700#discussion_r47233330

> Why use naked asserts here? Especially for the `in`.

My thought was that they're not test assertions, they're test harness assertions.

I do need to move these into the `wrapper` rather than the code that generates it. I want as few opportunities to raise errors at load time as possible, given how bad we are at dealing with `nose`'s error reporting on that class of problem.

ptnapoleon:
Yeah +! then.

mambocab:
https://github.com/riptano/cassandra-dtest/pull/700#issuecomment-163645009

> I like that you can tag a test multiple times, but only the outermost is printed/resolved. That lets us pop issues off a test's stack as we fix them.

Is there a persistent logical stack of issues that cause a given test to fail, though?

ptnapoleon:
Yes? That happens occasionally.

mambocab:
all right, good enough. Let me fix the placement of the assertions, and then I'll ask for a re-review.

mambocab:
ping @ptnapoleon -- have another look?

ptnapoleon:
+1

mambocab:
Requesting re-review for the code. I've added a script that outputs the added test attributes in JSON. I'll be adding a docstring momentarily.

ptnapoleon:
This dumps out one json object per test. Admittedly, I don't think you can fix that, but I think I can deal with it on the consumer end.

ptnapoleon:
I'm +1 on this script

mambocab:
Thanks for review. Merging.

PR #702, opened 2015-12-11
Check for correct log message on 2.1

ptnapoleon:
@mambocab to review. I broke this in the first place.

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -268,7 +268,8 @@ class TestCompaction(Tester):
         node.flush()
 
         node.nodetool('compact ks large')
-        node.watch_log_for('Writing large partition ks/large:user \(\d+ bytes\)', from_mark=mark, timeout=180)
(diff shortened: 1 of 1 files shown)


compaction_test.py
```diff
@@ -268,7 +268,11 @@ def large_compaction_warning_test(self):
         node.flush()
 
         node.nodetool('compact ks large')
-        node.watch_log_for('Writing large partition ks/large:user \(\d+ bytes\)', from_mark=mark, timeout=180)
+        if self.cluster.version() > '2.2':
```
mambocab: Go ahead and merge as-is if you like, but this whole block could be

```python
verb = 'Writing' if self.cluster.version() > '2.2' else 'Compacting'
```

mambocab:
This is fine as-is. Could be terser, but it's up to you if you want to make that change.

ptnapoleon:
Correction implemented, thanks.

PR #703, opened 2015-12-11
name tests so nose will pick them up

mambocab:
A number of tests were incorrectly named -- they didn't begin with 'test_' or end with '_test', so they weren't picked up by nose and so weren't run by nosetests. I missed this in review. @ptnapoleon to review.

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -233,7 +233,7 @@ class MiscellaneousCQLTester(CQLTester):
     required.
     """
 
-    def large_collection_errors(self):
+    def large_collection_errors_test(self):
         """
         For large collections, make sure that we are printing warnings.
         """
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -76,7 +76,7 @@ class TestCQL(UpgradeTester):
                 [UUID('550e8400-e29b-41d4-a716-446655440000'), 36, None, None],
             ], res
 
-    def large_collection_errors(self):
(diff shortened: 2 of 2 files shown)


ptnapoleon:
These all pass locally for me, except the first, which skips. +1, assuming that test is supposed to skip

mambocab:
Added another one, from `cql_tests.py`.

@thobbs `large_collection_error` needs to be skipped for versions where that protocol isn't supported, I believe. Can you verify that? and if so, when was protocol v2 deprecated?

thobbs:
> @thobbs large_collection_error needs to be skipped for versions where that protocol isn't supported, I believe. Can you verify that? and if so, when was protocol v2 deprecated? 

That's correct.  The v2 protocol was deprecated in 2.2 and removed in 3.0, so we should be able to test this on 2.1 and 2.2.

ptnapoleon:
@mambocab, should we merge this?

ptnapoleon:
The tests all pass, except for the one that Tyler confirms should be skipped. I'm merging.

mambocab:
@ptnapoleon thanks for merging, sorry I let this sit. Could you add the appropriate @since?

ptnapoleon:
What since is that?

mambocab:
the one for skipping large_collection_errors_test on... 3.1+?

ptnapoleon:
Added. And you meant 3.0+, as that's when v2 was removed.

mambocab:
Thanks.

PR #704, opened 2015-12-11
add `notes` param to `known_failure`

mambocab:
Small addition to the `known_failure` decorator. Just adds a freeform 'notes' attribute; it's not used in a meaningful way yet, but could be in the future, and might be better than comments for adding metadata like "this fails on 3.0+", which we can't handle at the moment.

@ptnapoleon to review.

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -326,7 +326,7 @@ def require(require_pattern, broken_in=None):
         return tagging_decorator
 
 
-def known_failure(failure_source, jira_url, flaky=False):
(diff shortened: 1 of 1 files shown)


mambocab:
Travis is mad for reasons unrelated to this PR:

https://travis-ci.org/riptano/cassandra-dtest/builds/96353608#L114

ptnapoleon:
+1. I fixed the travis complaint, fyi.

mambocab:
Cheers!

PR #705, opened 2015-12-11
upgrade test for CASSANDRA-10822

knifewine:


knifewine:
This is regrettably imperfect at triggering the bug, but it's better than nothing, and passes without exception on patched code.

ptnapoleon:
Is it possible to loop the entire test a few times to increase catch rate?

upgrade_tests/regression_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class TestForRegressions(UpgradeTester):
+    """
+    Catch-all class for regression tests on specific versions.
+    """
+    NODES, RF, __test__, CL = 2, 1, True, CL.ONE
+
+    def test_10822(self):
+        session = self.prepare()
```
ptnapoleon: You forgot to put in a docstring
```
"""
@jira_ticket CASSANDRA-10822
"""
```

knifewine:
Thanks to some great advice from @cdaw, writing more paritions and tombstones gets this to reliably pass/fail when it should. 10/10 failures unpatched, 10/10 pass on patched code.

ptnapoleon:
Would it be possible to create a list of [symbol, year] tuples, and iterate through that for each of these logical chunks? It would make the code smaller, and more easily maintained. Especially since we re-use the same set of values three times.

knifewine:
I don't expect this test to have a long lifespan but you make a good point. Committing momentarily.

knifewine:
Tested for good measure, still has reliable pass/fail. I'll squash before merging once I've got a +1.

ptnapoleon:
`for (symbol, year) in symbol_years:` can just be `for symbol, year in symbol_years:`, but otherwise, +1

knifewine:
yeah, I guess I like the explicit style there, but not a real strong opinion about it. updated and pushed a squashed version of this all.

PR #708, opened 2015-12-14
Adjust expected trace output

ptnapoleon:
I still have to check with a dev that this change is intentional, but it definitely doesn't seem accidental.


ptnapoleon:
No one will answer me on #dev and I can't find the change in Jira.

ptnapoleon:
@mambocab thoughts?

ptnapoleon:
Found it in CASSANDRA-9479

mambocab:
Well now we've got problems
And I think rebasing can solve them

mambocab:
Looks good, consistent with the description in the ticket. +1.

PR #709, opened 2015-12-14
Use FallthroughRetryPolicy on failure_threshold_deletions

ptnapoleon:
On 2.1 we were getting OperationTimedout instead of ReadTimeout.
This was because we now have a default retry policy set.
On retry, we were ignoring the ReadTimeout error that was
correctly being sent by C*. By overring the retry policy,
we have stopped retrying, and now propagate the correct exception.

@knifewine to review.


ptnapoleon:
I accidentally put other changes in this PR. Feel free to also review those?

knifewine:
+1 on both commits!

PR #712, opened 2015-12-15
add 3.1 to upgrade_through_versions

knifewine:


knifewine:
paging @ptnapoleon

knifewine:
damnit, need to remove that last line, one sec.

knifewine:
Looks like there's an assumption being made in the code that's causing it to break on 3.1, so not yet ready for merge.

ptnapoleon:
I'll hold off on checking it then

knifewine:
@ptnapoleon now that we have a 3.1 tag this should work fine.

ptnapoleon:
+1

ptnapoleon:
Don't we also want a path that goes from 3.0 to trunk, skipping 3.1?

knifewine:
Don't we also want a path that goes from 3.0 to trunk, skipping 3.1?

quite possibly, but I think that will need to wait until we better figure things out

PR #713, opened 2015-12-16
Add test for CASSANDRA-10806

yukim:
Added test to check sstableloader works with upper case keyspace name.
This is test for [CASSANDRA-10806](https://issues.apache.org/jira/browse/CASSANDRA-10806).

### DIFF
--- a/sstable_generation_loading_test.py
+++ b/sstable_generation_loading_test.py
@@ -5,7 +5,6 @@ import time
 
 from dtest import Tester, debug
 from ccmlib import common as ccmcommon
-from tools import require
 
 
 class TestSSTableGenerationAndLoading(Tester):
@@ -111,7 +110,6 @@ class TestSSTableGenerationAndLoading(Tester):
     def sstableloader_compression_deflate_to_deflate_test(self):
         self.load_sstable_with_configuration('Deflate', 'Deflate')
 
-    @require("10806")
     def sstableloader_uppercase_keyspace_name_test(self):
         """
         Make sure sstableloader works with upper case keyspace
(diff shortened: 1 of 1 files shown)


sstable_generation_loading_test.py
```diff
@@ -110,7 +111,15 @@ def sstableloader_compression_deflate_to_snappy_test(self):
     def sstableloader_compression_deflate_to_deflate_test(self):
         self.load_sstable_with_configuration('Deflate', 'Deflate')
 
-    def load_sstable_with_configuration(self, pre_compression=None, post_compression=None):
+    @require("10806")
```
mambocab: How long do you believe it will take to get CASSANDRA-10806 merged? We can wait to merge this until then, and not worry about using `@require` here.

sstable_generation_loading_test.py
```diff
@@ -176,7 +185,11 @@ def create_schema(session, compression):
         sstableloader = os.path.join(cdir, 'bin', ccmcommon.platform_binary('sstableloader'))
         env = ccmcommon.make_cassandra_env(cdir, node1.get_path())
         host = node1.address()
-        sstablecopy_dir = copy_root + '/ks'
+        if ks.startswith('"') and ks.endswith('"'):
```
mambocab: I believe this whole `if`-`else` block could be replaced with `ks = ks.strip('"')`:

```python
>>> '"Keyspace1"'.strip('"')
'Keyspace1'
>>> 'Keyspace1'.strip('"')
'Keyspace1'
```

sstable_generation_loading_test.py
```diff
@@ -176,7 +185,11 @@ def create_schema(session, compression):
...
+            ks_dir = ks[1:-1]
+        else:
+            ks_dir = ks
+        sstablecopy_dir = copy_root + '/' + ks_dir
```
mambocab: I know you didn't introduce this problem, but could you please change this to use `os.path.join`?

```python
sstablecopy_dir = os.path.join(copy_root, ks_dir)
```

mambocab:
After addressing my comments, I believe the code will be fine. It looks like CASSANDRA-10806 is on track to be merged soon, so I think we should hold off on merging this PR until then?

yukim:
thanks for review. I will commit 10806 first then.

yukim:
Merged CASSANDRA-10806.

mambocab:
Great. I've run this locally and it passes; if you'll address my comments and push your changes, I can run again locally and merge if they work.

ptnapoleon:
Still passing after the changes, merging.

PR #715, opened 2015-12-18
Fixes for CASSANDRA-9472

stef1927:
I enhanced `jmxmetrics_test.TestJMXMetrics.begin_test` to cope with offheap memtables, else it will consistently fail after CASSANDRA-9472 is committed.

I believe the known failure described in CASSANDRA-10846 for `sstableofflinerelevel_test` is probably due to a lack of overlapping tables. I had it consistently failing on the 9472 branch with offheap memtable objects whilst without offheap memtable ojects it would pass consistently. After investigating I discovered it was passing only because of 1 overlapping table. Increasing the number of records from 100K to 200K fixed it in both scenarios.


ptnapoleon:
Both changes seem good, +1. Jim is running it locally to confirm.

mambocab:
Works locally for me. +1, merging.

PR #716, opened 2015-12-18
changes default arg from dict to None

mambocab:
I don't think that `options` is manipulated in a way that would cause
problems, but we're asking for trouble leaving it like this.

@ptnapoleon, can you review? It's teeny, I don't see any potential for issue here


ptnapoleon:
Looks fine, +1

mambocab:
Thanks.

PR #717, opened 2015-12-18
Add test for CASSANDRA-10887

ptnapoleon:
@mambocab and @blambov, I am following the repro steps laid out on https://issues.apache.org/jira/browse/CASSANDRA-10887 , but I can't seem to repro with this test. Do you see any issues?

### DIFF
--- a/pending_range_test.py
+++ b/pending_range_test.py
@@ -1,0 +1,68 @@
+from cassandra.query import SimpleStatement
+from dtest import TRACE, Tester, debug
+from tools import known_failure, no_vnodes
+
+
+@no_vnodes()
+class TestPendingRangeMovements(Tester):
+
+    @known_failure(failure_source='cassandra',
+                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-10887')
+    def pending_range_test(self):
+        """
+        @jira_ticket CASSANDRA-10887
+        """
+        cluster = self.cluster
+        # If we are on 2.1, we need to set the log level to debug or higher, as debug.log does not exist.
+        if cluster.version() < '2.2' and not TRACE:
+            cluster.set_log_level('DEBUG')
+
+        # Create 5 node cluster
+        cluster.populate(5).start(wait_for_binary_proto=True)
+        node1, node2 = cluster.nodelist()[0:2]
+
+        # Set up RF=3 keyspace
+        session = self.patient_cql_connection(node1)
+        self.create_ks(session, 'ks', 3)
+
+        session.execute("CREATE TABLE users (login text PRIMARY KEY, email text, name text, login_count int)")
+
+        # We use the partition key 'jdoe3' because it belongs to node1.
+        # The key MUST belong to node1 to repro the bug.
+        session.execute("INSERT INTO users (login, email, name, login_count) VALUES ('jdoe3', 'jdoe@abc.com', 'Jane Doe', 1) IF NOT EXISTS;")
+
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I found the problem. I'll push up the changes, but the patch on 10887 doesn't merge cleanly up past 2.0, so I can't actually check the test passes with it on newer versions.

ptnapoleon:
@mambocab this is ready for review. It should error on the last set of LWT queries on every version, 2.1+. It should NOT error anywhere earlier. I haven't tested with the patch yet, as I can't get it to compile on 2.1+, even if I resolve the merge conflicts myself.

pending_range_test.py
```diff
@@ -0,0 +1,67 @@
...
+@no_vnodes()
+class TestPendingRangeMovements(Tester):
+
+    def pending_range_test(self):
+        """
+        @jira_ticket CASSANDRA-10887
+        """
+        cluster = self.cluster
+        # If we are on 2.1, we need to set the log level to debug or higher, as debug.log does not exist.
+        if cluster.version() < '2.2' and os.environ.get('TRACE', '').lower() not in ('yes', 'true'):
```
mambocab: Please use `dtest.TRACE` instead of redoing the work of grabbing `TRACE` from the env.

pending_range_test.py
```diff
@@ -0,0 +1,67 @@
...
+            cluster.set_log_level('DEBUG')
+
+        # Create 5 node cluster
+        cluster.populate(5).start(wait_for_binary_proto=True)
+        node1, node2 = cluster.nodelist()[0:2]
+
+        # Set up RF=3 table
```
mambocab: `s/table/keyspace`

pending_range_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # Watch the log so we know when the node is moving
+        node1.watch_log_for('Moving .* to {}'.format(token), timeout=10, from_mark=mark)
+        node1.watch_log_for('Sleeping 30000 ms before start streaming/fetching ranges', timeout=10, from_mark=mark)
+
+        if cluster.version() >= '2.2':
+            node2.watch_log_for('127.0.0.1 state moving', timeout=10, filename='debug.log')
+        elif cluster.version >= '2.1':
+            node2.watch_log_for('127.0.0.1 state moving', timeout=10, filename='system.log')
+        # 2.1 doesn't have debug.log, and so doesn't log this message.
```
mambocab: This is confusing. Just above, on 2.1, we checked in `system.log`; this comment describes something else.

pending_range_test.py
```diff
@@ -0,0 +1,67 @@
...
+        # Move a node
+        node1.nodetool('move {}'.format(token), capture_output=False, wait=False)
+
+        # Watch the log so we know when the node is moving
+        node1.watch_log_for('Moving .* to {}'.format(token), timeout=10, from_mark=mark)
+        node1.watch_log_for('Sleeping 30000 ms before start streaming/fetching ranges', timeout=10, from_mark=mark)
+
+        if cluster.version() >= '2.2':
+            node2.watch_log_for('127.0.0.1 state moving', timeout=10, filename='debug.log')
+        elif cluster.version >= '2.1':
```
mambocab: Do we care about the test's behavior pre-2.1? If not, this could just be an `else`.

mambocab:
AFAIC, tag it with `@known_failure` and merge it after my nits are addressed.

pending_range_test.py
ptnapoleon: I tried to find that variable but couldn't. Didn't realize dtest.TRACE would get me there

PR #718, opened 2015-12-22
CASSANDRA-9494 was only committed on trunk

stef1927:
We need a `@since` for `test_reading_with_ttl` because CASSANDRA-9494 is only on trunk, at least for the time being.


mambocab:
Great, thank you. Merging.

PR #603, opened 2015-10-13
For CASSANDRA-9526, add dtest for nodetool and JMX functionality to e…

aweisberg:
…xtract phti values from failure detector

This is pending review of CASSANDRA-9526

PR #720, opened 2015-12-22
C-9526

ptnapoleon:
Rebase of #603 
@mambocab to review

### DIFF
--- a/jmx_test.py
+++ b/jmx_test.py
@@ -203,7 +203,6 @@ class TestJMX(Tester):
                 time.sleep(2)
 
     @since('2.2')
-    @require(9526)
     def phi_test(self):
         """
         Check functioning of nodetool failuredetector.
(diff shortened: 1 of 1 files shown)


jmx_test.py
```diff
@@ -201,3 +201,34 @@ def test_compactionstats(self):
...
+        """
+        Check functioning of nodetool failuredetector.
+        @jira_ticket CASSANDRA-9526
+        """
+
+        cluster = self.cluster
+        cluster.populate(3).start(wait_for_binary_proto=True)
+        node1, node2, node3 = cluster.nodelist()
+
+        phivalues = node1.nodetool("failuredetector")[0].splitlines()
```
mambocab: It worries me a little that there's no validation of the output.

jmx_test.py
```diff
@@ -201,3 +201,34 @@ def test_compactionstats(self):
...
+        endpoint1Values = phivalues[1].split()
+        endpoint2Values = phivalues[2].split()
+
+        endpoint1 = endpoint1Values[0][1:-1]
+        endpoint2 = endpoint2Values[0][1:-1]
+
+        self.assertTrue(endpoint1.startswith("127.0.0"))
+        self.assertTrue(endpoint2.startswith("127.0.0"))
```
mambocab: Well, OK, not _no_ validation, but this seems insufficient.

ptnapoleon: Well one of these should be 127.0.0.2, and the other should be 127.0.0.3. I think the ordering is just not guaranteed. I'll improve the validation

mambocab:
Code looks good. I accidentally ran this with `--no-skip` and it passed on 2.2 and 2.1, which worries me.

ptnapoleon:
Don't worry about that.

mambocab:
Operator error on my part, not a problem

PR #721, opened 2015-12-23
Tests for CASSANDRA-10854

stef1927:
Two tests for CASSANDRA-10854, I did not use `@require` since the ticket is already in status `ready to commit` so you may want to wait until the ticket is committed before merging or I can merge myself once I have your +1 and the ticket is committed.


ptnapoleon:
+1

PR #722, opened 2015-12-23
Also set the cluster install dir, so ccm can track version

ptnapoleon:
@knifewine to review, and also run locally, as it fails on OSX because it's using 2.1.3.


knifewine:
+1

ptnapoleon:
New failure, but is now product related. Merging this and I'll file the bug.

PR #723, opened 2015-12-23
Remove upgrade ver magic

knifewine:
This includes some other simplifications and problems I found along the way, notably config wasn't being set properly for some upgrade tests, and some schema metadata code was trying to create database stuff with invalid names (in the case of bare versions like 1.2.3).

Also included is a small refactor to Tester to support a hook on subclasses for setting configuration. This is probably less drastic than we might like but it should keep other tests behaving the same without any surprising changes to config (for better or worse).

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -263,10 +263,10 @@ class Tester(TestCase):
         else:
             node.set_install_dir(install_dir=cdir)
 
-    def establish_config(self):
+    def init_config(self):
         raise NotImplementedError()
 
-    def establish_default_config(self):
+    def init_default_config(self):
         # the failure detector can be quite slow in such tests with quick start/stop
         self.cluster.set_configuration_options(values={'phi_convict_threshold': 5})
 
--- a/dtest.py
+++ b/dtest.py
@@ -5,6 +5,7 @@ import copy
 import errno
 import logging
 import os
+import pprint
 import re
 import shutil
 import subprocess
@@ -16,17 +17,16 @@ import traceback
 import types
 from unittest import TestCase
 
+from cassandra import ConsistencyLevel
 from cassandra.auth import PlainTextAuthProvider
 from cassandra.cluster import Cluster as PyCluster
 from cassandra.cluster import NoHostAvailable
-from cassandra.policies import WhiteListRoundRobinPolicy, RetryPolicy
-from cassandra import ConsistencyLevel
-from nose.exc import SkipTest
-
+from cassandra.policies import RetryPolicy, WhiteListRoundRobinPolicy
 from ccmlib.cluster import Cluster
 from ccmlib.cluster_factory import ClusterFactory
 from ccmlib.common import is_win
 from ccmlib.node import TimeoutError
+from nose.exc import SkipTest
 
 LOG_SAVED_DIR = "logs"
 try:
@@ -263,6 +263,28 @@ class Tester(TestCase):
         else:
             node.set_install_dir(install_dir=cdir)
 
+    def establish_config(self):
+        raise NotImplementedError()
+
+    def establish_default_config(self):
+        # the failure detector can be quite slow in such tests with quick start/stop
+        self.cluster.set_configuration_options(values={'phi_convict_threshold': 5})
+
+        timeout = 10000
+        if self.cluster_options is not None:
+            self.cluster.set_configuration_options(values=self.cluster_options)
+        else:
+            self.cluster.set_configuration_options(values={
+                'read_request_timeout_in_ms': timeout,
+                'range_request_timeout_in_ms': timeout,
+                'write_request_timeout_in_ms': timeout,
+                'truncate_request_timeout_in_ms': timeout,
+                'request_timeout_in_ms': timeout
+            })
+
+        debug("Done setting configuration options:")
+        debug(pprint.pformat(self.cluster._config_options, indent=4))
+
     def setUp(self):
         global CURRENT_TEST
         CURRENT_TEST = self.id() + self._testMethodName
@@ -303,20 +325,12 @@ class Tester(TestCase):
         self.cluster = self._get_cluster()
         if RECORD_COVERAGE:
             self.__setup_jacoco()
-        # the failure detector can be quite slow in such tests with quick start/stop
-        self.cluster.set_configuration_options(values={'phi_convict_threshold': 5})
 
-        timeout = 10000
-        if self.cluster_options is not None:
-            self.cluster.set_configuration_options(values=self.cluster_options)
-        else:
-            self.cluster.set_configuration_options(values={
-                'read_request_timeout_in_ms': timeout,
-                'range_request_timeout_in_ms': timeout,
-                'write_request_timeout_in_ms': timeout,
-                'truncate_request_timeout_in_ms': timeout,
-                'request_timeout_in_ms': timeout
-            })
+        try:
+            self.establish_config()
+        except NotImplementedError:
+            debug("Custom establish_config not found. Setting defaults.")
+            self.establish_default_config()
 
         with open(LAST_TEST_DIR, 'w') as f:
             f.write(self.test_path + '\n')
--- a/dtest.py
+++ b/dtest.py
@@ -326,10 +326,10 @@ class Tester(TestCase):
             self.__setup_jacoco()
 
         try:
-            self.establish_config()
+            self.init_config()
         except NotImplementedError:
-            debug("Custom establish_config not found. Setting defaults.")
-            self.establish_default_config()
+            debug("Custom init_config not found. Setting defaults.")
+            self.init_default_config()
 
         with open(LAST_TEST_DIR, 'w') as f:
             f.write(self.test_path + '\n')
--- a/dtest.py
+++ b/dtest.py
@@ -282,8 +282,7 @@ class Tester(TestCase):
                 'request_timeout_in_ms': timeout
             })
 
-        debug("Done setting configuration options:")
-        debug(pprint.pformat(self.cluster._config_options, indent=4))
+        debug("Done setting configuration options:\n" + pprint.pformat(self.cluster._config_options, indent=4))
 
     def setUp(self):
         global CURRENT_TEST
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -280,6 +280,7 @@ class UpgradeTester(Tester):
             # and when it does, it gets replayed and everything is fine.
             r'Can\'t send migration request: node.*is down',
             r'RejectedExecutionException.*ThreadPoolExecutor has shut down',
+            r'Cannot update data center or rack from.*for live host'  # occurs due to test/ccm writing topo on down nodes
         ]
         self.subprocs = []
         Tester.__init__(self, *args, **kwargs)
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -12,14 +12,14 @@ from multiprocessing import Process, Queue
 from Queue import Empty, Full
 
 import psutil
-import schema_metadata_test
 from cassandra import ConsistencyLevel, WriteTimeout
 from cassandra.query import SimpleStatement
+
+import schema_metadata_test
 from dtest import Tester, debug
 from tools import generate_ssl_stores, known_failure, new_node
 
 trunk_version = '3.2'
-DEFAULT_PROTOCOL_VERSION = int(os.environ.get('DEFAULT_PROTOCOL_VERSION', 3))
 latest_2dot0 = '2.0.17'
 latest_2dot1 = '2.1.12'
 latest_2dot2 = '2.2.4'
@@ -267,7 +267,6 @@ class UpgradeTester(Tester):
     test_versions = None  # set on init to know which versions to use
     subprocs = None  # holds any subprocesses, for status checking and cleanup
     extra_config = None  # holds a non-mutable structure that can be cast as dict()
-    protocol_version = DEFAULT_PROTOCOL_VERSION
 
     def __init__(self, *args, **kwargs):
         # Ignore these log patterns:
@@ -804,7 +803,7 @@ class BootstrapMixin(object):
                 );""")
 
 
-def create_upgrade_class(clsname, version_list, protocol_version=DEFAULT_PROTOCOL_VERSION,
+def create_upgrade_class(clsname, version_list, protocol_version,
                          bootstrap_test=False, extra_config=(('partitioner', 'org.apache.cassandra.dht.Murmur3Partitioner'),)):
     """
     Dynamically creates a test subclass for testing the given versions.
@@ -925,7 +924,8 @@ create_upgrade_class(
 create_upgrade_class(
     'ProtoV3Upgrade_3_1_UpTo_Trunk',
     [latest_3dot1, trunk_ccm_string],
-    bootstrap_test=True
+    bootstrap_test=True,
+    protocol_version=3
 )
 create_upgrade_class(
     'ProtoV3Upgrade_AllVersions',
@@ -971,7 +971,8 @@ create_upgrade_class(
 create_upgrade_class(
     'ProtoV4Upgrade_3_1_UpTo_Trunk',
     [latest_3dot1, trunk_ccm_string],
-    bootstrap_test=True
+    bootstrap_test=True,
+    protocol_version=4
 )
 create_upgrade_class(
     'ProtoV4Upgrade_AllVersions',
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -289,8 +289,8 @@ class UpgradeTester(Tester):
         switch_jdks(self.test_versions[0])
         self.cluster.set_install_dir(version=self.test_versions[0])
 
-    def establish_config(self):
-        self.establish_default_config()
+    def init_config(self):
+        self.init_default_config()
 
         if self.extra_config is not None:
             debug("Setting extra configuration options:\n" + pprint.pformat(dict(self.extra_config), indent=4))
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -14,6 +14,7 @@ from Queue import Empty, Full
 import psutil
 from cassandra import ConsistencyLevel, WriteTimeout
 from cassandra.query import SimpleStatement
+from six import print_
 
 import schema_metadata_test
 from dtest import Tester, debug
@@ -823,10 +824,10 @@ def create_upgrade_class(clsname, version_list, protocol_version,
     # short names for debug output
     parent_class_names = [cls.__name__ for cls in parent_classes]
 
-    print("Creating test class {} ".format(clsname))
-    print("  for C* versions: {} ".format(version_list))
-    print("  using protocol: v{}, and parent classes: {}".format(protocol_version, parent_class_names))
-    print("  to run these tests alone, use `nosetests {}.py:{}`".format(__name__, clsname))
+    print_("Creating test class {} ".format(clsname))
+    print_("  for C* versions: {} ".format(version_list))
+    print_("  using protocol: v{}, and parent classes: {}".format(protocol_version, parent_class_names))
+    print_("  to run these tests alone, use `nosetests {}.py:{}`".format(__name__, clsname))
 
     newcls = type(
         clsname,
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -20,12 +20,14 @@ import schema_metadata_test
 from dtest import Tester, debug
 from tools import generate_ssl_stores, known_failure, new_node
 
-trunk_version = '3.2'
+trunk_version = '3.4'
 latest_2dot0 = '2.0.17'
 latest_2dot1 = '2.1.12'
 latest_2dot2 = '2.2.4'
 latest_3dot0 = '3.0.2'
 latest_3dot1 = '3.1.1'
+latest_3dot2 = '3.2'
+latest_3dot3 = '3.3'
 trunk_ccm_string = 'git:trunk'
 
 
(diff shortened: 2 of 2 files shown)


knifewine:
cassci job here http://cassci.datastax.com/view/Upgrades/job/improved_upgrade_through_versions/

ptnapoleon:
This didn't appear to break any non-upgrade dtests, based on a CI run

dtest.py
```diff
@@ -16,17 +17,16 @@
...
 
+from cassandra import ConsistencyLevel
 from cassandra.auth import PlainTextAuthProvider
 from cassandra.cluster import Cluster as PyCluster
 from cassandra.cluster import NoHostAvailable
-from cassandra.policies import WhiteListRoundRobinPolicy, RetryPolicy
-from cassandra import ConsistencyLevel
-from nose.exc import SkipTest
-
+from cassandra.policies import RetryPolicy, WhiteListRoundRobinPolicy
```
ptnapoleon: Did you isort this? cassandra.policies should be in the same block as the other cassandra imports

dtest.py
```diff
@@ -5,6 +5,7 @@
 import errno
 import logging
 import os
+import pprint
```
ptnapoleon: Is this in the stdlib?

dtest.py
```diff
@@ -303,20 +324,12 @@ def setUp(self):
...
-            self.cluster.set_configuration_options(values={
-                'read_request_timeout_in_ms': timeout,
-                'range_request_timeout_in_ms': timeout,
-                'write_request_timeout_in_ms': timeout,
-                'truncate_request_timeout_in_ms': timeout,
-                'request_timeout_in_ms': timeout
-            })
+        try:
+            self.establish_config()
+        except NotImplementedError:
```
ptnapoleon: I understand the intent behind this, but I can't actually tell that we are realizing a benefit from this implementation.

upgrade_through_versions_test.py
```diff
@@ -15,160 +12,49 @@
...
-    _, ref_type, ref = _fullref.split('/')
-    MAPPED_REFS[ref_type][ref.split('^')[0]] = sha
+trunk_version = '3.2'
+DEFAULT_PROTOCOL_VERSION = int(os.environ.get('DEFAULT_PROTOCOL_VERSION', 3))
+latest_2dot0 = '2.0.17'
+latest_2dot1 = '2.1.12'
+latest_2dot2 = '2.2.4'
+latest_3dot0 = '3.0.2'
+latest_3dot1 = '3.1.1'
+trunk_ccm_string = 'git:trunk'
```
ptnapoleon: Won't updating these versions be difficult? And shouldn't we have `latest_tick` and `latest_tock`?

upgrade_through_versions_test.py
```diff
@@ -973,60 +803,198 @@ def _multidc_schema_create(self):
...
+def create_upgrade_class(clsname, version_list, protocol_version=DEFAULT_PROTOCOL_VERSION,
+                         bootstrap_test=False, extra_config=(('partitioner', 'org.apache.cassandra.dht.Murmur3Partitioner'),)):
+    """
+    Dynamically creates a test subclass for testing the given versions.
+
+    'clsname' is the name of the new class.
+    'protocol_version' is an int.
+    'bootstrap_test' is a boolean, if True bootstrap testing will be included. Default False.
+    'version_list' is a list of versions ccm will recognize, to be upgraded in order.
+    'extra_config' is tuple of config options that can (eventually) be cast as a dict,
```
ptnapoleon: Why isn't this just a dict from the start.

upgrade_through_versions_test.py
```diff
@@ -973,60 +803,198 @@ def _multidc_schema_create(self):
...
+    else:
+        parent_classes = (UpgradeTester,)
+
+    # short names for debug output
+    parent_class_names = [cls.__name__ for cls in parent_classes]
+
+    print("Creating test class {} ".format(clsname))
+    print("  for C* versions: {} ".format(version_list))
+    print("  using protocol: v{}, and parent classes: {}".format(protocol_version, parent_class_names))
+    print("  to run these tests alone, use `nosetests {}.py:{}`".format(__name__, clsname))
```
ptnapoleon: How do these work? I don't see us importing a `print()` anywhere.

ptnapoleon:
Have you given thought to how we'll support downgrading in these tests?

upgrade_through_versions_test.py
```diff
@@ -15,160 +12,49 @@
...
-GIT_LS = subprocess.check_output(["git", "ls-remote", "-h", "-t", REPO_LOCATION]).rstrip()
-
-# maps ref type (branch, tags) to ref names and sha's
-MAPPED_REFS = defaultdict(dict)
-for row in GIT_LS.split('\n'):
-    sha, _fullref = row.split('\t')
-    _, ref_type, ref = _fullref.split('/')
-    MAPPED_REFS[ref_type][ref.split('^')[0]] = sha
+trunk_version = '3.2'
+DEFAULT_PROTOCOL_VERSION = int(os.environ.get('DEFAULT_PROTOCOL_VERSION', 3))
```
ptnapoleon: Aren't we moving away from configuration in environment variables?

ptnapoleon:
Don't we need to test upgrading from X.Y.Z to X.Y.LATEST for all Z in X.Y?

dtest.py
knifewine: yes.

dtest.py
knifewine: I'm using an auto-isort provided by atom.io but maybe goofed this when resolving a conflict.

dtest.py
knifewine: It's intended as a place where dtest can establish sane global-ish defaults, with the ability to override them. But more importantly it shouldn't make waves with existing tests. This is probably more like a stepping stone to a better solution eventually. And it made the upgrade test config a lot less kludgy.

upgrade_through_versions_test.py
knifewine: This will be ongoing maintenance yes, but I wouldn't say difficult. Prior tests were burdened by trying to be too clever. What we need here is predictability moreso than anything else.

Even if we get a few versions behind, these are mostly used for intermediary versions, not the final upgrade in the path (which is typically going to be a branch). I think that would be a much rarer class of bug.

As for the tick/tock vars, seems like an interesting idea, need to give it some thought.

upgrade_through_versions_test.py
knifewine: Mutability hygeine. We're going to put that var into a new class (not instance), so don't want to risk sharing mutable refs. As the code is now, a dict would probably be safe, but if we goofed up it would be an obnoxious bug to track down.

upgrade_through_versions_test.py
knifewine: on python3 print() is a builtin. I think similar for python2 unless I'm misunderstanding the question.

upgrade_through_versions_test.py
knifewine: Most of them used for this suite we're eliminated (thank goodness). This particular one can probably go away when we solve the problem of mixed upgrade protocol version heterogeneity.

knifewine:
Downgrades should be possible, but aren't tackled in this effort.
We can easily expand the upgrade paths if you think there's missing coverage. More concrete examples would probably help.

upgrade_through_versions_test.py
knifewine: On second thought, maybe we can do without this. The one case I was seeing was if we make a bunch of calls to create_upgrade_class but don't specify a protocol version, the shell env could push a default into the tests. In practice maybe not a practical thing to plan for.

Maybe we remove the env var and use proto v4 when unspecified?

upgrade_through_versions_test.py
ptnapoleon: print() is not a built-in in python2, which is what dtest runs on. I have no idea how you got dtest to run in py3, as I assumed even Tester wasn't compliant.

dtest.py
```diff
@@ -263,6 +263,27 @@ def set_node_to_current_version(self, node):
         else:
             node.set_install_dir(install_dir=cdir)
 
+    def establish_config(self):
+        raise NotImplementedError()
+
+    def establish_default_config(self):
```
mambocab: I don't understand the use of the verb `establish` -- maybe `set` would be better? Is this different from `set`ing?

dtest.py
```diff
@@ -263,6 +263,27 @@ def set_node_to_current_version(self, node):
...
+        # the failure detector can be quite slow in such tests with quick start/stop
+        self.cluster.set_configuration_options(values={'phi_convict_threshold': 5})
+
+        timeout = 10000
+        if self.cluster_options is not None:
+            self.cluster.set_configuration_options(values=self.cluster_options)
+        else:
+            self.cluster.set_configuration_options(values={
```
mambocab: Do I recall correctly that there was a problem with not setting defaults per-value? Like, in the case where `self.cluster_options` is `None`, none of these options are set.

mambocab: Lol, never mind, just saw the changes below.

dtest.py
mambocab: In the long run, we might configure this the same way we do `ignore_log_patterns`. This is fine for now.

upgrade_through_versions_test.py
mambocab: It's probably better to require a protocol version than for us to maintain an env var. If we specify a default, will we run into trouble when protocols change? I'm ok with requiring it be specified; it isn't difficult for users to wrap the function and provide their own default.

upgrade_through_versions_test.py
mambocab: The way we do this elsewhere in the codebase is:

```python
def func(param_with_default_value_of_mutable_type=None):
    if param_with_default_value_of_mutable_type is None:
        param_with_default_value_of_mutable_type = default
```

This breaks down if `None` is a valid input, but I don't think that's the case here, and you can work around that by creating a sigil with `object()` anyway.

upgrade_through_versions_test.py
mambocab: IIRC this works in py2 because `(x)` evaluates to `x`. In any event, this is a py2-only codebase right now, and in that context this is confusing.

Shouldn't this use `debug` anyway?

upgrade_through_versions_test.py
```diff
@@ -973,60 +803,198 @@ def _multidc_schema_create(self):
...
+    print("  using protocol: v{}, and parent classes: {}".format(protocol_version, parent_class_names))
+    print("  to run these tests alone, use `nosetests {}.py:{}`".format(__name__, clsname))
+
+    newcls = type(
+        clsname,
+        parent_classes,
+        {'test_versions': version_list, '__test__': True, 'protocol_version': protocol_version, 'extra_config': extra_config}
+    )
+
+    if globals().get(clsname):
```
mambocab: I belieeeeeve this could be `if clsname in globals()`.

upgrade_through_versions_test.py
```diff
@@ -973,60 +803,198 @@ def _multidc_schema_create(self):
...
+        raise RuntimeError("Class by name already exists!")
+
+    globals()[clsname] = newcls
```
mambocab: I wrote this:

> I know this is a nit, but could you pull out the `globals` logic into a different function and call that? That separation could be nice in the future. Should be as simple as moving the `globals` logic to an `set_global_if_unique(cls)` function, and then wrapping calls to `create_upgrade_class` in that.

I know I've been harping on it, but after getting back from vacation, I can't think of a possible use case for generating a class without adding it to `globals()`. Generating the class, then attaching it to the current module with uniqueness enforcement, might actually be the atomic unit of what we want to do.

So, those are my thoughts. What I mean is, I'm torn, but I think this implementation is probably good?

upgrade_through_versions_test.py
```diff
@@ -973,60 +803,198 @@ def _multidc_schema_create(self):
...
+# FOR CUSTOM/LOCAL UPGRADE PATH TESTING:
+#    Define UPGRADE_PATH in your env as a comma-separated list of versions
+#    Versions can be any format CCM works with, including: git:cassandra-2.1, 2.1.12, local:somebranch, etc.
+#    Set your desired protocol version in your env's PROTOCOL_VERSION
+if os.environ.get('UPGRADE_PATH'):
+    create_upgrade_class(
+        'UserDefinedUpgradeTest',
+        os.environ.get('UPGRADE_PATH').split(','),
+        bootstrap_test=True,
+        protocol_version=int(os.environ.get('PROTOCOL_VERSION', 3))
```
mambocab: We'll want to make sure we use the same default everywhere.

upgrade_through_versions_test.py
ptnapoleon: sigil --> sentinel?

mambocab: D'oh, yep

upgrade_through_versions_test.py
knifewine: Env var removed in latest commit.

dtest.py
knifewine: Admittedly murky, I felt 'set' might give the impression that it's idempotent, or that it can be run mid-test. That said, I'm fine with 'set' if it feels less confusing.

upgrade_through_versions_test.py
knifewine: ah, much better. Juist committed.

upgrade_through_versions_test.py
knifewine: I personally like how it ended up, and things should be straightforward enough to change if we need it later. That said, I don't feel super strongly either way. It mostly took this shape for simplicity.

upgrade_through_versions_test.py
knifewine: I think we're good here since I removed the other env var entirely, unless I'm misunderstanding.

upgrade_through_versions_test.py
knifewine: This wasn't just for the function default arg mutability gotcha, this is also for types (our generated classes) having mutables which can end up shared between multiple instances of those types. If we later start copying or subclassing generated types it could become significant. Just didn't want to leave extra rope here.

upgrade_through_versions_test.py
knifewine: This is one case where I don't want folks to have to fiddle with PRINT_DEBUG to get the right information. I'm thinking C* devs who want upgrade tests being as sane as possible OOTB.

What if I pull in print_ from six, seem like a reasonable solution?

knifewine: er wait, print_ from six doesn't do what I think it does.

knifewine: Just committed a change to use print_ from six.

upgrade_through_versions_test.py
mambocab: > this is also for types (our generated classes) having mutables which can end up shared between multiple instances of those types.

Sorry, I don't follow -- which mutables wind up being shared if you use the `if param is None:` trick? or do I misunderstand your comment?

upgrade_through_versions_test.py
mambocab: @ptnapoleon There was some discussion of adding meta-tests that can scrape tags from the git repo to figure out the most recent release in a given X.Y version and make sure the `latest_XdotY` is correct, so we could use that to make maintenance a little easier.

upgrade_through_versions_test.py
mambocab: works for me.

dtest.py
mambocab: Maybe `init` conveys what you want?

upgrade_through_versions_test.py
mambocab: I mean the `3` in `protocol_version=int(os.environ.get('PROTOCOL_VERSION', 3))` -- we had a discussion elsewhere about setting a default protocol version, but it looks like you've removed the default there, which I think is correct. If it were still there, we'd want to make sure we had a var that the two defaults shared.

upgrade_through_versions_test.py
```diff
@@ -15,160 +12,52 @@
...
-    Returns just the version string 'X.Y.Z'
-    """
-    if version.find('-') >= 0:
-        return LooseVersion(version.split('-')[1])
-    elif version == 'trunk':
-        return LooseVersion(make_ver_str(TRUNK_VER))
-    else:
-        return LooseVersion(version)
+def switch_jdks(version):
+    cleaned_version = sanitize_version(version)
```
mambocab: This needs to be changed to account for `santize_version`'s error-raising API.

upgrade_through_versions_test.py
```diff
@@ -15,160 +12,52 @@
...
-    Takes versions of the form cassandra-1.2, 2.0.10, or trunk.
-    Returns just the version string 'X.Y.Z'
-    """
-    if version.find('-') >= 0:
-        return LooseVersion(version.split('-')[1])
-    elif version == 'trunk':
-        return LooseVersion(make_ver_str(TRUNK_VER))
-    else:
-        return LooseVersion(version)
+def switch_jdks(version):
```
mambocab: This may be out of scope for this change, but: Some C\* versions work on Javas 7 and 8, right? So, we may want to make this configurable somehow. Maybe this method doesn't need to change, but it's something to think about.

upgrade_through_versions_test.py
knifewine: It's basically for this gotcha:

    >>> class SomeObj(object):
    ...   oops_im_mutable = []
    ... 
    >>> a = SomeObj()
    >>> a.oops_im_mutable.append('foo')
    >>> b = SomeObj()
    >>> b.oops_im_mutable
    ['foo']

dtest.py
knifewine: I like 'init'. Committed.

upgrade_through_versions_test.py
knifewine: right.

upgrade_through_versions_test.py
knifewine: That looks like an out of date diff. switch_jdks will call sanitize_version with the default args of allow_ambiguity=True, which means we'll log the inability to switch but won't let that kill the test with an unhandled exception.

upgrade_through_versions_test.py
knifewine: Yeah, this isn't something I want to tackle with the other changes here. If you want to capture it in a cstar issue let's make it happen.

upgrade_through_versions_test.py
knifewine: Actually, I can make this a lot less dumb by using a shallow copy where I'm concerned about it. Will push a fresh commit in a few.

knifewine: bah, actually a shallow copy would have the same issue. All the same I committed a change to make the default arg more idiomatic. It's still going to have to be a two-tuple though for the reason shown in my example code above.

mambocab: Since it's creating a new object when `create_upgrade_class` is called, I think we're ok whether `extra_config` is mutable or not.

Of course, I just wrapped my head around the fact that default you had before was an immutable tuple of strings, which are themselves immutable. It's a perfectly fine default parameter. If you prefer to move this default back into the args, go ahead, sorry for my confusion.

upgrade_through_versions_test.py
mambocab: Got it, ok?

knifewine:
CI is running decently now: http://cassci.datastax.com/view/Upgrades/job/newer_upgrades_bucketized/

knifewine:
@ptnapoleon @mambocab -- could you take one more quick here, and give a +1 if this looks good to you?

upgrade_through_versions_test.py
mambocab: For future reference: filed internally as CSTAR-202.

upgrade_through_versions_test.py
```diff
@@ -629,29 +498,33 @@ def _log_current_ver(self, current_tag):
...
         self.created_metadata_versions.append((self.cluster.version(), tag))
         session = self.patient_cql_connection(self.node2)
         session.execute('use upgrade')
-        debug("schema metadata establish tables tag: {0}".format(tag))
+        debug("schema metadata establish tables tag: {0}".format(safe_name))
 
         for m in filter(lambda mtd: mtd.startswith('establish_'), dir(schema_metadata_test)):
             debug("schema establish calling: [{0}]".format(m))
-            getattr(schema_metadata_test, m)(self.cluster.version(), session, tag)
+            getattr(schema_metadata_test, m)(self.cluster.version(), session, safe_name)
```
mambocab: Out of scope for this PR, but could you file a ticket to pull out whatever's shared between these tests into some utility module? I don't think we should be importing other test modules.

mambocab:
The code looks great. The CassCI run is full of flaps though: http://cassci.datastax.com/view/Upgrades/job/newer_upgrades_bucketized/10/testReport/upgrade_through_versions_test/

Is this better than/the same as/worse than before these changes? If this isn't a stability regression, then I'm +1.

knifewine:
These don't quite compare apples to apples with the old jobs, but my basic sense is we are in the same place stability-wise as before the refactor (which is to be expected).

knifewine:
in fact, we may be a bit better off (here's the 4 jobs that will be supplanted by the singular new job):
http://cassci.datastax.com/view/Upgrades/job/cassandra_upgrade_proto_v1/lastCompletedBuild/testReport/
http://cassci.datastax.com/view/Upgrades/job/cassandra_upgrade_proto_v2/lastCompletedBuild/testReport/
http://cassci.datastax.com/view/Upgrades/job/cassandra_upgrade_proto_v3/lastCompletedBuild/testReport/
http://cassci.datastax.com/view/Upgrades/job/cassandra_upgrade_proto_v4/lastCompletedBuild/testReport/

knifewine:
so yeah, 80 failures on the old jobs/test code vs 37 on the new job/code. Though we're generating less tests with the newer code so the comparison isn't exactly fair.

mambocab:
\o/

PR #724, opened 2015-12-24
Tests for CASSANDRA-9303

stef1927:
These are the tests for the new COPY options introduced by CASSANDRA-9303. Because of the new parameters of `format_value()` in pylib `formatting.py`, once CASSANDRA-9303 is committed the existing tests will break. For the same reason we should not merge until 9303 is committed, I also omitted `@require`.

There is a bit of code to review so take your time, we can hold the commit of 9303 until this review is completed.

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -14,6 +14,7 @@ from uuid import UUID, uuid4
 
 from cassandra import InvalidRequest
 from cassandra.concurrent import execute_concurrent_with_args
+from cassandra.query import BatchStatement, BatchType
 
 from assertions import assert_all, assert_none
 from ccmlib import common
@@ -1227,6 +1228,9 @@ Tracing session:""")
         - an unlogged batch across multiple partitions should generate a WARNING if at least one partition is
         not local.
 
+        Execute two unlogged batches: one only with local partitions and the other one with non local partitions.
+        Check that only the second one generates a client warning.
+
         @jira_ticket CASSNADRA-9399
         @jira_ticket CASSANDRA-9303
         """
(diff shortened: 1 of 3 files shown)


ptnapoleon:
Sorry about the need to rebase, that's my fault, and I can handle it when we need to merge this. Github isn't letting me comment inline because the diff is too large, and I can't seem to comment on the file directly(?) so I'm going to leave review here. I see Paulo is covering it as well on Jira though, and you are both checking test failures on linux and windows, which is great.

ptnapoleon:
For `test_reading_with_skip_and_max_rows`, the variable names are slightly confusing. We have `max_num_rows`, but `num_rows` is what we're setting as MAXROWS in COPY FROM

ptnapoleon:
Also for `test_reading_with_skip_and_max_rows`, what happens if we set MAXROWS to greater than the number of rows in the file?

ptnapoleon:
In `test_reading_with_skip_cols`, and elsewhere, don't we want `csvfile.close()`, not `csvfile.close`? I think the `with:` is handling it either way, though.

ptnapoleon:
Can you add to the docstring of `test_reading_counters_with_skip_cols` that COPY FROM for counters is not idempotent and should result in continued increases? I know the asserts show that, but it's better to also document it.

ptnapoleon:
In `_test_writing_with_token_boundaries`
where you have
```
# debug(result)
# debug(csv_values)
```

these should either be uncommented or removed.

ptnapoleon:
For `test_boolstyle_round_trip`, what happens if we mix bool styles? Such as  `do_round_trip('TRUE', 'no')` or ` do_round_trip('True', '0')`?

stef1927:
Thanks for the review. The need to rebase is absolutely not your fault and it is taken care of. 

* `test_reading_with_skip_and_max_rows`: I've renamed `max_num_rows` to `num_file_rows` for clarity and I believe the case MAXROWS greater than the number of rows in the file is coverted by `do_test(num_file_rows + 1, 0)` or did you mean something else?

* `csvfile.close()`: in fact we don't even need to call close since it's already called by the `with` context manager.

* `test_reading_counters_with_skip_cols`: OK

* `_test_writing_with_token_boundaries`: OK

* `test_boolstyle_round_trip`: I've added the two new cases, it doesn't matter what the labels are as long as they are different.

ptnapoleon:
`test_reading_with_skip_and_max_rows`, sorry I missed `do_test(num_file_rows + 1, 0)`.

For `test_boolstyle_round_trip` what happens if they are the same? Can you assert the proper exception/error is thrown, and that it doesn't try to process the data?

Otherwise, +1, this is all great, thanks.

stef1927:
I've added some invalid cases to `test_boolstyle_round_trip`, thanks.

I am also working on stabilizing the bulk tests on this branch (CASSANDRA-10938) but hopefully these tests won't change much from now on.

stef1927:
There's a couple more commits to review, thanks.

I was also wondering if you would like on master these changes: https://github.com/stef1927/cassandra-dtest/commit/a57d8ec6e15c7dc7715b12f2b9a4fba048e41770#diff-4cbdcd9896ce5009acd37ed830ce8959R667. I normally use these utility methods to run tests with Java flight recorder.

ptnapoleon:
Does using Java Flight Recorder need anything installed on a machine that wouldn't already be there with a normal jdk installation? If no, go ahead and commit those. If not, can you commit just the utility methods in dtest.py, thanks? I'll review the new commits now.

stef1927:
A normal installation is all that's required for flight recorder. However, I will only commit the utility methods. The idea is that people call them when required, we are not going to record JFR files during ordinary tests.

ptnapoleon:
The new commits look good, +1.

stef1927:
JFR utility methods added, thanks for the review.

I will merge once the ticket is committed.

cqlsh_tests/cqlsh_tests.py
```diff
@@ -1238,10 +1244,38 @@ def test_client_warnings(self):
...
 
-        self.verify_output("USE client_warnings; BEGIN UNLOGGED BATCH INSERT INTO test (id, val) VALUES (1, 'abc') INSERT INTO test (id, val) VALUES (2, 'def') APPLY BATCH",
-                           node1, """
+        session = self.patient_cql_connection(node1)
+        metadata = session.cluster.metadata
+        prepared = session.prepare("INSERT INTO client_warnings.test (id, val) VALUES (?, '?')")
+
+        i = 0
+        vals_by_replicas = defaultdict(list)
+        while len(vals_by_replicas) < 3 or len(vals_by_replicas[node1.address()]) < 2:
```
mambocab: Sorry to swoop in when you thought review was over, but I'm having trouble reading this loop. I think it'd be more readable if you pulled out the `while` guard into a function, so `while not pk_map_covers_all_replicas(vals_by_replicas)`, or something like that. You could also (as per some discussion with @ptnapoleon) pull the first two lines of the loop into a function (you could define it in the test and close over prepared.protocol_version so you don't have to pass that over and over). I think that'd be helpful for maintenance.

cqlsh_tests/cqlsh_tests.py
```diff
@@ -1238,10 +1244,38 @@ def test_client_warnings(self):
...
+            replica = metadata.get_replicas('client_warnings', pk)[0].address
+            debug("{} -> {}".format(i, replica))
+            vals_by_replicas[replica].append(i)
+            i += 1
+
+        prefix = "USE client_warnings; BEGIN UNLOGGED BATCH "
+        suffix = " APPLY BATCH"
+        all_nodes_statments = []
+        node1_statements = []
+        for r, vals in vals_by_replicas.iteritems():
```
mambocab: You can avoid this loop using the Python driver's batch api: https://datastax.github.io/python-driver/api/cassandra/query.html#cassandra.query.BatchStatement

dtest.py
```diff
@@ -664,6 +664,37 @@ def wait_for_any_log(self, nodes, pattern, timeout):
         raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
                            " Unable to find :" + pattern + " in any node log within " + str(timeout) + "s")
 
+    def get_jfr_jvm_args(self):
+        """
+        :return: The JVM arguments required for attaching flight recorder to a Java process.
```
mambocab: Doxygen uses `@return`, not `:return:`.

mambocab:
> On windows we cannot open temporary files after creating them unless we close them first.

Whoa. Terrifying.

But, I have some review for `getTempFile`:

- The name should be `get_temp_file`, not `getTempFile`. Similar comments for `deleteTempFile`.
- Could you note in the docstring that we want to make sure the test object owns a reference to the file object? That's why these values are attached to `self` in the first place, so that the file object isn't deleted when it falls out of scope in the test, which would delete the file.

mambocab:
Other than my comments, the changes look good to me. There's a lot but I think I've gotten most of the value out of a review. I have not run the tests locally, but I'll trust @ptnapoleon and @stef1927. Thanks for the tests!

stef1927:
Thank you for your comment @mambocab, they should have all been addressed in this commit: https://github.com/riptano/cassandra-dtest/commit/b0e501ca7d5952fbf8025d0be55eb41da0eda288.

Only one minor point: I don't think named temporary files are deleted when they go out of scope, rather when they are closed? So I did not update the comment exactly how you suggested, let me know if I am incorrect though.

mambocab:
I'm probably just misremembering a change from using bare `tempfile` to using `self.tempfile` then. I can't find anything about it in the file's history.

Thanks for your changes, feel free to merge when ready.

PR #726, opened 2015-12-25
Disable PerfDisableSharedMem in jvm.options instead of cassandra-env

pauloricardomg:
Tests for CASSANDRA-10494


ptnapoleon:
Won't this break on every version without the jvm options file?

ptnapoleon:
Looking at cassci, it did. I will fix it.

ptnapoleon:
In fact, it looks like this is breaking on trunk as well? Oh I expect that is because 10494 wasn't merged yet. I'm going to push my version gating fix, and trust in the original code that it will still work when 10494 is merged.

PR #729, opened 2016-01-05
case sensitive drop/alter user tests

krummas:


auth_test.py
```diff
@@ -144,6 +144,21 @@ def dropping_nonexistent_user_throws_exception_test(self):
...
+        cassandra.execute("CREATE USER Test WITH PASSWORD '12345'")
+        usercount = len(list(cassandra.execute("LIST USERS")))
+        cassandra.execute("DROP USER Test")
+        self.assertEqual(usercount - 1, len(list(cassandra.execute("LIST USERS"))))
+
+    def alter_user_case_sensitive(self):
+        self.prepare()
+        cassandra = self.get_session(user='cassandra', password='cassandra')
+        cassandra.execute("CREATE USER Test WITH PASSWORD '12345'")
+        cassandra.execute("ALTER USER Test WITH PASSWORD '54321'")
```
ptnapoleon: This test doesn't have any assertions. We didn't even check that logging in as `Test` works with either password.

krummas: I'm sure we test that elsewhere, we get an exception if the user does not exist (this is CASSANDRA-10817)

ptnapoleon:
If you insist, then +1.

PR #732, opened 2016-01-06
Use jmx instead of lsof

carlyeks:
These changes are to help with CASSANDRA-10863 and CASSANDRA-10879

### DIFF
--- a/thrift_hsha_test.py
+++ b/thrift_hsha_test.py
@@ -3,12 +3,11 @@ import os
 import pycassa
 import shlex
 import subprocess
-import sys
 import time
 import unittest
 
 from dtest import Tester, debug, DEFAULT_DIR
-from tools import known_failure
+from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
 
 JNA_PATH = '/usr/share/java/jna.jar'
 ATTACK_JAR = 'lib/cassandra-attack.jar'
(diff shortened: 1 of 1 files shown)


thrift_hsha_test.py
```diff
@@ -72,13 +71,11 @@ def make_connection():
...
-                stdout = subprocess.Popen(["lsof -a -p %s -iTCP -sTCP:CLOSE_WAIT" % node1.pid], stdout=subprocess.PIPE, stderr=subprocess.STDOUT, shell=True).communicate()[0]
-                lines = stdout.splitlines()
-                if len(lines) == 0:
-                    break
+
+            with JolokiaAgent(node1) as jmx:
+                num_clients = jmx.read_attribute(connected_thrift_clients, "Value")
                 time.sleep(1)
-            self.assertEqual(len(lines), 0, "There are non-closed connections: %s" % stdout)
+                self.assertEqual(int(num_clients), 0)
```
mambocab: Could you add an error message to this, please?

mambocab:
Awesome side effect of this change: we may be able to run it on Windows now. The original commit:

https://github.com/carlyeks/cassandra-dtest/commit/e96e5205c5cc4e9ca24f7cd98caeac2897e2b7c5

indicates that the reason it can't be run on Windows is because is uses `lsof`. @ptnapoleon do you see any other reason it wouldn't work on windows? @carlyeks if we think it will run on Windows, could you please remove the `skipIf` on this test?

ptnapoleon:
Looks like it can work on windows to me.

mambocab:
Great, then @carlyeks if you can remove that `skipIf` and add that error message, we're golden. +1.

carlyeks:
I've made the changes, but I don't have access to a Windows machine; can someone test to make sure that this test does work on Windows before merging?

ptnapoleon:
Yes, I'll run it. Also, if you could fix the travis complaint, that would be great.

ptnapoleon:
Test passes on windows locally for me.

PR #734, opened 2016-01-06
Add tests for nodetool get/settimeout commands

thobbs:
Done for CASSANDRA-10953

### DIFF
--- a/nodetool_test.py
+++ b/nodetool_test.py
@@ -73,19 +73,19 @@ class TestNodetool(Tester):
 
         # read all of the timeouts, make sure we get a sane response
         for timeout_type in types:
-            out, err = node.nodetool('gettimeout %s' % timeout_type)
+            out, err = node.nodetool('gettimeout {}'.format(timeout_type))
             self.assertEqual(0, len(err), err)
             debug(out)
             self.assertRegexpMatches(out, r'.* \d+ ms')
 
         # set all of the timeouts to 123
         for timeout_type in types:
-            _, err = node.nodetool('settimeout %s 123' % timeout_type)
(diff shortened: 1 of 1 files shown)


nodetool_test.py
```diff
@@ -57,3 +57,35 @@ def test_correct_dc_rack_in_nodetool_info(self):
...
+        for timeout_type in types:
+            out, err = node.nodetool('gettimeout %s' % timeout_type)
+            self.assertEqual(0, len(err), err)
+            debug(out)
+            self.assertRegexpMatches(out, r'.* \d+ ms')
+
+        # set all of the timeouts to 123
+        for timeout_type in types:
+            _, err = node.nodetool('settimeout %s 123' % timeout_type)
+            self.assertEqual(0, len(err), err)
```
ptnapoleon: This won't work in OSX. Nodetool always prints to stderr there.

nodetool_test.py
```diff
@@ -57,3 +57,35 @@ def test_correct_dc_rack_in_nodetool_info(self):
...
+
+        # verify that they're all reported as 123
+        for timeout_type in types:
+            out, err = node.nodetool('gettimeout %s' % timeout_type)
```
ptnapoleon: For this and the other string formatting, can you use the form `'gettimeout {}'.format(timeout_type)`

nodetool_test.py
thobbs: Hmm, the test above this has the same logic.

nodetool_test.py
thobbs: This should be fixed now.

ptnapoleon:
+1, then. We can deal with it not passing locally on OSX when we deal with the other test. (I thought I had fixed all of these).

PR #737, opened 2016-01-08
Add test for CASSANDRA-8072

beobal:
Verifies that a node which has been decommed and wiped can rejoin and,
by implication, that it can successfully establish gossip connections
with existing seeds.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -342,7 +342,7 @@ class TestBootstrap(Tester):
         session = self.patient_cql_connection(node2)
         self.assertEquals(original_rows, list(session.execute("SELECT * FROM {}".format(stress_table,))))
 
-        # Decommision the new node and wipe its data
+        # Decommission the new node and wipe its data
         node2.decommission()
         node2.stop(wait_other_notice=True)
         self._cleanup(node2)
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -351,6 +351,41 @@ class TestBootstrap(Tester):
         node2.start(wait_other_notice=True)
         node2.watch_log_for("JOINING:", from_mark=mark)
 
+    def decommissioned_wiped_node_can_gossip_to_single_seed_test(self):
+        """
+        @jira_ticket CASSANDRA-8072
+        @jira_ticket CASSANDRA-8422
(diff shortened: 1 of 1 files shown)


bootstrap_test.py
```diff
@@ -351,6 +351,41 @@ def decommissioned_wiped_node_can_join_test(self):
...
+        """
+        cluster = self.cluster
+        cluster.populate(1)
+        cluster.start(wait_for_binary_proto=True)
+
+        # Add a new node, bootstrap=True ensures that it is not a seed
+        node2 = new_node(cluster, bootstrap=True)
+        node2.start(wait_for_binary_proto=True, wait_other_notice=True)
+
+        # Decommision the new node and kill it
```
ptnapoleon: nit: `Decommission`

bootstrap_test.py
```diff
@@ -351,6 +351,41 @@ def decommissioned_wiped_node_can_join_test(self):
...
+
+        commitlog_dir = os.path.join(node2.get_path(), 'commitlogs')
+        debug("Deleting {}".format(commitlog_dir))
+        shutil.rmtree(commitlog_dir)
+
+        # Now start it, it should be allowed to join
+        mark = node2.mark_log()
+        debug("Restarting wiped node2")
+        node2.start(wait_other_notice=False)
+        node2.watch_log_for("JOINING:", from_mark=mark)
```
ptnapoleon: Is this sufficient to assert it has joined appropriately?

ptnapoleon:
Have you checked this fails before 8072, and passes with the fix?

bootstrap_test.py
beobal: Yes, the crux of 8072 is that the initial shadow round of gossip fails in this scenario. This is called from SS::prepareToJoin and unless it completes successfully, the wiped node won't move on to actually join the ring.

bootstrap_test.py
beobal: fixed, thanks

beobal:
Yes. This is actually @stef1927 's test, just slightly reworked to take accommodate the CASSANDRA-6696 changes. Both she and I have verified that it fails without and passes with the patch.

ptnapoleon:
+1

PR #738, opened 2016-01-08
CASSANDRA-10978 - increase wait time for migration tasks in MV bootstrapping tests

jkni:
After CASSANDRA-6696 and CASSANDRA-10421, flushing the schema keyspace is more likely to cause timeouts on waiting for the completion of the migration test callback.

This is particularly a problem in dtests because it hits a pathological case of 9 data directories on the same physical disk.

This PR adds a JVM argument when starting bootstrapping nodes to increase this wait time.

### DIFF
--- a/materialized_views_test.py
+++ b/materialized_views_test.py
@@ -335,11 +335,11 @@ class TestMaterializedViews(Tester):
 
         debug("Bootstrapping new node in another dc")
         node4 = new_node(self.cluster, data_center='dc2')
-        node4.start(wait_other_notice=True, wait_for_binary_proto=True)
+        node4.start(wait_other_notice=True, wait_for_binary_proto=True, jvm_args=["-Dcassandra.migration_task_wait_in_seconds=2"])
 
         debug("Bootstrapping new node in another dc")
         node5 = new_node(self.cluster, remote_debug_port='1414', data_center='dc2')
-        node5.start()
+        node5.start(jvm_args=["-Dcassandra.migration_task_wait_in_seconds=2"])
 
         session2 = self.patient_exclusive_cql_connection(node4)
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I saw the discussion about this. I'm confident that the timeout is not a bug, if you are.

+1

mambocab:
Might be good to pull out the `"-Dcassandra.migration_task_wait_in_seconds=2"`, or at least the `2`, into a separate variable so we can easily increase or decrease it in the future.

carlyeks:
+1 to @mambocab's suggestion; also, if you can add a description as to why we need this and `@jira_ticket` so that we can remember later on why.

jkni:
I've pushed a commit with description of the parameter, @jira_ticket references, and parameterized migration wait.

materialized_views_test.py
```diff
@@ -19,6 +19,13 @@
 from assertions import assert_all, assert_one, assert_invalid, assert_unavailable, assert_none, assert_crc_check_chance_equal
 
 
+# CASSANDRA-10978. Migration wait (in seconds) to use in bootstrapping tests. Needed to handle
+# pathological case of flushing schema keyspace for multiple data directories. See CASSANDRA-6696
+# for multiple data directory changes and CASSANDRA-10421 for compaction logging that must be
+# written.
+migration_wait = 2
```
ptnapoleon: It is a nit, but shouldn't global constants be capitalized, like `MIGRATION_WAIT = 2`? @mambocab, can you confirm?

mambocab: Please: https://www.python.org/dev/peps/pep-0008/#constants

I'm extremely pleased with this well-commented constant though. Pretty awesome.

mambocab:
Address the style nit and I'm +1. Thanks!

jkni:
Pushed with nit addressed. Thanks - Python style stuff is almost always new to me.

ptnapoleon:
+1.

PR #740, opened 2016-01-11
Pcmanus review 8099 upgrade tests

mambocab:
I'm helping get these changes, from @pcmanus, ready for merge. Some known changes needed at time of writing:

- docstrings, including `@jira_ticket` annotations (@pcmanus, may I ask you to do this, since you understand the tests best?)
- s/`%`/`.format` (I can do this if necessary.)

The big part of this review, though, is to consider the relationship of these tests to existing upgrade tests and integrate them if it's appropriate. I don't think it's right to integrate them into the `upgrade_tests` module because they don't test a mixed cluster -- they test 3.0's ability to read old SSTables. @pcmanus can you confirm that I understand that correctly? and @knifewine could you have a look at these to see if they belong in the `upgrade_through_versions` suite, or could reuse some of its code?

I'm not dead-set against leaving these as standalone tests, since they test something fairly specific, but we should at least have a look.


ptnapoleon:
Even if this stays standalone, I think it should be kept in the upgrade submodule, no?

mambocab:
> I think it should be kept in the upgrade submodule, no?

In principle, yes. Right now, though, that's not how the upgrade code is organized. If this winds up reusing code from that module, then we should move it there, but it may reuse from `upgrade_through_versions_test.py`, in which case we should leave it out of the module for now, I think.

pcmanus:
> they test 3.0's ability to read old SSTables

Yes, though really, there is nothing specific to 3.0. Those tests should pass for any valid upgrade to any version. In particular, the file should probably be renamed and the test class too, because they are *not* specific to 8099.

They are also not really that specific. There is 2 main thing we must ensure on upgrade: 1) that you can restart an upgrade node without failure, and 2) that this upgraded node can still talk to older nodes (so you can do rolling upgrade). The existing test in the upgrade submodule focus on 2 (though they do test 1 as a byproduct), while these new tests focus on 1. We can certainly transform those into multi-node tests, that wouldn't hurt, but it's just that having single node test make it easier to debug upgrade problems that are not related to nodes talking to each other.

pcmanus:
Side note: travis seems to complain about something but my reading of the details doesn't make it clear that this is related to this branch, so feel free to clarify if someone understand.

mambocab:
@pcmanus The linter errors are caused by an old problem with our `.travis.yaml`, which was fixed here:

https://github.com/riptano/cassandra-dtest/commit/e3a731d3b050b67661f43748b213dc3489055d24

The errors Travis has found (at time of writing) are definitely not related to your changes, so no worries there. Feel free to rebase onto master (or I can do it) if you want better feedback from Travis.

mambocab:
@knifewine any thoughts on this? I've thought harder about @pcmanus's comments, and it might be good to integrate these into the full upgrade suite, since they apply so generally. Now, at this stage, I don't know whether that would mean putting them in `upgrade_through_versions` or in `upgrade_tests`, but it sounds like a thing.

knifewine:
I'm not certain where these should live long term, but I think for now we just add them to upgrade_tests/ . I think as we make progress on the overall upgrade testing strategy it will hopefully become more clear how to organize the various upgrade tests.

mambocab:
@ptnapoleon/@knifewine, I've done some cleanup; could you give this a review please? I know I still need to `s/%/.format/`, but I don't want to block merge any longer; I can file a followup PR.

@knifewine and I discussed the upgrade strategy here/in a similar PR. In the future, it'll be nice to consolidate upgrade logic into a single helper class of some sort, but for now, this duplication is fine by us.

mambocab:
ping @ptnapoleon / @knifewine

knifewine:
+1 if they're currently passing.

ptnapoleon:
I'm +1 on merging this

mambocab:
Thanks, y'all -- merging.

PR #741, opened 2016-01-11
Hard refresh schema metadata before checking it

ptnapoleon:
@mambocab Review please, and thanks.


mambocab:
+1 on these changes in particular, but this whole test needs some attention, as discussed offline. @ptnapoleon did you file a bug somewhere for the points I brought up about the comments and the order of parts of the test?

ptnapoleon:
Nope! Will do that though.

PR #742, opened 2016-01-11
Use stricter CL on test_type_secondary_indexing

ptnapoleon:
We were running into EC issues with the Write, Read, Assert pattern. As sometimes we would read from a replica w/o the data yet

@knifewine to review, as he and I broke this in the first place.


knifewine:
LGMT. +1 if it runs clean.

PR #743, opened 2016-01-14
Add test to reproduce CASSANDRA-11010

pauloricardomg:


bootstrap_test.py
```diff
@@ -88,6 +88,35 @@ def simple_bootstrap_test(self):
...
+
+        # Bootstraping a new node
+        node3 = new_node(cluster)
+        node3.set_configuration_options(values={'initial_token': tokens[1]})
+        node3.start()
+
+        node3.watch_log_for("Starting listening for CQL clients", timeout=60)
+        session = self.exclusive_cql_connection(node3)
+        rows = session.execute("SELECT bootstrapped FROM system.local WHERE key='local'")
+        assert rows[0][0] == 'COMPLETED', rows[0][0]
```
ptnapoleon: Could you use one of the unittest assert functions here?
`self.assertEqual(rows[0][0], 'COMPLETED')`
or
`self.assertIn('COMPLETED', rows[0][0])`

should work. Do you have a CI run with this test, with the fix?

bootstrap_test.py
```diff
@@ -88,6 +88,35 @@ def simple_bootstrap_test(self):
...
 
+    def simple_bootstrap_test_nodata(self):
+        """
+        @jira_ticket CASSANDRA-11010
+        Test that bootstrap completes if streaming from nodes with no data
+        """
+
+        cluster = self.cluster
+        tokens = cluster.balanced_tokens(2)
+        cluster.set_configuration_options(values={'num_tokens': 1})
```
ptnapoleon: Would it not be better to simply restrict this test to not use vnodes, with the `@no_vnodes` decorator?

bootstrap_test.py
```diff
@@ -88,6 +88,35 @@ def simple_bootstrap_test(self):
...
+
+        debug("[node1, node2] tokens: %r" % (tokens,))
+
+        # Create a single node cluster
+        cluster.populate(2)
```
ptnapoleon: This is a two node cluster, which does not match the comment.

ptnapoleon: I am not sure if the comment, or the populate call is wrong.

ptnapoleon:
This is correctly catching the bug. I've tested locally.

pauloricardomg:
Thanks for the review. Updated branch with suggested changes (the initial_token was a irrelevant leftover from a previous copied test, so I removed it). Also added since annotation.

ptnapoleon:
Why add the since notation? This *should* work on every version yes?

pauloricardomg:
that's right, removed since annotation. thanks!

ptnapoleon:
+1

PR #744, opened 2016-01-14
updage cfg's to remove redundant upgrade excludes

knifewine:
@mshuler to check that I'm sane


mshuler:
lgtm - +1 (untested)

knifewine:
tested with and without bucketizer, same results each time (1097 tests found).

PR #745, opened 2016-01-14
Remove known_failures for 10997

ptnapoleon:
CASSANDRA-10997 was committed. These are no longer failures.
Either @knifewine or @mambocab to review

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -321,8 +321,7 @@ class CqlshCopyTest(Tester):
         # into a bare function if cqlshlib is made easier to interact with.
         return [[self.format_for_csv(v, time_format, float_precision) for v in row] for row in result]
 
-    @known_failure(failure_source='cassandra',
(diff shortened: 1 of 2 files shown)


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -321,8 +321,7 @@ def result_to_csv_rows(self, result, time_format=DEFAULT_TIME_FORMAT, float_prec
         # into a bare function if cqlshlib is made easier to interact with.
         return [[self.format_for_csv(v, time_format, float_precision) for v in row] for row in result]
 
-    @known_failure(failure_source='cassandra',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-10997')
+
```
mambocab: This introduces a style regression; there should only be 1 empty line between methods.

mambocab:
Except for my comment above (which has been fixed), looks good: no remaining `10997`s, `known_failure` removed from imports where appropriate.

mambocab:
so, +1.

PR #747, opened 2016-01-14
increase timeouts

knifewine:
looks like some tests on CI are struggling to meet the original timeout that worked fine locally.


ptnapoleon:
Lgtm, +1

PR #749, opened 2016-01-15
Wait on nodes detecting one another on upgrade (CASSANDRA-11016)

pcmanus:
Some of the tests do a truncate just after a do_upgrade(). As do_upgrade() restart a node but doesn't wait for the other(s) to have detected the restart of that new node, the truncate might actually fail saying some node are down as pointed in CASSANDRA-11016. The fix is trivial, we just wait for other nodes to notice the restart.


ptnapoleon:
+1

PR #750, opened 2016-01-15
Check for nodetool move to finish on topology_test

ptnapoleon:
By not watching the logs, we were trying to move too many nodes at once, which would throw an error. Unfortunately, watching the logs is not enough, as the state jump to normal happens just before the rest of the ring is ready for another move again.

@knifewine to review

### DIFF
--- a/topology_test.py
+++ b/topology_test.py
@@ -63,11 +63,18 @@ class TestTopology(Tester):
         cluster.flush()
 
         # Move nodes to balance the cluster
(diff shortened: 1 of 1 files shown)


knifewine:
Sleep isn't ideal, but lgtm.

Maybe we should consider pushing a watch_log_for_seconds method to ccm at some point.

As long as the sleep won't harm the determinism of the test, +1.

PR #751, opened 2016-01-18
Fix pep8 2015 01 18

lmr:
A few PEP8 fixes necessary for the PEP checks to only ignore E501 errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ script:
   # we want pyflakes to check all files for unused imports only
   # we use flake8 because it allows us to ignore other warnings
   # exclude the thrift directories - they contain auto-generated code
-  - flake8 --ignore=E,W,F811,F812,F821,F822,F823,F831,F841,N8,C9 --exclude=thrift_bindings,cassandra-thrift .
+  - flake8 --ignore=E501,F811,F812,F821,F822,F823,F831,F841,N8,C9 --exclude=thrift_bindings,cassandra-thrift .
   - git remote add riptano git://github.com/riptano/cassandra-dtest.git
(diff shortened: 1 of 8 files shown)


ptnapoleon:
@mambocab I'll let you handle review here, specifically the travis changes. It all LGTM though.

mambocab:
+1 on the PEP8 and `travis.yml` changes. I'm not comfortable with introducing another local linting script, though. Personally I think I'd prefer keeping only the existing shell script, but I never use the local linting script. @aboudreault You wrote and use the local script -- do you have any feelings about it? I strongly want to avoid maintaining 2 linting scripts, and less strongly feel we should avoid introducing a dependency on `inspektor`.

aboudreault:
From my point of view, I don't see any problem modifying/improving our existing linter script but I wouldn't introduce a new one. @lmr @mambocab , I agree with Jim about the new dependency.. do we really need anything else than pep8 and flake8?

lmr:
@mambocab @aboudreault It's really an optional script. If you guys don't like it, I can simply remove that commit from this Pull Request.

lmr:
There, I've removed the script. Thanks.

mambocab:
@lmr Yeah, I think that's appropriate here. Sorry to reject the script, but thank you for the suggestion. I'll check out `inspektor` one of these days, but today is not that day.

Thank you for your help; it's nice that we'll be able to be a stricter with our linting going forward. We appreciate it! Merging.

lmr:
@mambocab no worries, I sent the PR with the commit knowing it'd be controversial. Better to ask for forgiveness than permission, right?

mambocab:
Sure is :)

PR #752, opened 2016-01-19
Be lenient to almost right computed throughput

pcmanus:
Patch to fix CASSANDRA-10842

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -213,7 +213,10 @@ class TestCompaction(Tester):
         avgthroughput = re.match(throughput_pattern, stringline).group(1).strip()
         debug(avgthroughput)
 
-        self.assertGreaterEqual(float(threshold), round(float(avgthroughput)))
+        # The throughput in the log is computed independantly from the throttling and on the output files while
+        # throttling is on the input files, so while that throughput shouldn't be higher than the one set in
+        # principle, a bit of wiggle room is expected
+        self.assertGreaterEqual(float(threshold) + 0.5, float(avgthroughput))
 
     def compaction_strategy_switching_test(self):
         """Ensure that switching strategies does not result in problems.
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -213,7 +213,7 @@ class TestCompaction(Tester):
         avgthroughput = re.match(throughput_pattern, stringline).group(1).strip()
         debug(avgthroughput)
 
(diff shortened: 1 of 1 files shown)


mambocab:
I'd prefer something like `self.assertGreaterEqual(float(threshold) + 0.5, round(float(avgthroughput)))` or `self.assertGreaterEqual(float(threshold) * 1.05, round(float(avgthroughput)))`. I think that expresses the intended leniency better than rounding the second operand. Another downside of rounding is that it's not going to change the effective threshold by a fixed amount.

(For future reference, the proposed change won't always have the intended effect -- sometimes it will round up, making the comparison less lenient. In the future, if you want to truncate, you can use `math.floor` or `int`.)

mambocab:
The errors travis is finding are unrelated to this change.

mambocab:
Looks like the error Travis found was introduced here:

https://github.com/riptano/cassandra-dtest/commit/18f2f684e82c0e807a8068f73d53645991c568a8

@krummas and @pcmanus if you'll enable Travis on your dtest repositories, or exclusively introduce changes through PRs, you can run the checks before merging. I know it's a little regression, but it's something we like to keep clean.

krummas:
@mambocab do we have a guide how to enable travis on dtest repos?

pcmanus:
> (For future reference, the proposed change won't always have the intended effect -- sometimes it will round up, making the comparison less lenient. In the future, if you want to truncate, you can use math.floor or int.)

Except that I didn't wanted to truncate. That it might round up is expected, I want 5.2 to be fine, but 5.8 not necessarily. And if 4.9 is rounded to 5.0, that makes no difference to that comparison. But anyway, I don't really care about the exact logic so changed it.

mambocab:
> Except that I didn't wanted to truncate. 

Fair, I assumed wrong. Thanks for the explanation.

mambocab:
and thanks for changing the logic, I definitely find it more readable.

mambocab:
> do we have a guide how to enable travis on dtest repos?

@krummas It should be as simple as going to https://travis-ci.org/profile while logged in with your GitHub credentials, then enabling the `krummas/cassandra-dtest` repository there. You may want to make sure "Build pushes" is enabled at https://travis-ci.org/krummas/cassandra-dtest/settings. If that doesn't work, feel free to bug me and I'll see if I can find some docs on it.

mambocab:
Ok, I've run the tests locally and it looks good. Thanks, merging!

PR #753, opened 2016-01-20
Add test for GROUP BY queries (CASSANDRA-10707)

blerer:


paging_test.py
```diff
@@ -4,12 +4,13 @@
...
 from cassandra.policies import FallthroughRetryPolicy
-from cassandra.query import SimpleStatement, dict_factory, named_tuple_factory
+from cassandra.query import SimpleStatement, dict_factory, named_tuple_factory, tuple_factory
 
 from assertions import assert_invalid
 from datahelp import create_rows, flatten_into_set, parse_data_into_dicts
 from dtest import Tester, run_scenarios
-from tools import known_failure, since
+from tools import known_failure, rows_to_list, since
+
```
ptnapoleon: Could you remove this extra blank line please

paging_test.py
```diff
@@ -676,6 +677,665 @@ def test_paging_with_in_orderby_and_two_partition_keys(self):
         assert_invalid(session, "select * from paging_test where col_1=1 and col_2 IN (1, 2) order by col_3 desc;", expected=InvalidRequest)
         assert_invalid(session, "select * from paging_test where col_2 IN (1, 2) and col_1=1 order by col_3 desc;", expected=InvalidRequest)
 
+    def group_by_paging_test(self):
+
```
ptnapoleon: Could you add a docstring to this test that looks like
```
"""
@jira_ticket CASSANDRA-10707
"""
```
please?

ptnapoleon:
For all of the asserts you have in these tests, rather than using 
`assert res == [x ,y ,z], res`
or even moreso for `assert res == [x, y, z]`

Could you please use
`self.assertEqual(res, [x,y,z])`?

paging_test.py
```diff
@@ -676,6 +677,665 @@ def test_paging_with_in_orderby_and_two_partition_keys(self):
...
+            res = rows_to_list(
+                session.execute("SELECT DISTINCT a, count(a)FROM test WHERE a IN (1, 2, 4) GROUP BY a LIMIT 2"))
+            assert res == [[1, 1L],
+                           [2, 1L]], res
+
+            res = rows_to_list(session.execute("SELECT DISTINCT a, count(a)FROM test WHERE a IN (1, 2, 4) LIMIT 2"))
+            assert res == [[1, 3L]], res
+
+    def group_by_with_static_columns_paging_test(self):
+
```
ptnapoleon: Could you give this test the same docstring as the other?

paging_test.py
```diff
@@ -676,6 +676,694 @@ def test_paging_with_in_orderby_and_two_partition_keys(self):
...
+        # Makes sure that we have some tombstones
+        session.execute("DELETE FROM test WHERE a = 1 AND b = 3 AND c = 2")
+        session.execute("DELETE FROM test WHERE a = 5")
+
+        for page_size in (2, 3, 4, 5, 7, 10):
+            session.default_fetch_size = page_size
+
+            # Range queries
+            res = rows_to_list(session.execute("SELECT a, b, e, count(b), max(e) FROM test GROUP BY a"))
```
mambocab: The documentation for `GROUP BY`:

https://github.com/apache/cassandra/compare/trunk...blerer:10707-trunk#diff-54538a41709d699d5836a1dfbac8dc63R1148

states the query will produce one value for each group. What determines which row is `SELECT`ed from? As an example, in this test, why does it return values from the `INSERT INTO test (a, b, c, d, e) VALUES (1, 2, 1, 3, 6)` row, and not some other row? I just want to make sure I understand `GROUP BY` for this review.

paging_test.py
```diff
@@ -676,6 +676,694 @@ def test_paging_with_in_orderby_and_two_partition_keys(self):
...
+                                   [1, 4, 12, 2L, 24],
+                                   [2, 2, 6, 1L, 6],
+                                   [2, 4, 12, 1L, 12],
+                                   [4, 8, 24, 1L, 24]])
+
+            res = rows_to_list(session.execute("SELECT a, b, e, count(b), max(e) FROM test"))
+            self.assertEqual(res, [[1, 2, 6, 7L, 24]])
+
+            res = rows_to_list(
+                session.execute("SELECT a, b, e, count(b), max(e) FROM test WHERE b = 2 GROUP BY a, b ALLOW FILTERING"))
```
mambocab: For queries that fail without `ALLOW FILTERING`, could you please add assertions that they raise the expected error without `ALLOW FILTERING`?

paging_test.py
```diff
@@ -676,6 +676,694 @@ def test_paging_with_in_orderby_and_two_partition_keys(self):
...
+            self.assertEqual(res, [[1, 2, 6, 2L, 12],
+                                   [2, 2, 6, 1L, 6]])
+
+            res = rows_to_list(
+                session.execute("SELECT a, b, e, count(b), max(e) FROM test WHERE b = 2 ALLOW FILTERING"))
```
mambocab: Same comment here as for the `ALLOW FILTERING` assertion above.

paging_test.py
```diff
@@ -676,6 +676,694 @@ def test_paging_with_in_orderby_and_two_partition_keys(self):
...
+            res = rows_to_list(session.execute("SELECT DISTINCT a, count(a)FROM test GROUP BY a"))
+            self.assertEqual(res, [[1, 1L],
+                                   [2, 1L],
+                                   [4, 1L]])
+
+            res = rows_to_list(session.execute("SELECT DISTINCT a, count(a)FROM test"))
+            self.assertEqual(res, [[1, 3L]])
+
+            # Range query with DISTINCT and LIMIT
+            res = rows_to_list(session.execute("SELECT DISTINCT a, count(a)FROM test GROUP BY a LIMIT 2"))
```
mambocab: What about when you `GROUP BY` multiple keys?

paging_test.py
blerer: If the SELECT clause contains a column without aggregate function the first value of the column for the group will be returned.

In this case we group on the partion key so for the first row for each group are (1, 4, 2, 12, 24), (2, 2, 3, 3 6) and (4, 8, 2, 12, 24). The values for a, b and e will be taken from these rows

We already support the same behavior for query with aggregates.

mambocab: Great, thanks for confirming.

mambocab:
Existing tests look correct to me. Could you please add tests to address my comments? Then I think we'll be good to merge once the ticket is merged.

mambocab:
Sorry, I mean add _assertions_ to test address my comments.

mambocab:
@blerer I've marked this as `PR Pending C* Ticket Commit`. Can you confirm that that's all we're waiting on for this PR?

blerer:
All the test looks good and CASSANDRA-10707 has been committed in trunk.
If you are fine with the tests we can commit.

ptnapoleon:
Thanks Benjamin!

PR #754, opened 2016-01-20
Modify map keys index upgrade test for 3.0

beobal:
Since 3.0, multiple indexes per-column are supported, so attempting to create a values index on a map which already has a keys index should succeed.


ptnapoleon:
+1, thanks Sam.

PR #756, opened 2016-01-20
Use binary dist for upgrade supercolumns test

mambocab:
Addresses CASSANDRA-10868 and removes dependency on an environment variable.

All 4 tests in this file now fail :/ Do we have a procedure for adding `known_failure` to things before merge? I don't know if I should really file the bug until I have some CassCI runs, but I also don't want the dtest triage person to have to deal with it, since I already know about the failure.

### DIFF
--- a/upgrade_supercolumns_test.py
+++ b/upgrade_supercolumns_test.py
@@ -89,7 +89,7 @@ class TestSCUpgrade(Tester):
         client.transport.close()
 
         if self.cluster.version() >= '2.1':
-            # Upgrade nodes to 2.0.
+            # Upgrade nodes to 2.0 for intermediate sstable conversion
             # See CASSANDRA-7008
             self.upgrade_to_version("binary:2.0.17")
             time.sleep(.5)
--- a/upgrade_supercolumns_test.py
+++ b/upgrade_supercolumns_test.py
@@ -95,7 +95,7 @@ class TestSCUpgrade(Tester):
         if get_version_from_build(CASSANDRA_DIR) >= '2.1':
             # Upgrade nodes to 2.0.
             # See CASSANDRA-7008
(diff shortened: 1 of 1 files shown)


ptnapoleon:
What is the failure? Also, I am.... 63% sure that you can't use `binary:cassandra-2.0` as a version. I think you want `binary:2.0.19`

mambocab:
You're correct about the version, whoops.

As for the errors: I still see some after correcting the version. See here:

https://gist.github.com/mambocab/3f32477ee1c1f80dff0c

Basically, `upgrade_with_counters_test` times out looking for a message about a node coming up, and `upgrade_with_counters_test` fails when the (I believe) upgraded node finds an incompatibile SSTable.

ptnapoleon:
So now I'm sure we want `binary:2.0.17` as we never released a 2.0.18 or 2.0.19. I was thinking of the 1.2 line.

ptnapoleon:
Other than that, yeah, tag it as a known failure and assign it to someone to fix.

mambocab:
@knifewine Could you have a super-quick look at this and let me know if I'm missing something obvious? Here are the failures currently happening on 2.1 when I run this locally:
 
[2.1_failures_out.txt](https://github.com/riptano/cassandra-dtest/files/103461/2.1_failures_out.txt)

Please don't spend much time on this; I'm just asking for a quick once-over before I file a JIRA.

knifewine:
Looks like possibly need to run upgradesstables, but if this is just a
single upgrade from v1 to v2 that seems suspicious.

mambocab:
I've filed [CASSANDRA-11078](https://issues.apache.org/jira/browse/CASSANDRA-11078) to track these tests, which are still failing, and squashed to make history a little prettier. Merging.

PR #758, opened 2016-01-21
Ensure we fix the coordinator for query (CASSANDRA-10905)

beobal:
Use an exclusive cql connection to ensure that the coordinator is deterministic


pcmanus:
+1, merged

PR #760, opened 2016-01-21
clean up the way we iterate over data directories

krummas:
Quite ugly to do the 'for x in xrange(0, self.cluster.data_dir_count)' everywhere, this uses https://github.com/pcmanus/ccm/pull/441 to get the actual data directories instead (note that the ccm-PR needs to be merged first)

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -123,9 +123,9 @@ class TestCompaction(Tester):
         block_on_compaction_log(node1, ks='ks', table='cf')
         time.sleep(1)
         try:
-            for x in xrange(0, cluster.data_dir_count):
-                cfs = os.listdir(node1.get_path() + "/data{0}/ks".format(x))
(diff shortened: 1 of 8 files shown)


compaction_test.py
```diff
@@ -123,14 +123,14 @@ def sstable_deletion_test(self):
...
+            for data_dir in node1.data_directories():
+                cfs = os.listdir(os.path.join(data_dir, "ks"))
+                ssdir = os.listdir(os.path.join(data_dir, "ks", cfs[0]))
                 for afile in ssdir:
                     self.assertFalse("Data" in afile, afile)
 
-        except OSError:
-            self.fail("Path to sstables not valid.")
+        except OSError as e:
+            self.fail("Path to sstables not valid."  + e.strerror)
```
ptnapoleon: You put in extra whitespace before this `+`, and travis is complaining.

ptnapoleon:
CCM PR is merged. I'm +1, after the style nit is fixed. Thanks.

PR #764, opened 2016-01-21
Harden replication tests

ptnapoleon:
@mambocab to review

### DIFF
--- a/replication_test.py
+++ b/replication_test.py
@@ -4,7 +4,7 @@ import re
 import time
 
 from dtest import Tester, debug, PRINT_DEBUG
-from tools import known_failure, no_vnodes, since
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Sorry about the sleep, but nothing I could do about it. There is no programmatic way to check for trace completion effectively.

mambocab:
LGTM. Can you file a PYTHON ticket for the sleep you need?

ptnapoleon:
...No? There is no way to know how long to sleep. So it's not something the driver can add.

mambocab:
buh, sorry -- I meant adding information about when traces were completed. Adam explained over on the PYTHON Jira that that would require a server change and why, though. :+1:

PR #767, opened 2016-01-22
Don't wait other notice after decommission

ptnapoleon:
The decommission will have already caused the node DOWN
notifications to come up. So we won't see more on stop, so waiting
will cause us to timeout

@mambocab to review. Please make your preferred fixes and merge yourself, as I'll be on vacation, and don't want to let this set until I return.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -364,7 +364,7 @@ class TestBootstrap(Tester):
 
         # Decommission the new node and wipe its data
         node4.decommission()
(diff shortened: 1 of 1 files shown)


mambocab:
I'm going to go ahead and merge as-is -- I understand why this would cause problems, but not why the `stop()` call should remain. Seems to be a no-op after a `decommission()`. I'll leave a not for you for when you get back.

PR #770, opened 2016-01-25
Always use LWT read and writes for conditional tests

pcmanus:
This should hopefully fix CASSANDRA-10836

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -3071,12 +3071,6 @@ class TestCQL(UpgradeTester):
             assert_none(cursor, "SELECT * FROM test WHERE k=1", cl=ConsistencyLevel.SERIAL)
             assert_one(cursor, "DELETE FROM test WHERE k=1 IF EXISTS", [False])
 
-            # Using non-conditional update is dodgy here, but it's probably fine given the sleep
-            cursor.execute("UPDATE test USING TTL 1 SET v1=2 WHERE k=1")
-            time.sleep(1.5)
-            assert_one(cursor, "DELETE FROM test WHERE k=1 IF EXISTS", [False])
-            assert_none(cursor, "SELECT * FROM test WHERE k=1", cl=ConsistencyLevel.SERIAL)
-
             assert_one(cursor, "INSERT INTO test (k, v1) VALUES (2, 2) IF NOT EXISTS USING TTL 1", [True])
             time.sleep(1.5)
             assert_one(cursor, "DELETE FROM test WHERE k=2 IF EXISTS", [False])
@@ -3084,7 +3078,7 @@ class TestCQL(UpgradeTester):
 
             assert_one(cursor, "INSERT INTO test (k, v1) VALUES (3, 2) IF NOT EXISTS", [True])
             assert_one(cursor, "DELETE v1 FROM test WHERE k=3 IF EXISTS", [True])
-            assert_one(cursor, "SELECT * FROM test WHERE k=3", [3, None])
+            assert_one(cursor, "SELECT * FROM test WHERE k=3", [3, None], cl=ConsistencyLevel.SERIAL)
             assert_one(cursor, "DELETE v1 FROM test WHERE k=3 IF EXISTS", [True])
             assert_one(cursor, "DELETE FROM test WHERE k=3 IF EXISTS", [True])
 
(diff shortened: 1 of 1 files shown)


upgrade_tests/cql_tests.py
```diff
@@ -3072,28 +3066,29 @@ def conditional_delete_test(self):
...
             assert_one(cursor, "DELETE FROM test WHERE k=1 IF EXISTS", [False])
 
-            cursor.execute("INSERT INTO test (k, v1) VALUES (1, 2)")
+            assert_one(cursor, "INSERT INTO test (k, v1) VALUES (1, 2) IF NOT EXISTS", [True])
             assert_one(cursor, "DELETE FROM test WHERE k=1 IF EXISTS", [True])
-            assert_none(cursor, "SELECT * FROM test WHERE k=1")
+            assert_none(cursor, "SELECT * FROM test WHERE k=1", cl=ConsistencyLevel.SERIAL)
             assert_one(cursor, "DELETE FROM test WHERE k=1 IF EXISTS", [False])
 
+            # Using non-conditional update is dodgy here, but it's probably fine given the sleep
```
mambocab: Any reason not to make this conditional, or use `SERIAL`?

mambocab: If so, could you please expand this comment to add the explanation?

upgrade_tests/cql_tests.py
```diff
@@ -3072,28 +3066,29 @@ def conditional_delete_test(self):
...
+            assert_one(cursor, "INSERT INTO test (k, v1) VALUES (2, 2) IF NOT EXISTS USING TTL 1", [True])
             time.sleep(1.5)
             assert_one(cursor, "DELETE FROM test WHERE k=2 IF EXISTS", [False])
-            assert_none(cursor, "SELECT * FROM test WHERE k=2")
+            assert_none(cursor, "SELECT * FROM test WHERE k=2", cl=ConsistencyLevel.SERIAL)
 
-            cursor.execute("INSERT INTO test (k, v1) VALUES (3, 2)")
+            assert_one(cursor, "INSERT INTO test (k, v1) VALUES (3, 2) IF NOT EXISTS", [True])
             assert_one(cursor, "DELETE v1 FROM test WHERE k=3 IF EXISTS", [True])
             assert_one(cursor, "SELECT * FROM test WHERE k=3", [3, None])
```
mambocab: Should this statement use `SERIAL`?

upgrade_tests/cql_tests.py
```diff
@@ -3786,22 +3778,24 @@ def static_columns_cas_test(self):
...
             # Test that INSERT IF NOT EXISTS concerns only the static column if no clustering nor regular columns
             # is provided, but concerns the CQL3 row targetted by the clustering columns otherwise
-            cursor.execute("INSERT INTO test(id, k, v) VALUES (1, 'foo', 'foo')")
+            assert_one(cursor, "INSERT INTO test(id, k, v) VALUES (1, 'foo', 'foo') IF NOT EXISTS", [True])
             assert_one(cursor, "INSERT INTO test(id, k, version) VALUES (1, 'foo', 1) IF NOT EXISTS", [False, 1, 'foo', None, 'foo'])
             assert_one(cursor, "INSERT INTO test(id, version) VALUES (1, 1) IF NOT EXISTS", [True])
-            assert_one(cursor, "SELECT * FROM test", [1, 'foo', 1, 'foo'])
+            assert_one(cursor, "SELECT * FROM test", [1, 'foo', 1, 'foo'], ConsistencyLevel.SERIAL)
+
+            # Dodgy as its not conditional, but this is not allowed with a condition and that's probably fine in practice so go with it
```
mambocab: As above: any reason not to use `SERIAL` here?

mambocab:
For the batches in `static_columns_cas_test`: I'm assuming that if any statement in a `BATCH` statement is conditional, the entire batch will have the same consistency guarantees as a conditional statement. Is that correct?

upgrade_tests/cql_tests.py
```diff
@@ -4573,14 +4558,14 @@ def map_item_conditional_test(self):
             for frozen in (False, True):
 
                 table = "frozentmap" if frozen else "tmap"
-                cursor.execute("INSERT INTO %s(k, m) VALUES (0, {'foo' : 'bar'})" % (table,))
+                assert_one(cursor, "INSERT INTO %s(k, m) VALUES (0, {'foo' : 'bar'}) IF NOT EXISTS" % (table,), [True])
                 assert_invalid(cursor, "DELETE FROM %s WHERE k=0 IF m[null] = 'foo'" % (table,))
                 assert_one(cursor, "DELETE FROM %s WHERE k=0 IF m['foo'] = 'foo'" % (table,), [False, {'foo': 'bar'}])
                 assert_one(cursor, "DELETE FROM %s WHERE k=0 IF m['foo'] = null" % (table,), [False, {'foo': 'bar'}])
                 assert_one(cursor, "SELECT * FROM %s" % (table,), [0, {'foo': 'bar'}])
```
mambocab: Is not using `SERIAL` here intentional?

mambocab:
The changes look good -- address my comments and I'm +1. Thanks!

upgrade_tests/cql_tests.py
pcmanus: We can't make it SERIAL without switching to an INSERT, in which case that's an exact duplicate of the very next test. But anyway, there is not real reason to try it with an UPDATE so I just removed it in favor of the next lines.

upgrade_tests/cql_tests.py
pcmanus: Not sure to follow that one. Are you talking of the comment or of the line above? Cause the line above does use `SERIAL` and the comment does explain the reason.

pcmanus:
> For the batches in static_columns_cas_test: I'm assuming that if any statement in a BATCH statement is conditional, the entire batch will have the same consistency guarantees as a conditional statement. Is that correct?

Yes

upgrade_tests/cql_tests.py
mambocab: Never mind, I thought you'd be able to use `SERIAL` in the non-conditional `DELETE` below, but I was mistaken.

mambocab:
Great, thanks for the explanation and for addressing my comments. Merging.

PR #772, opened 2016-01-26
Unskip required

mambocab:
@knifewine can you review? These are passing now in the skipped-with-require CassCI job, and the relevant JIRAs were closed.

I also removed the re-import of `debug` in `batch_test.py`. The `debug` defined in `tools.py` is the one that's imported from `dtest`, so they actually point to the same function. Thus, this change won't change the behavior of `debug` in the module.


knifewine:
+1 lgtm

mambocab:
Cheers! merging.

PR #775, opened 2016-01-29
Adds a @no_offheap skip test decorator and applies it to index_summary_upgrade

EnigmaCurry:
@mambocab mentioned it is probably a good candidate for porting this test over to the new upgrade framework, but for now I just disabled this test for offheap runs as I don't see any offheap considerations here, and the upgrade mechanics here are already pretty manual.

### DIFF
--- a/index_summary_upgrade_test.py
+++ b/index_summary_upgrade_test.py
@@ -1,11 +1,13 @@
 from cassandra.concurrent import execute_concurrent_with_args
 
 from dtest import Tester, debug
+from tools import no_offheap
 from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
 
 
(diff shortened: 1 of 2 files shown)


mambocab:
Thought I commented on this, but apparently not. The code itself is fine, but I don't see any reason not to solve this problem by feature-gating the 2.0 part of the test.

mambocab:
Ping @EnigmaCurry

ptnapoleon:
@mambocab Clean this up?

mambocab:
Yeah, I can take this.

mambocab:
Ah, looks like this was addressed in #801. @mshuler can you confirm?

I don't think the fix implemented in this PR is the best solution anyway, so I'l close.

PR #776, opened 2016-02-01
Fixes for CASSANDRA-10938

stef1927:
This introduces a new test to simulate connection failures and test that COPY TO handles them correctly. It should also fix the occasional failures for test_bulk_round_trip_blogposts as described in CASSANDRA-10938.

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -23,7 +23,7 @@ from cqlsh_tools import (DummyColorMap, assert_csvs_items_equal, csv_rows,
                          strip_timezone_if_time_string, unmonkeypatch_driver,
                          write_rows_to_csv)
 from dtest import Tester, canReuseCluster, freshCluster, debug, DISABLE_VNODES
-from tools import known_failure, rows_to_list, since
+from tools import rows_to_list, since
 
 DEFAULT_FLOAT_PRECISION = 5  # magic number copied from cqlsh script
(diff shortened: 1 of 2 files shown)


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -2071,7 +2071,8 @@ def create_records():
         def run_copy_to(filename):
             debug('Exporting to csv file: {}'.format(filename.name))
             start = datetime.datetime.now()
-            copy_to_cmd = "COPY {} TO '{}'".format(stress_table, filename.name)
+            copy_to_cmd = "CONSISTENCY ALL;"
+            copy_to_cmd += "COPY {} TO '{}'".format(stress_table, filename.name)
```
mambocab: Any particular reason for this not to be one of the following?

```python
copy_to_cmd = "CONSISTENCY ALL; COPY {} TO '{}'".format(stress_table, filename.name)
```

or

```python
copy_to_cmd = ("CONSISTENCY ALL; "
               "COPY {} TO '{}'".format(stress_table, filename.name))
```

I find these a little more readable than using string concatenation.

mambocab:
Other than my style comment, this looks good. Could you address that? Then feel free to merge. Thanks!

stef1927:
Style comment addressed, thanks. Let's merge after CASSANDRA-10938 is committed, otherwise `test_bulk_round_trip_blogposts_with_max_connections` will fail.

PR #777, opened 2016-02-01
Add test of reading compact tables during upgrade

beobal:
CASSANDRA-11087

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -2666,6 +2666,54 @@ class TestCQL(UpgradeTester):
             res = cursor.execute("SELECT * FROM bar")
             assert rows_to_list(res) == [[1, 2]], res
 
+    def query_compact_tables_during_upgrade_test(self):
+        """
+        Check that un-upgraded sstables for compact storage tables
+        can be read after an upgrade. Checks for a regression where
+        when the coordinator is on < 3.0, a replica at >= 3.0 returns
+        0 results for any read request. When the >= 3.0 node is
+        the coordinator, the problem does not manifest. Likewise, if
+        the data is inserted after the replica is upgraded, or if
+        upgradesstables is run after upgrade, the query succeeds, so
+        the issue is with reading legacy format sstables in response to
+        a legacy format read request
+        @jira_ticket CASSANDRA-11087
+        """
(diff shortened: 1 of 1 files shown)


upgrade_tests/cql_tests.py
```diff
@@ -2666,6 +2666,54 @@ def compact_metadata_test(self):
...
+            debug("Querying for individual keys retrieved {c} results".format(c=read_count))
+            assert read_count == 100
+            # now a range slice, again all 100 rows should be retrievable
+            res = cursor.execute("SELECT * FROM t1")
+            read_count = len(rows_to_list(res))
+            debug("Range request retrieved {c} rows".format(c=read_count))
+            assert read_count == 100
+
+        for is_upgraded, cursor in self.do_upgrade(cursor):
+            debug("Querying %s node" % ("upgraded" if is_upgraded else "old",))
```
mambocab: I'm sure this is pasted from elsewhere, but could you please use `.format` here instead of `%`?

upgrade_tests/cql_tests.py
```diff
@@ -2666,6 +2666,54 @@ def compact_metadata_test(self):
...
+        self.cluster.flush()
+
+        def check_read_all(cursor):
+            read_count = 0
+            # first read each row separately - obviously, we should be able to retrieve all 100
+            for i in xrange(100):
+                res = cursor.execute("SELECT * FROM t1 WHERE a = {a}".format(a=i))
+                read_count += len(rows_to_list(res))
+            debug("Querying for individual keys retrieved {c} results".format(c=read_count))
+            assert read_count == 100
```
mambocab: Could you please use `self.assertEqual` here, and elsewhere, rather than bare assertions?

mambocab:
I have a slight preference for using `excute_concurrent_with_args` over loops over `execute` statements, but not enough to gate merging on it -- if you agree that that'd be nicer, go ahead and change that. Either way, address my style nits and squash into a single commit, then go ahead and merge.

beobal:
Thanks @mambocab, I've made those changes, squashed & committed in 43abcfcbf433c161ebc553eaf06236919c826904

mambocab:
Awesome, thanks for the test!

PR #779, opened 2016-02-01
version name simple sanitize

knifewine:
fix for dynamic cql object names not working due to ':' that snuck in from ccm ver specifier.

this is cherry picked from my 21_22_tentatives branch because that branch isn't intented to every get merged. code is running on that other branch here http://cassci.datastax.com/view/Upgrades/job/upgrade_through_versions_all/30/

### DIFF
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -501,8 +501,13 @@ class UpgradeTester(Tester):
             "Current upgrade path: {}".format(
                 vers[:curr_index] + ['***' + current_tag + '***'] + vers[curr_index + 1:]))
 
+    def _safe_name(self, tag):
+        # add prefix letter so we don't break C* naming rules in the case of bare versions
+        # and make sure no ccm ':' delimiters sneak in
+        return "t{tag}".format(tag).replace(":", "_")
+
     def _create_metadata_schemas(self, tag):
-        safe_name = "t" + tag  # add a letter so we don't break C* naming rules in the case of bare versions
+        safe_name = self._safe_name(tag)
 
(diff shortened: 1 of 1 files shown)


upgrade_through_versions_test.py
```diff
@@ -501,8 +501,13 @@ def _log_current_ver(self, current_tag):
             "Current upgrade path: {}".format(
                 vers[:curr_index] + ['***' + current_tag + '***'] + vers[curr_index + 1:]))
 
+    def _safe_name(self, tag):
+        # add prefix letter so we don't break C* naming rules in the case of bare versions
+        # and make sure no ccm ':' delimiters sneak in
+        return "t{tag}".format(tag).replace(":", "_")
```
mambocab: I personally prefer `.replace(':', 'COLON')`, an idea I cribbed from `oh-my-zsh`, since underscores are reasonably common in, e.g., dev branch names.

mambocab:
Looks good. The comment is a nit, though I do think `s/_/COLON/` there would make collisions less likely. Do what thou wilt with my feedback, squash into a single commit, and merge away AFIAC.

knifewine:
keeping at '_' rather than 'COLON' just to avoid potential issues with long cql object names (may not be a problem in actuality however).

mambocab:
Sounds good. If it's easy, might be good to add some collision detection. If it's not easy, go ahead and merge; collisions seem unlikely anyway, if I understand this code correctly.

knifewine:
I got this wrong, doesn't quite fix the issue. closing for now to bring a better solution.

PR #780, opened 2016-02-01
Add test for 2i queries with vnodes (CASSANDRA-11104)

beobal:


mambocab:
Looks great and passes where I expect, thanks for the test. This seems to be just a smoke test more or less -- do we need more comprehensive tests than that?

mambocab:
But, go ahead and squash and merge!

beobal:
It kind of is a smoke test, but prior to 11104 being committed it would fail because the keys searcher on the compact table would duplicate results for each of the upgraded nodes vnodes. Like so:
```======================================================================
FAIL: test_query_indexes_with_vnodes (secondary_indexes_test.TestSecondaryIndexes)
  File "/home/sam/cassandra/dtests/trunk/cassandra-dtest/secondary_indexes_test.py", line 440, in test_query_indexes_with_vnodes
    self.assertEqual(len(rows_to_list(res)), 50)
AssertionError: 5074 != 50
```

beobal:
Merged in 20f27985dbb7a06459bf519eae64517cd16dc25d

mambocab:
Yep, just wondering if we needed more than a simple regression test. If that's all we care about right now, then we're good to go. (Regardless, having this test is good, just wondering if we needed to write any more tests.)

beobal:
I don't think we need anything more, the root cause was simply that the filtering by range was overlooked when KeysSearcher was re-implemented for 3.0. It's not upgrade related as such or anything more complex like that.

mambocab:
:+1: thanks!

PR #781, opened 2016-02-02
Make test of number work with negative numbers

pcmanus:
Pretty sure this should fix https://issues.apache.org/jira/browse/CASSANDRA-11058

### DIFF
--- a/largecolumn_test.py
+++ b/largecolumn_test.py
@@ -1,5 +1,5 @@
 from dtest import Tester, debug
-from tools import known_failure, since
+from tools import since
 
 
 @since('2.2')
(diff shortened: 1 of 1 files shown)


mambocab:
If the argument to `is_number` is a string representing a float, this will error out:

```python
❯ python
>>> int('2.1')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '2.1'
```

So, do you expect the input to always be represent an integer? If not, use `float` instead of `int`. If you need to address that, please do, then go ahead and squash and merge. Looks great, thank you!

pcmanus:
> do you expect the input to always be represent an integer?

Yes, that's currently what gcstats will give us, but I guess there is no harm in making it work if that ever change. I'll make the change and merge.

mambocab:
Great, thank you!

PR #782, opened 2016-02-02
Ignore Failed to apply mutation locally errors

carlyeks:
For CASSANDRA-10779; since we are already checking the values that we are writing is consistent, we can ignore a failed mutation locally, so long as the end values check out.


ptnapoleon:
Why is this being logged at ERROR if it's okay?

carlyeks:
It still is an error, but we are checking different conditions to make sure that the mutations are all applied correctly such that we can ignore this error when it comes up.

pcmanus:
I disagree here, the test don't seem to do anything particularly weird that would justify an ERROR log server side. If relatively normal operation can log at ERROR, that's a good sign we might flood the log in production. See my comment on the C* ticket too.

carlyeks:
Fixed with last commit for 10779, so this PR isn't necessary.

PR #785, opened 2016-02-02
consistent hash for naming db objects

knifewine:
fixes issues of db object names being too long or not compatible.

idea and most of implementation thanks to @mambocab!

### DIFF
--- a/schema_metadata_test.py
+++ b/schema_metadata_test.py
@@ -1,8 +1,8 @@
-import re
-
-from nose.tools import assert_equal, assert_in
+from collections import defaultdict
+from uuid import uuid4
 
 from dtest import Tester, debug
+from nose.tools import assert_equal, assert_in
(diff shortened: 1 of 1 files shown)


schema_metadata_test.py
```diff
@@ -481,11 +481,17 @@ def verify_basic_datatype_table(created_on_version, current_version, keyspace, s
         assert_equal('tinyint', meta.columns['s'].cql_type)
 
 
+# TODO: rename this function since it generates names for not just tables
```
ptnapoleon: Why not rename it here as well?

mambocab:
Simple enough, +1. I guess I'm not super gung-ho about leaving TODOs around, but file an internal ticket and I'm ok with it, personally.

knifewine:
you guys are right, todo removed and function renamed.

PR #786, opened 2016-02-03
Fix version for CASSANDRA-10428

stef1927:
When tests for CASSANDRA-10428 were written, trunk was still at 3.2 but now it is at 3.4 and therefore the version tags need updating. At the moment 3 cqlsh tests are failing on 3.3.


stef1927:
I also fixed a time format problem that was causing a cqlsh test to fail on 2.1.

ptnapoleon:
+1

PR #784, opened 2016-02-02
Remove upgrade tests env vars

mambocab:
@ptnapoleon and @knifewine to review.

Please review carefully to make sure the changes are correct. Correctness here means that the changed code executes as though `QUERY_UPGRADED=true`, `QUERY_OLD=true`, and `UPGRADE_MODE=normal`.

PR #787, opened 2016-02-03
Generate upgrade topology classes

mambocab:
This depends on #784 (perhaps unnecessarily?), so wait until that's merged. This consolidates class generation for the `upgrade_tests/` test suite, which will make it easier to add further test-generation logic to run the tests over upgraded and mixed-version clusters.


mambocab:
> so wait until that's merged

That is, hold _merge_ on #784. Would love review in advance of it. Looking for feedback from @knifewine, optionally @ptnapoleon.

upgrade_tests/upgrade_base.py
```diff
@@ -33,6 +23,52 @@
 UpgradePath = namedtuple('UpgradePath', ('starting_version', 'upgrade_version'))
 
 
+trunk_version = '3.4'
```
ptnapoleon: What do we think about these all being CAPITALIZED, since they're string constants?

ptnapoleon:
So far I love it.

upgrade_tests/upgrade_base.py
mambocab: I do, but for a separate PR. This is already big enough IMO.

mambocab: Language is hard. Yes, these should be capitalized, but not in this PR.

mambocab:
I forgot to mention: it's important that I successfully removed all the existing classes at the leaves of the inheritance tree and replaced them with automatically generated classes. Please review that there are no leaves left.

ptnapoleon:
I found none

mambocab:
IMPORTANT: Don't merge this until we can change some CassCI jobs 



and babysit them to accommodate having moved `upgrade_through_versions.py`.

mambocab:
About to push over https://github.com/mambocab/cassandra-dtest/commit/6ec105625c7453ec86c06f966212f1ed13eea3a1 with a rebased version.

upgrade_tests/paging_test.py
```diff
@@ -1752,9 +1712,20 @@ def test_failure_threshold_deletions(self):
             self.assertTrue(failed, "Cannot find tombstone failure threshold error in log for {} node".format(("upgraded" if is_upgraded else "old")))
 
 
-class TestPagingWithDeletionsNodes3RF3(TestPagingWithDeletions):
-    NODES, RF, __test__, CL = 3, 3, True, CL.ALL
-
-
-class TestPagingWithDeletionsNodes2RF1(TestPagingWithDeletions):
-    NODES, RF, __test__ = 2, 1, True
+for klaus in BasePagingTester.__subclasses__():
```
knifewine: awesome, this is a really nice way to get additional tests for "free"

knifewine: might be useful later for classes to be able to opt-out of class-gen upgrade testing -- probably yagni for now though. can implement it if we need it.

mambocab: Yeah! Haven't used `abc` for anything particularly fancy, but this is a really useful use case.

And for excluding classes: if nothing else, we can do stuff post-test-class-generation, e.g. `TestClassWeDontWantToRun.__test__ = False` or  `TestClass.test_we_dont_want_to_run = unittest.skip(TestClass.test_we_dont_want_to_run)`.

I guess there's a bigger code-style debate -- should "don't run this test under condition X" logic live in in the test-class-generation code, or should we, e.g., create decorators to annotate tests? I lean toward the latter, even though it involves a little more machinery, because it makes the code document the test as well as skip it.

knifewine: Yeah, I think decorators would eventually be nice for that. We're going to start needing more meta tests as our decorator config possibilities expand (esp wrt combination testing).

knifewine:
+1, looks good to me.

upgrade_tests/upgrade_base.py
```diff
@@ -19,6 +23,73 @@
...
 
 
+# these should be latest tentative tags, falling back to the most recent release if no pending tentative
+latest_2dot0 = '2.0.17'
+latest_2dot1 = '2.1.13'
+latest_2dot2 = '2.2.5'
+latest_3dot0 = 'git:3.0.3-tentative'
+latest_3dot1 = '3.1.1'
+latest_3dot2 = '3.2.1'
+latest_3dot3 = 'git:3.3-tentative'
```
ptnapoleon: 3.3 and 3.0.3 are no longer tentative.

upgrade_tests/upgrade_base.py
```diff
@@ -19,6 +23,73 @@
...
+head_2dot2 = 'git:cassandra-2.2'
+head_3dot0 = 'git:cassandra-3.0'
+head_3dot1 = 'git:cassandra-3.1'
+head_3dot2 = 'git:cassandra-3.2'
+head_3dot3 = 'git:cassandra-3.3'
+head_trunk = 'git:trunk'
+
+
+VALID_UPGRADE_PAIRS = (
+    # UpgradePath(name='2_0_UpTo_2_1_HEAD', starting_version=latest_2dot0, upgrade_version='git:cassandra-2.1'),
```
ptnapoleon: Why commented out?

ptnapoleon:
Left two notes, otherwise +1

upgrade_tests/upgrade_base.py
knifewine: ah, good point @ptnapoleon

upgrade_tests/upgrade_base.py
mambocab: Because we don't yet have a way to specify Java versions from within the tests, and we need to either use Java 7 or switch from 7 to 8 mid-upgrade. At least, I think that's right. @knifewine?

upgrade_tests/upgrade_base.py
mambocab: Thanks!

mambocab: Pushed a change.

upgrade_tests/upgrade_base.py
knifewine: I'm not certain on that one.

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -2,76 +2,26 @@
...
-        else:
-            os.environ['JAVA_HOME'] = os.environ['JAVA8_HOME']
-    except KeyError:
-        raise RuntimeError("You need to set JAVA7_HOME and JAVA8_HOME to run these tests!")
-    debug("Set JAVA_HOME: [{}] for cassandra version: [{}]".format(os.environ['JAVA_HOME'], version))
+from upgrade_base import (head_2dot1, head_2dot2, head_3dot0, head_3dot1,
+                          head_3dot2, head_3dot3, head_trunk, latest_2dot0,
+                          latest_2dot1, latest_2dot2, latest_3dot0,
+                          latest_3dot1, latest_3dot2, latest_3dot3,
+                          switch_jdks)
```
knifewine: a nice-to-have here would be something we can import without calling out all the names individually (and not using the evil import *). not sure how we'd structure it atm, don't see it as necessary for this pr.

knifewine:
+1

upgrade_tests/upgrade_through_versions_test.py
mambocab: I agree on all counts. Once we figure out how we're specifying the upgrade graph, this should be as simple as importing that data structure and using its attributes/values.

knifewine: that's even better than what I was picturing.

knifewine:
+1 on the latest commits.

mambocab:
Ok, cool. Thanks for all the help and feedback, y'all. Merging.

PR #789, opened 2016-02-04
Add test for CASSANDRA-10176

pcmanus:


nodetool_test.py
```diff
@@ -91,3 +91,40 @@ def test_nodetool_timeout_commands(self):
...
+        cluster = self.cluster
+        cluster.populate([3]).start()
+
+        node = cluster.nodelist()[0]
+
+        notice_message = r'effective ownership information is meaningless'
+
+        # Do a first try without any keypace, we shouldn't have the notice
+        out, err = node.nodetool('status')
+        self.assertEqual(0, len(err), err)
```
mambocab: If I understand correctly, this will fail on OS X -- there's a bug on the platform that causes `java` to always print something to `stderr`. @ptnapoleon can you confirm?

nodetool_test.py
```diff
@@ -91,3 +91,40 @@ def test_nodetool_timeout_commands(self):
...
+        self.assertNotRegexpMatches(out, notice_message)
+
+        session = self.patient_cql_connection(node)
+        session.execute("CREATE KEYSPACE ks1 WITH replication = { 'class':'SimpleStrategy', 'replication_factor':1}")
+
+        # With 1 keyspace, we should still not get the notice
+        out, err = node.nodetool('status')
+        self.assertEqual(0, len(err), err)
```
mambocab: Same as above.

nodetool_test.py
```diff
@@ -91,3 +91,40 @@ def test_nodetool_timeout_commands(self):
...
+        self.assertNotRegexpMatches(out, notice_message)
+
+        session.execute("CREATE KEYSPACE ks2 WITH replication = { 'class':'SimpleStrategy', 'replication_factor':1}")
+
+        # With 2 keyspaces, we should now get the notice
+        out, err = node.nodetool('status')
+        self.assertEqual(0, len(err), err)
```
mambocab: Same as above.

mambocab:
Other than my comment about OS X, this looks good -- locally on Linux, it passes on your patched versions, and fails on unpatched trunk. I guess await instructions from @ptnapoleon about how to address my OS X comment, then feel free to merge. Thanks!

nodetool_test.py
ptnapoleon: Yes, but don't worry about it. I have a ticket to fix all instances of this, of which there are many.

mambocab:
Ok, looks like this is fine, then. Go ahead and merge when ready @pcmanus, thanks!

pcmanus:
Ok, I'll commit once CASSANDRA-10176 goes in.

mambocab:
Great, thank you!

PR #790, opened 2016-02-04
Block on traces in replication_test.py

ptnapoleon:
@mambocab to review

### DIFF
--- a/replication_test.py
+++ b/replication_test.py
@@ -41,14 +41,48 @@ murmur3_hashes = {
 }
 
 
+def query_system_traces_length(session):
+    return len(list(session.execute("SELECT * FROM system_traces.events")))
+
+
+def last_n_values_same(n, iterable):
+    last_n_values = iterable[-n:]
+    if len(last_n_values) != n:
+        return False
+    num_unique_values_in_last_n = len(set(last_n_values))
+    return num_unique_values_in_last_n == 1
+
+
+def block_on_trace(session):
+    results_from_query = []
+    num_same_results_required = 5
+
+    # We should never run into a timeout, because
+    # eventually trace events should stop being generated.
+    # Just in case though, we add a large timeout, to prevent
+    # deadlock.
+    start = time.time()
+    timeout = start + 180
+
+    while not last_n_values_same(num_same_results_required, results_from_query):
+        results_from_query.append(query_system_traces_length(session))
+        time.sleep(1)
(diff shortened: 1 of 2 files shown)


replication_test.py
```diff
@@ -150,15 +155,38 @@ def get_replicas_for_token(self, token, replication_factor,
...
+        last_n_values = iterable[-n:]
+        if len(last_n_values) != n:
+            return False
+        num_unique_values_in_last_n = len(set(last_n_values))
+        return num_unique_values_in_last_n == 1
+
+    def block_on_trace(self, session, future):
+        results_from_query = []
+        num_same_results_required = 3
+        while (self.last_n_values_same(num_same_results_required, results_from_query)):
```
mambocab: Unwrap the condition here.

replication_test.py
```diff
@@ -150,15 +155,38 @@ def get_replicas_for_token(self, token, replication_factor,
...
+            results_from_query.append(query_system_traces_length())
```
mambocab: You'll probably want to add a `sleep` here in the loop, and add a `timeout` parameter to this whole method.

replication_test.py
```diff
@@ -150,15 +155,38 @@ def get_replicas_for_token(self, token, replication_factor,
...
+
+        trace = future.get_query_trace(max_wait=120)
+        return trace
```
mambocab: Don't give `block_on_trace` a return value!

replication_test.py
```diff
@@ -150,15 +155,41 @@ def get_replicas_for_token(self, token, replication_factor,
...
+        results_from_query = []
+        num_same_results_required = 3
+        start = time.time()
+
+        while self.last_n_values_same(num_same_results_required, results_from_query) and time.time() < start + 15:
+            results_from_query.append(query_system_traces_length())
+            time.sleep(1)
+
+        trace = future.get_query_trace(max_wait=120)
+        return trace
```
mambocab: I think you should get rid of this return value and just have the use be:

```
block_on_trace(session)
trace = future.get_query_trace(max_wait=120)
```

This allows callers to configure the call to `future.get_query_trace` without adding extra parameters to `block_on_trace`, and allows callers to call `block_on_trace` without creating a future.

Also, all of these can be functions -- move them off the class, please.

mambocab:
noice, +1

PR #791, opened 2016-02-04
Merge auth_roles_test.py into auth_test.py

ptnapoleon:
@mambocab to review


auth_test.py
```diff
@@ -660,6 +662,1230 @@ def data_resource_creator_permissions(creator, resource):
             permissions.append((creator, '<all functions in %s>' % keyspace, perm))
     return permissions
 
+# Second value is superuser status
+# Third value is login status, See #7653 for explanation.
+mike_role = ['mike', False, True, {}]
+role1_role = ['role1', False, False, {}]
+role2_role = ['role2', False, False, {}]
+cassandra_role = ['cassandra', True, True, {}]
```
mambocab: aaaaahhhhhhhhhh

mambocab: In particular, these should be `namedtuple`s

mambocab:
In principle, this merge is fine for now, but at some point we should have a more detailed look and separate these out into submodules in a single folder.

mambocab:
Execution looks good too. I'm +1 on the changes; please either make those lists `namedtuple`s or file a JIRA for us to do that in this sprint.

ptnapoleon:
Filed and assigned to myself.

PR #793, opened 2016-02-05
Unrequire metadata test

mambocab:
As discussed privately: `@require`ing this test is incorrect -- we want to skip it everywhere, all the time. Its hanging is hanging our `@require`d-test-auditing job, which we don't want.

Also a couple cleanup commits in there to deal with stuff that `flake8` found.


ptnapoleon:
+1

mambocab:
Cheers. Merging.

PR #794, opened 2016-02-05
Start cluster in replication test one node at a time

ptnapoleon:
Seeing an issue on cassci where with jdk7, nodes all started
at once could have the same vnodes set. Starting the cluster
in serial should prevent this.

@mambocab or @jkni to review


mambocab:
-1. Since the underlying bug is a JDK7 bug can affect any call to `ccmlib.Cluster.cluster.start`, it should be dealt with there.

PR #795, opened 2016-02-05
New upgrade classes

knifewine:
we need to pay closer attention to both head of branch testing, as well as testing of provisional releases as an upgrade's terminating version.

these changes introduce vars to track both the head of various branches, and also others to track 'latest' versions available. in specific release testing we'll be using tags assigned to these latest vars. to reduce unnecessary testing and noise, these are currently gated on whether or not the 'latest' appears to be a provisional unreleased version (currently simply if it has 'tentative' in the name we include it for testing).

this also adds some missing cases for testing 3.3 (along with the new conditional classes that correspond).

### DIFF
--- a/upgrade_through_versions_test.py
+++ b/upgrade_through_versions_test.py
@@ -20,15 +20,23 @@ from cassandra.query import SimpleStatement
 from dtest import Tester, debug
 from tools import generate_ssl_stores, known_failure, new_node
 
-trunk_version = '3.4'
+# these should be latest tentative tags, falling back to the most recent release if no pending tentative
 latest_2dot0 = '2.0.17'
 latest_2dot1 = '2.1.12'
 latest_2dot2 = '2.2.4'
-latest_3dot0 = '3.0.2'
+latest_3dot0 = 'git:tentative-3.0.3'
 latest_3dot1 = '3.1.1'
 latest_3dot2 = '3.2.1'
-latest_3dot3 = '3.3'
-trunk_ccm_string = 'git:trunk'
+latest_3dot3 = 'git:tentative-3.3'
+
+head_2dot0 = 'git:cassandra-2.0'
+head_2dot1 = 'git:cassandra-2.1'
+head_2dot2 = 'git:cassandra-2.2'
+head_3dot0 = 'git:cassandra-3.0'
+head_3dot1 = 'git:cassandra-3.1'
+head_3dot2 = 'git:cassandra-3.2'
+head_3dot3 = 'git:cassandra-3.3'
+head_trunk = 'git:trunk'
 
 
 def sanitize_version(version, allow_ambiguous=True):
@@ -39,7 +47,7 @@ def sanitize_version(version, allow_ambiguous=True):
     If allow_ambiguous is False, will raise RuntimeError if no version is found.
     """
     if (version == 'git:trunk') or (version == 'trunk'):
-        return LooseVersion(trunk_ccm_string)
+        return LooseVersion(head_trunk)
 
     match = re.match('^.*(\d+\.+\d+\.*\d*).*$', unicode(version))
     if match:
@@ -856,24 +864,24 @@ def create_upgrade_class(clsname, version_list, protocol_version,
 # Proto v1 upgrade classes (v1 supported on 2.0, 2.1, 2.2)
 create_upgrade_class(
     'ProtoV1Upgrade_2_0_UpTo_2_1',
-    [latest_2dot0, 'git:cassandra-2.1'],
+    [latest_2dot0, head_2dot1],
     bootstrap_test=True,
     protocol_version=1
 )
 create_upgrade_class(
     'ProtoV1Upgrade_2_1_UpTo_2_2',
-    [latest_2dot1, 'git:cassandra-2.2'],
+    [latest_2dot1, head_2dot2],
     bootstrap_test=True,
     protocol_version=1
 )
 create_upgrade_class(
     'ProtoV1Upgrade_AllVersions',
-    [latest_2dot0, latest_2dot1, 'git:cassandra-2.2'],
+    [latest_2dot0, latest_2dot1, head_2dot2],
     protocol_version=1
 )
 create_upgrade_class(
     'ProtoV1Upgrade_AllVersions_RandomPartitioner',
-    [latest_2dot0, latest_2dot1, 'git:cassandra-2.2'],
+    [latest_2dot0, latest_2dot1, head_2dot2],
     protocol_version=1,
     extra_config=(
         ('partitioner', 'org.apache.cassandra.dht.RandomPartitioner'),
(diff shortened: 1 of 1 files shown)


knifewine:
@mambocab and @ptnapoleon for review, and to make sure I didn't make silly mistakes.

ptnapoleon:
I feel like there is a lot of duplication here. Jim will probably have both opinions and ideas on that. The thought of constantly updating tentative versions is distasteful, but I lack an alternative to suggest. I don't see any mistakes, but will give a closer reading on Monday.

upgrade_through_versions_test.py
```diff
@@ -20,15 +20,23 @@
 from dtest import Tester, debug
 from tools import generate_ssl_stores, known_failure, new_node
 
-trunk_version = '3.4'
+# these should be latest tentative tags, falling back to the most recent release if no pending tentative
 latest_2dot0 = '2.0.17'
 latest_2dot1 = '2.1.12'
 latest_2dot2 = '2.2.4'
-latest_3dot0 = '3.0.2'
+latest_3dot0 = 'git:3.0.3-tentative'
```
mambocab: Is this really the latest 3.0? My initial reaction is that the latest would be the latest release, and that this is... 3.0 `HEAD`? I think the distinction may be necessary so that we can test both "`HEAD` on 2.2 can correctly upgrade to the most recent 3.0 release" and "`HEAD` on 3.0 can be correctly upgraded from the latest 2.2 release".

upgrade_through_versions_test.py
```diff
@@ -853,119 +861,235 @@ def create_upgrade_class(clsname, version_list, protocol_version,
     return newcls
 
 
+def is_unreleased(version_or_tag):
+    """
+    For gating unnecessary testing of already released upgrade tests endpoints.
+    """
+    if 'tentative' in version_or_tag:
```
mambocab: Could be more succinct as `return 'tentative' in version_or_tag`.

upgrade_through_versions_test.py
```diff
@@ -20,15 +20,23 @@
...
+latest_3dot3 = 'git:3.3-tentative'
+
+head_2dot0 = 'git:cassandra-2.0'
+head_2dot1 = 'git:cassandra-2.1'
+head_2dot2 = 'git:cassandra-2.2'
+head_3dot0 = 'git:cassandra-3.0'
+head_3dot1 = 'git:cassandra-3.1'
+head_3dot2 = 'git:cassandra-3.2'
+head_3dot3 = 'git:cassandra-3.3'
+head_trunk = 'git:trunk'
```
mambocab: +1 on moving `git:` strings out into vars.

upgrade_through_versions_test.py
mambocab: Sorry, I see the `HEAD`-release distinction. However, I think the same logic may apply, but wrt the distinction between the latest release and a tentative tag.

mambocab: That said: I do see the benefit of treating the tagged release as the latest release, since we get to just change a value and run the same tests.

knifewine: right, and generally testing and upgrade path that ends with an already released version is pointless because if a bug is found the bug is, uh, already released. stepping through releases in the early points of the upgrade path, however, is important.

mambocab: A failure scenario I imagine here is:

- `3.0.X-tentative` gets tagged, bumping `3.0.W` (where `W = X-1`) out of the 'latest' slot
- Before `3.0.X` is released, a change is introduced to `git:cassandra-2.2` that breaks upgrade to `3.0.W`
- We don't catch the regression in 2.2 because we stopped testing the `git:cassandra-2.2` :arrow_forward: `3.0.W` upgrade path before that path was no longer supported.

knifewine: if there are better ways we can name things here, I'm all for it -- the 'latest' and 'head' notation is indeed a bit murky, but I'm at a loss for anything with more clarity to the purpose.

it gets even weirder when we might have less branches in git for tick-tock, and more temporary tags (just a hypothetical concern at this point though) -- since a tag isn't a head.

upgrade_through_versions_test.py
knifewine: yeah, despite introducing more classes, this makes the maintenance burden actually lower, as we can do most maintenance right here at the top of the module. the way we had it before kinda led to false confidence that things were being tested.

knifewine:
overall @ptnapoleon I agree it's not ideal as far as duplication goes, but I'd argue at this point we need coverage more than yet more upgrade code optimization (it's getting more difficult to justify time on these somewhat peripheral aspects).

@mambocab have discussed going back to a more generative approach but I think that's going to take time to get right, and we need to strike a balance there wrt keeping a nice user surface. dev's don't want to work with tests that require them to have an advanced degreee in upgrade-ology.

upgrade_through_versions_test.py
```diff
@@ -853,119 +861,235 @@ def create_upgrade_class(clsname, version_list, protocol_version,
...
+        bootstrap_test=True,
+        protocol_version=3
+    )
+create_upgrade_class(
+    'ProtoV3Upgrade_3_0_UpTo_3_1_HEAD',
+    [latest_3dot0, head_3dot1],
+    bootstrap_test=True,
+    protocol_version=3
+)
+if is_unreleased(latest_3dot1):
```
mambocab: I don't understand this one. Is there ever going to be another tentative 3.1?

Maybe the thing to do is to have a `latest_3dotX` constant?

upgrade_through_versions_test.py
mambocab: Agh, right -- and sometimes we're going to have point releases for already-released 3.X releases.

I guess the model in this PR is about as complex as we probably want to handle for the moment, so +1. That said, these concerns show are a reason we might want to move to something like a graph of valid upgrades that we use to generate upgrade paths, like we've been discussing privately.

mambocab:
I actually don't see problematic duplication here -- there's room to abstract, but I don't think we know the right abstractions yet.

upgrade_through_versions_test.py
```diff
@@ -853,119 +861,232 @@ def create_upgrade_class(clsname, version_list, protocol_version,
...
+    )
 create_upgrade_class(
-    'ProtoV3Upgrade_2_2_UpTo_3_0',
-    [latest_2dot2, 'git:cassandra-3.0'],
+    'ProtoV3Upgrade_3_0_UpTo_3_1_HEAD',
+    [latest_3dot0, head_3dot1],
     bootstrap_test=True,
     protocol_version=3
 )
+if is_unreleased(latest_3dot1):
```
mambocab: Ok, I think I understand this now: until 3.3 is released, we might come out with a 3.1.X to address some bug. Can you confirm that?

knifewine: right -- any release lines that could still see a release need the be targeted for testing when applicable (at this point via the dumb 'tentative' check).

upgrade_through_versions_test.py
knifewine: yeah, like we discussed in the thread below we want to keep that possibility open. in practicality, if there's a 3.1 patch release coming, we'd point latest_3dot1 to the suggested tag, it would be tested, and post-release we'd update latest_3dot1 again and it would be omitted from testing (via the is_unreleased mechanism).

mambocab:
I was not clear enough in my most recent comments -- I'm +1 on merging. We're pushing the bounds of the complexity we can manage successfully with the current approach, but that's fine for the time being.

PR #797, opened 2016-02-09
Expand docstring comments for the auth tests.

ptnapoleon:
@mambocab and @knifewine to review.

### DIFF
--- a/assertions.py
+++ b/assertions.py
@@ -1,5 +1,5 @@
 import re
-from cassandra import InvalidRequest, Unavailable, ConsistencyLevel, WriteFailure, WriteTimeout, ReadFailure, ReadTimeout
+from cassandra import InvalidRequest, Unavailable, ConsistencyLevel, WriteFailure, WriteTimeout, ReadFailure, ReadTimeout, Unauthorized
 from cassandra.query import SimpleStatement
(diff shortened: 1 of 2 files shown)


ptnapoleon:
If you're wondering why I changed all the assertions that check we have no permissions set, from comparing len() == 0 to comparing sets, is to improve debug output.

Previously if the tests failed, it would just say "Expected 0 rows, got X" where X is an int. Now it will say "Expected 0 rows, found X", where X are the actual rows pulled from C*.

mambocab:
My comment on `prepare()` applies everywhere it's described as `launch a one-node cluster`. It's a little nit, but it should be pretty easy to clean up too, so I think you should hold merge on fixing it.

mambocab:
Logging off, but I just wanted to note that I've reviewed through to `create_ks_auth_test`.

ptnapoleon:
@knifewine do you have bandwidth to resume Jim's review?

ptnapoleon:
The newest commit about your desire for namedtuples breaks everything [specifically the assertions]? Is there a more correct way to make these into namedtuples? If not, why did you want that?

knifewine:
+1

PR #798, opened 2016-02-09
Remove upgrade_tests default upgrade logic

mambocab:
I feel like there's some cleanup to do in this history, and these commits will need to be rebased onto master after #787 is merged, but the code is in place.

The big change is the removal of `upgrade_tests.upgrade_base.get_default_upgrade_path` and the code that calls it. It's replaced by code that extends the upgrade-test-generation code introduced in #787 by attaching the specified upgrade path to the test class, rather than specifying it through a mess of environment variables and weird logic defining the default path.

This is currently running here:

http://cassci.datastax.com/view/Dev/view/mambocab/job/mambocab-consolidated-upgrades/20/

We can use this job to run all our upgrade tests on a single job if we like.

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 
+import itertools
 import math
 import random
 import struct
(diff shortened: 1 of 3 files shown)


mambocab:
I will want feedback from @ptnapoleon and @knifewine for this one. Like #787, we can't just merge this, because there are changes that will need to be made in our CassCI jobs. We may also need to change our exclude config files.

ptnapoleon:
All seems good to me

mambocab:
Rebasing now.

mambocab:
Ok, rebase done. Requesting a review from @hatch.

upgrade_tests/cql_tests.py
```diff
@@ -5225,26 +5228,27 @@ def bug_10652_test(self):
...
-     'CL': ConsistencyLevel.ALL,
-     '__test__': True},
+     'CL': ConsistencyLevel.ALL},
     {'NODES': 2,
-     'RF': 1,
-     '__test__': True},
+     'RF': 1},
 ]
+specs = [dict(s, UPGRADE_PATH=p, __test__=True)
+         for s, p in itertools.product(topology_specs, VALID_UPGRADE_PAIRS)]
```
knifewine: nice

knifewine:
+1, lgtm

mambocab:
Seems to be running more or less happily here:

http://cassci.datastax.com/view/Dev/view/mambocab/job/mambocab-consolidated-upgrades/21/console

So, merging \o/

PR #799, opened 2016-02-10
gate upgrade test discovery on env var

mambocab:
This sets `__test__` based on the value of the `UPGRADE_TEST_RUN` env var -- so, rather than skip the test, it removes it from test discovery entirely. This works around problems we had on non-upgrade jobs after merging #798, where `nosetests` threw up when passed too many test names.

On my machine, this seems to work:

```sh
$ CASSANDRA_VERSION=git:trunk nosetests --collect-only 2>&1 -vv | grep upgrade_tests  
upgrade_tests/regression_test.py:TestForRegressions.test_10822 ... ok
                                                                                                                                  
$ UPGRADE_TEST_RUN=true CASSANDRA_VERSION=git:trunk nosetests --collect-only 2>&1 -vv | grep upgrade_tests | wc -l
2921
```

Can a reviewer please verify?

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -23,7 +23,7 @@ from thrift_bindings.v22.ttypes import (CfDef, Column, ColumnOrSuperColumn,
                                         Mutation)
 from thrift_tests import get_thrift_client
 from tools import known_failure, require, rows_to_list, since
-from upgrade_base import VALID_UPGRADE_PAIRS, UpgradeTester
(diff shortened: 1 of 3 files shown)


mshuler:
+1 lgtm - we looked at the test buckets created for a dtest run with this patch, and no upgrade_tests/ tests!

mambocab:
ftr, here's the job we ran: http://cassci.datastax.com/view/Dev/view/mambocab/job/mambocab-cassandra-3.0-dtest/69

PR #800, opened 2016-02-10
Test for CASSANDRA-11050

pcmanus:
Test to reproduce CASSANDRA-11050, to commit when the issue itself is committed.


ptnapoleon:
+1

PR #801, opened 2016-02-10
Fix offheap configurations in index_summary_upgrade_test

mshuler:
Local testing works as expected now - notes on https://datastax.jira.com/browse/CSTAR-276

### DIFF
--- a/index_summary_upgrade_test.py
+++ b/index_summary_upgrade_test.py
@@ -1,6 +1,6 @@
 from cassandra.concurrent import execute_concurrent_with_args
 
-from dtest import Tester, debug
+from dtest import OFFHEAP_MEMTABLES, Tester, debug
 from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
 from tools import known_failure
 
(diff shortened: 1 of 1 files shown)


index_summary_upgrade_test.py
```diff
@@ -80,6 +86,11 @@ def test_upgrade_index_summary(self):
         node.watch_log_for("DRAINED")
         node.stop()
         cluster.set_install_dir(original_install_dir)
+        if OFFHEAP_MEMTABLES:
+            # memtable_allocation_type is not supported in the 3.0-3.3 branches
+            if '3.0' <= cluster.version() < '3.4':
+                cluster._config_options.__delitem__("memtable_allocation_type")
+                debug("Removed memtable_allocation_type from %s configuration" % cluster.version())
```
ptnapoleon: Could you change the string formatting you did here? Instead make it
`debug("Removed memtable_allocation_type from {} configuration".format(cluster.version()))`

ptnapoleon:
+1 after the minor fix

PR #802, opened 2016-02-10
Novnode flag

mambocab:
Probably not ready for merge yet, just looking to start a conversation. @ptnapoleon?

### DIFF
--- a/plugins/dtestconfig.py
+++ b/plugins/dtestconfig.py
@@ -1,0 +1,42 @@
+from collections import namedtuple
+
+from nose import plugins
+
+# A class that defines the attributes that have to be defined for configuring
+# a dtest run. namedtuple does what we want -- it's immutable and requires
+# all the attributes to be passed in to be instantiated.
+GlobalConfigObject = namedtuple('GlobalConfigObject', [
+    'vnodes',  # disable or enable vnodes
+])
+
+_CONFIG = None
+
+
+class DtestConfigPlugin(plugins.Plugin):
+    """
+    Pass in configuration options for the dtests.
+    """
+    enabled = True  # if this plugin is loaded at all, we're using it
+    name = 'dtest_config'
+
+    def __init__(self, config=None):
+        """
+        Instantiate this plugin with a GlobalConfigObject or, by default, None.
+        Then, set  the global _CONFIG constant with the value of the plugin.
+
+        This is a little weird, yes, but nose seems to generally be built
+        around the idea that a given plugin will be instantiated only once, so
+        this provides a way for test framework code to grab the value off this
+        module. We want that, since the plugin itself isn't available to test
+        code.
+
+        @param config an object meeting the GlobalConfigObject spec that will
+                      be used as configuration for a dtest run.
+        """
+        self.CONFIG = config
+
+        global _CONFIG
+        _CONFIG = self.CONFIG
+
+    def configure(self, options, conf):
+        pass
--- a/dtest.py
+++ b/dtest.py
@@ -76,7 +76,7 @@ DISABLE_VNODES = not CONFIG.vnodes
 
 
 if os.environ.get('DISABLE_VNODES', '').lower() in ('yes', 'true'):
-    print 'DISABLE_VNODES environment variable deprecated. Use `runner.py --no-vnodes` instead.'
+    print 'DISABLE_VNODES environment variable deprecated. Use `./run_dtests.py --vnodes false` instead.'
 
 
 CURRENT_TEST = ""
--- a/dtest.py
+++ b/dtest.py
@@ -28,6 +28,11 @@ from ccmlib.common import is_win
 from ccmlib.node import TimeoutError
 from nose.exc import SkipTest
 
+# We don't want test files to know about the plugins module, so we import
+# constants here and re-export them.
+from plugins.dtestconfig import _CONFIG as CONFIG
+from plugins.dtestconfig import GlobalConfigObject
+
 LOG_SAVED_DIR = "logs"
 try:
     os.mkdir(LOG_SAVED_DIR)
@@ -51,7 +56,6 @@ TRACE = os.environ.get('TRACE', '').lower() in ('yes', 'true')
 KEEP_LOGS = os.environ.get('KEEP_LOGS', '').lower() in ('yes', 'true')
 KEEP_TEST_DIR = os.environ.get('KEEP_TEST_DIR', '').lower() in ('yes', 'true')
 PRINT_DEBUG = os.environ.get('PRINT_DEBUG', '').lower() in ('yes', 'true')
-DISABLE_VNODES = os.environ.get('DISABLE_VNODES', '').lower() in ('yes', 'true')
 OFFHEAP_MEMTABLES = os.environ.get('OFFHEAP_MEMTABLES', '').lower() in ('yes', 'true')
 NUM_TOKENS = os.environ.get('NUM_TOKENS', '256')
 RECORD_COVERAGE = os.environ.get('RECORD_COVERAGE', '').lower() in ('yes', 'true')
@@ -59,6 +63,21 @@ REUSE_CLUSTER = os.environ.get('REUSE_CLUSTER', '').lower() in ('yes', 'true')
 SILENCE_DRIVER_ON_SHUTDOWN = os.environ.get('SILENCE_DRIVER_ON_SHUTDOWN', 'true').lower() in ('yes', 'true')
 IGNORE_REQUIRE = os.environ.get('IGNORE_REQUIRE', '').lower() in ('yes', 'true')
 DATADIR_COUNT = os.environ.get('DATADIR_COUNT', '3')
+
+# devault values for configuration from configuration plugin
+_default_config = GlobalConfigObject(
+    vnodes=True,
+)
+
+if CONFIG is None:
+    CONFIG = _default_config
+
+DISABLE_VNODES = not CONFIG.vnodes
+
+
+if os.environ.get('DISABLE_VNODES', '').lower() in ('yes', 'true'):
+    print 'DISABLE_VNODES environment variable deprecated. Use `runner.py --no-vnodes` instead.'
+
 
 CURRENT_TEST = ""
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ psutil
 pycassa
 six
 thrift
+docopt
--- a/run_dtests.py
+++ b/run_dtests.py
@@ -1,0 +1,251 @@
+#!/usr/bin/env python
+"""
+Usage: runner.py [--nose-options NOSE_OPTIONS] [TESTS...] [--vnodes VNODES_OPTIONS...]
+                 [--runner-debug | --runner-quiet] [--dry-run]
+
+nosetests options:
+    --nose-options NOSE_OPTIONS  specify options to pass to `nosetests`.
+    TESTS                        space-separated list of tests to pass to `nosetests`
+
+script configuration options:
+    --runner-debug -d            print debug statements in this script
+    --runner-quiet -q            quiet all output from this script
+
+cluster configuration options:
+    --vnodes VNODES_OPTIONS...   specify whether to run with or without vnodes.
+                                 valid values: 'true' and 'false'
+"""
+from __future__ import print_function
+
+import subprocess
+from collections import namedtuple
+from itertools import product
+from os import getcwd
+from tempfile import NamedTemporaryFile
+
+from docopt import docopt
+
+from plugins.dtestconfig import GlobalConfigObject
+
+
+# Generate values in a matrix from these lists of values for each attribute
+# not defined in arguments to the runner script.
+default_config_matrix = GlobalConfigObject(
+    vnodes=(True, False),
+)
+
+
+def _noop(*args, **kwargs):
+    pass
+
+
+class ValidationResult(namedtuple('_ValidationResult', ['serialized', 'error_messages'])):
+    """
+    A value to be returned from validation functions. If serialization works,
+    return one with 'serialized' set, otherwise return a list of string on the
+    'error_messages' attribute.
+    """
+    __slots__ = ()
+
+    def __new__(cls, serialized=None, error_messages=None):
+        if error_messages is None:
+            error_messages = []
+
+        success_result = serialized is not None
+        failure_result = bool(error_messages)
+
+        if success_result + failure_result != 1:
+            msg = ('attempted to instantiate a {cls_name} with serialized='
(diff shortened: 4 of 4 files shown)


ptnapoleon:
Looking good so far? I'm interested in where you're putting the logic for running a matrix

knifewine:
Didn't dive very deep, but I like it.

mambocab:
Pushed a couple small changes to names. I think this actually is ready for merge, though we would need to change the `novnodes` job to do so. Are we good with that? We will also be changing CassCI jobs up to use the testing matrix soon, so we might want to get all the CassCI churn out of the way in one shot.

mambocab:
Oh yeah, and you can smoke test this with this file:

https://gist.github.com/mambocab/31e94af3cf649f41ade7

mambocab:
The history needs some serious grooming and apparently Travis is mad, but I've changed the API here -- instead of just letting `nose.main` grab `sys.argv`, we're passing it an argv programatically. I think this implies some changes to the configuration plugin, since configuration will happen in the args to `runner.py` and not to `nosetests`.

Thoughts, @knifewine? I've already discussed this privately with @ptnapoleon and he says it is Fine.

mambocab:
Example command:

```sh
CASSANDRA_VERSION=git:trunk ./runner.py --nose-options '--nocapture --collect-only' putget_test.py
```

knifewine:
if it's tested and working, I'm a +1

mambocab:
Ok, code is re-ready for review. This now goes beyond the original scope -- it doesn't just replace the environment variable for turning off vnodes, it also uses `runner.py` to generate a test matrix.

I recommend reviewers mess around with this file as a smoke test:

https://gist.github.com/mambocab/31e94af3cf649f41ade7

A good starting point might be

```sh
wget https://gist.githubusercontent.com/mambocab/31e94af3cf649f41ade7/raw/f6ce5cc97ed8d72da3bf32e7ca43b7020d251fcc/novnode_smoke_test.py

set -x

CASSANDRA_VERSION=git:trunk ./runner.py 2>&1 novnode_smoke_test.py --nose-option="--with-id --verbosity=3 --nocapture" --debug
CASSANDRA_VERSION=git:trunk ./runner.py 2>&1 novnode_smoke_test.py --nose-option="--with-id --verbosity=3 --nocapture" --debug --vnodes true
CASSANDRA_VERSION=git:trunk ./runner.py 2>&1 novnode_smoke_test.py --nose-option="--with-id --verbosity=3 --nocapture" --debug --vnodes false
CASSANDRA_VERSION=git:trunk ./runner.py 2>&1 novnode_smoke_test.py --nose-option="--with-id --verbosity=3 --nocapture" --debug --vnodes true --vnodes false
```

Do all of these behave as you expect?

mambocab:
(ping @ptnapoleon and @knifewine for review!)

plugins/dtestconfig.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class DtestConfigPlugin(plugins.Plugin):
+    """
+    Pass in configuration options for the dtests.
+    """
+    enabled = True  # if this plugin is loaded at all, we're using it
+    name = 'dtest_config'
+
+    def __init__(self, config=None):
```
ptnapoleon: Could you use the doxygen `@param` markup in here?

Also, it doesn't seem that we are checking that `config` is a `GlobalConfigObject`

runner.py
```diff
@@ -0,0 +1,205 @@
...
+                 [--debug]
+
+nosetests options:
+    --nose-options NOSE_OPTIONS  specify options to pass to `nosetests`.
+    TESTS                        space-separated list of tests to pass to `nosetests`
+    --debug -d                   print debug statements
+
+cluster configuration options:
+    --vnodes VNODES_OPTIONS...   specify whether to run with or without vnodes.
+                                 valid values: 'true' and 'false'
```
ptnapoleon: nit: valid options

runner.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __new__(cls, serialized=None, error_messages=None):
+        if error_messages is None:
+            error_messages = []
+        return super(ValidationResult, cls).__new__(cls, serialized=serialized, error_messages=error_messages)
+
+
+def _validate_and_serialize_vnodes(vnodes_value):
+    """
+    Validate the values recieved for vnodes configuration. Returns a
```
ptnapoleon: nit: `received`.

runner.py
```diff
@@ -0,0 +1,205 @@
...
+    options = docopt(__doc__)
+    validated_options = validate_and_serialize_options(options)
+
+    nose_option_list = options['--nose-options'].split()
+    test_list = options['TESTS']
+    nose_argv = nose_option_list + test_list
+
+    # function for printing debug statements
+    if options['--debug']:
+        debug = print
```
ptnapoleon: Does this work in py2?

runner.py
```diff
@@ -0,0 +1,205 @@
...
+        to_execute = (
+            'import nose\n'
+            'from plugins.dtestconfig import DtestConfigPlugin, GlobalConfigObject\n'
+            'nose.main(addplugins=[DtestConfigPlugin({config})],\n'
+            '          argv={nose_argv})\n'
+        ).format(config=repr(config), nose_argv=repr(nose_argv))
+        temp = NamedTemporaryFile(dir=getcwd())
+        debug('Writing the following to {}:'.format(temp.name))
+        debug(to_execute)
+        debug()
```
ptnapoleon: Do we really need this? Can't you just add a newline on the previous call

runner.py
```diff
@@ -0,0 +1,205 @@
...
+        temp.write(to_execute)
+        temp.flush()
+
+        results.append(subprocess.call(['python', temp.name]))
+        print()
```
ptnapoleon: Why print() here?

runner.py
```diff
@@ -0,0 +1,205 @@
...
+    return GlobalConfigObject(
+        vnodes=vnodes.serialized or default_config_matrix.vnodes
+    )
+
+
+def product_of_values(d):
+    """
+    Transforms a dictionary of {key: list(configuration_options} into a list of
+    dictionaries that takes the product of the lists at the keys and returns a
+    tuple of dictionaries, each corresponding to a point in the product, with
```
ptnapoleon: So which are we returning? A list of dicts, or a tuple of dicts?

runner.py
mambocab: Based on the documentation for `docopt`, I think that `VNODES_OPTIONS` is an "argument" and valid "values" for it are `true` and `false`. That's my understanding.

runner.py
mambocab: Yes. I started the file with `from __future__ import print_function`.

runner.py
mambocab: Sure.

runner.py
mambocab: I'll add a comment. The `subprocess.call` call uses `stdout` and `stderr` from the parent process, and the `print()` call adds a newline after all that gets printed, separating it from the beginning of the next `subprocess.call` output.

runner.py
mambocab: Good catch, thanks.

mambocab:
@ptnapoleon I believe I've addressed all your comments in the most recent commit.

Can you think of a better way to explain what `product_of_values` does? because, as you can tell, I had a heck of a time.

plugins/dtestconfig.py
mambocab: Just pushed the markup.

I don't think it's good style to check types. I think the idiomatic thing to do is just let it fail where it'd fail if it doesn't quack like a duck.

runner.py
ptnapoleon: Good thing I noticed that.

ptnapoleon:
The new version of `product_of_values` is much better.

ptnapoleon:
I am +1

ptnapoleon:
Running `runner.py` without options should print out the help text, not

```
 python runner.py
  File "runner.py", line 30, in <module>
    nose_option_list = options['--nose-option'].split()
AttributeError: 'NoneType' object has no attribute 'split'
```

ptnapoleon:
Actually anytime we leave out nose options it throws that exception.

ptnapoleon:
Shouldn't `python runner.py --nose-option '-xsv' topology_test.py` run with and without vnodes?

ptnapoleon:
Why am I seeing this?
```
python runner.py --nose-option '-xsv' topology_test.py --debug
Usage: runner.py [--nose-option NOSE_OPTIONS] [TESTS...]
```

ptnapoleon:
Ugh, I've been running against the wrong SHA. H/o

ptnapoleon:
I'm passing `--nose-option` instead of `--nose-options`, and it isn't complaining. Also not working, at least.

ptnapoleon:
`--nose-options '-xs -vvv'` Is as verbose as I'd expect, but `--nose-options '-xsv'` definitely is not. It seems like something is wrong

mambocab:
> `--nose-options '-xs -vvv'` Is as verbose as I'd expect, but `--nose-options '-xsv'` definitely is not. It seems like something is wrong

See the comment here 

https://github.com/riptano/cassandra-dtest/commit/6963764ec7b26c5a2313bbf167f4938b30bfbe28#diff-2ef65ef8520cef01c37a957e90545296R208

:angry: Dangit, computers.

mambocab:
> I'm passing `--nose-option` instead of `--nose-options`, and it isn't complaining.

Turns out that is a feature, not a bug:

https://github.com/docopt/docopt/blob/master/docopt.py#L308

```
❯ cat wat.py
"""
Usage: wat.py [--options OPTIONS...]
"""
from docopt import docopt
from pprint import pprint

if __name__ == '__main__':
    pprint(docopt(__doc__))
                                                                                                                                  
❯ python wat.py --opti WAT
{'--options': True,
 'OPTIONS': ['WAT']}
```

`docopt` will assume that, e.g. `--opti` means `--option` because it's an unambiguous prefix. Same goes for `--nose-option` and `--nose-options`.

knifewine:
if possible, I'd love some output on what's about to happen before tests run. debug gives some output but I think that only covers the first setting permutation at the beginning.

ideally we'd be able to see up front all permutations before the first one starts off (if possible). rather than having this tied to --debug, which may include other verbose output, maybe we make it something like --preview/--no-preview ?

we should maybe formalize how --debug of runner.py relates to debug output in dtest -- are they the same or distinct?

mambocab:
> ideally we'd be able to see up front all permutations before the first one starts off (if possible). rather than having this tied to --debug, which may include other verbose output, maybe we make it something like --preview/--no-preview ?

I can just use a verbosity flag, like nose does -- `-v` is basically `quiet`, `-vv` is default, with the preview, `-vvv` is debug. Sound reasonable?

> we should maybe formalize how --debug of runner.py relates to debug output in dtest -- are they the same or distinct?

They're completely distinct; debug stuff in `dtest` is still (unfortunately) controlled by environment variables. The right way to fix that is probably to use built-in logging facilities, since `nose` plays nicely with that.

ptnapoleon:
I'd like a verbosity flag, yes. But as mentioned offline, that isn't quite nose's behavior. And yes the two debug outputs should be different

knifewine:
If some kind of permutation preview (all shown before running starts) is the default, I'm happy. In my mind this would be the first level of debug output, with other noise being opted-in to at higher levels.

knifewine:
I'm still a bit uneasy about runner.py and dtest both having something called "debug" because I think that's likely to be a point of confusion, but I also don't want to tangle up the command line arguments code. Seems that eventually this script could also have an argument for dtest debugging? -- so ideally we just give the runner debug cmdline a distinct name to leave room for the dtest debug option later?

ptnapoleon:
I agree with Russ.

mambocab:
Just pushed a couple more changes to improve debug output and change the names of the `--debug` and `--quiet` flags to `--runner-{debug,quiet}`.

knifewine:
one more user surface nit, I think 'runner.py' should maybe be renamed to something like 'testrunner.py' (well, crud, I guess that might mean nose would try and pick it up).

right now with a name like 'runner' it's kinda non-obvious that it's meant to be the test running entry point.

mambocab:
At @knifewine's suggestion, I added the output about the full config matrix to the default output. I also added a `--dry-run` option so users can see the effects of the options they specify before running the tests.

mambocab:
I've also renamed the runner. We don't have to worry about test discovery; `nosetests` automatically excludes executable files.

knifewine:
I got my console output wishes, i'm +1 based on my local testing (didn't dig into the code itself).

ptnapoleon:
+1

mambocab:
I've smushed all of these into one big commit. Now the question is: how do we want to handle CassCI jobs post-merge? The regular jobs -- the ones that run with vnodes -- should be unaffected. We will either need to 1) change all the novnodes jobs to use the runner script with `--vnodes false` or 2) migrate to matrix jobs. [Number 2 isn't ready yet](http://cassci.datastax.com/job/mambocab-sample-matrix-job/10/testReport/) but I'll keep working at that. So maybe number 1? Thoughts, @ptnapoleon and @mshuler?

mshuler:
No.1 should be a simple template edit.

As long as we work through all the details to making sure matrix jobs do most of the things we want, No.2 is fine.
* do they work with autojobs templates?
* does the jenkins reporting work for us?
* does that work with cstar_report, too?
* does the bucketizer work with matrices? (seems to be working in the test job)
* can we tell which report goes with which matrix option easily?

What do we gain by using matrix jobs? :)

mambocab:
> * do they work with autojobs templates?

I don't know enough about how templating works -- I'm guessing we'd have to make a new matrix template job.

The other option is to migrate to the [job creator](https://github.com/riptano/job-creator) for this. I have no idea how that'd interact with `autojobs`' templating.

> * does the jenkins reporting work for us?

Remains to be seen what the reporting will actually look like; I still have to fix these jobs up.

> * does that work with cstar_report, too?

Oof, great question, and one I hadn't looked at at all.

> * does the bucketizer work with matrices? (seems to be working in the test job)

Yep, not a problem as far as I can tell. Each subjob is more or less a normal CassCI job, just kicked off differently.

> * can we tell which report goes with which matrix option easily?

Same as above -- I don't know yet, we'll see how reporting turns out.

> What do we gain by using matrix jobs? :)

The gain now is that we'll be able to delete all the `*novnode_dtest` jobs:

http://cassci.datastax.com/search/?q=novnode_dtest

This is a maintenance win for us -- we wind up running the same tests under the same conditions, but with fewer jobs to keep updated and to monitor. In the long term, it'll also be pretty easy to add rows to the matrix for things like offheap memtables.

mambocab:
I have set up the jenkins template for novnodes jobs and am merging this.

PR #803, opened 2016-02-11
Harden copy bulk tests by increasing truncate timeout

stef1927:
`test_bulk_round_trip_blogposts_with_max_connections` failed because of a timeout during a truncate operation, more details in CASSANDRA-11157. 

I've also removed a known failure for a ticket that was resolved.

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -21,7 +21,7 @@ from ccmlib.common import is_win
 from cqlsh_tools import (DummyColorMap, assert_csvs_items_equal, csv_rows,
                          monkeypatch_driver, random_list, unmonkeypatch_driver, write_rows_to_csv)
 from dtest import Tester, canReuseCluster, freshCluster, debug, DISABLE_VNODES
-from tools import known_failure, rows_to_list, since
+from tools import rows_to_list, since
 
 DEFAULT_FLOAT_PRECISION = 5  # magic number copied from cqlsh script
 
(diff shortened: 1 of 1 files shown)


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -2087,7 +2085,7 @@ def test_source_copy_round_trip(self):
     def _test_bulk_round_trip(self, nodes, partitioner,
                               num_operations, profile=None,
                               stress_table='keyspace1.standard1',
-                              configuration_options=None,
+                              configuration_options={},
```
mambocab: I see the error here, but this fix can create some problems, as the dictionary object is created at function-definition time, and changes to it will persist across calls to `_test_bulk_round_trip` (see [this article](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) for a more thorough explanation).

Looks like the `copy_to_options` parameter has the same problem. I've fixed these problems in #804, which should be merged by the time you read this. Could you please rebase onto this change (or at least the commit from #804)?

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -2159,6 +2163,7 @@ def run_copy_from(filename):
         self.assertEqual(sum(1 for _ in open(tempfile1.name)),
                          sum(1 for _ in open(tempfile2.name)))
 
+    @freshCluster()
```
mambocab: Why have these been added? They're probably fine, but I haven't seen you explain them anywhere.

stef1927: They should have been already there because these tests require a cluster with a specific number of nodes or configuration options.

cqlsh_tests/cqlsh_copy_tests.py
stef1927: Thank you for the fixes and for the article, I definitely learned something today! :)

stef1927:
Rebased, thank you.

cqlsh_tests/cqlsh_copy_tests.py
mambocab: :+1: Thanks for the explanation.

mambocab:
Looks good, thanks for rebasing and for the explanation! Merging.

PR #806, opened 2016-02-11
generate upgrade classes for regression_test

mambocab:
Can I get a review from @knifewine on this one? I don't know if the intent for this test is to test all the upgrade paths we care about (in which case this approach is the one we want), or if it tests for a regression on one particular upgrade path (in which case this isn't what we want to do).

### DIFF
--- a/upgrade_tests/regression_test.py
+++ b/upgrade_tests/regression_test.py
@@ -2,14 +2,15 @@
 Home for upgrade-related tests that don't fit in with the core upgrade testing in dtest.upgrade_through_versions
 """
 from cassandra import ConsistencyLevel as CL
-from upgrade_base import UpgradeTester
+
+from upgrade_base import UPGRADE_TEST_RUN, VALID_UPGRADE_PAIRS, UpgradeTester
 
 
 class TestForRegressions(UpgradeTester):
     """
     Catch-all class for regression tests on specific versions.
     """
-    NODES, RF, __test__, CL = 2, 1, True, CL.ONE
+    NODES, RF, __test__, CL = 2, 1, False, CL.ONE
 
     def test_10822(self):
         """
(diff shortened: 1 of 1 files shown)


upgrade_tests/regression_test.py
```diff
@@ -2,14 +2,15 @@
...
+
+from upgrade_base import UPGRADE_TEST_RUN, VALID_UPGRADE_PAIRS, UpgradeTester
 
 
 class TestForRegressions(UpgradeTester):
     """
     Catch-all class for regression tests on specific versions.
     """
-    NODES, RF, __test__, CL = 2, 1, True, CL.ONE
+    NODES, RF, __test__, CL = 2, 1, False, CL.ONE
```
knifewine: remind me what the boolean is here?

knifewine:
I think the original case was intended to target a singular upgrade, not all upgrade pairs. So we could consider trimming this down to that single upgrade, or let it run on all. Not sure what the cost/benefit is here.

upgrade_tests/regression_test.py
mambocab: it's `__test__`; setting it to `False` removes it from test discovery.

knifewine: oh right, thanks

mambocab:
I'm ok with erring on the side of testing too much -- the cost isn't super-high since we're only running one test method for now. @knifewine Could you document the upgrade path this originally failed over? As long as the documentation's there, there's no harm in running more tests than necessary right now IMO.

knifewine:
OK, @mambocab I made a pr on top of your fork branch to update the docstring, https://github.com/mambocab/cassandra-dtest/pull/4

knifewine:
also, +1

mambocab:
Great, thank you.

PR #807, opened 2016-02-11
tag known failures

mambocab:
Tags failures from [CASSANDRA-11162](https://issues.apache.org/jira/browse/CASSANDRA-11162).

Looks like we might not be able to do the `TestClass.test_method = known_failure(**known_failure_args)(TestClass.test_method)` thing. For one thing, because each instance of `test_method` is actually the same object, you might as well just decorate the test on the base class. Also, I got a kinda cryptic errorThe error I got when I tried to actually do this was kinda cryptic, though, so I could be wrong here. I didn't dig too far. @knifewine and @ptnapoleon, could you review, please, and try to think of a different solution?

### DIFF
--- a/upgrade_tests/upgrade_through_versions_test.py
+++ b/upgrade_tests/upgrade_through_versions_test.py
@@ -260,6 +260,9 @@ class UpgradeTester(Tester):
                 values=dict(self.extra_config)
             )
 
+    @known_failure(failure_source='test',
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I cannot think of a better thing to do? Discussion of merits on rewriting `known_failure` happened offline. +1

mambocab:
Yeah, rewriting `known_failure` would be pretty low-value at this point. We'll see how things go as-is. Should be fine, IMO.

Thanks for the review, merging.

PR #809, opened 2016-02-13
Logwatch callback

knifewine:
adds support for continuous watching of cassandra logs, to enable closer association between dtest actions and resulting log failures (esp among upgrade tests).

when enabled, this will trigger ccm to open a continuous log watching thread, which will call back when errors are encountered, and dtest can then take actions to halt and cleanup test execution.

currently gated by an env var, but that will change soon, in line with command line flags

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -368,7 +368,7 @@ class Tester(TestCase):
         """
         Callback handler used in conjunction with begin_active_log_watch.
         When called prepares exception instance, then will indirectly
-        cause catch_interrupt to be called, which can raise the exception in the main
+        cause _catch_interrupt to be called, which can raise the exception in the main
         program thread.
         """
         # in some cases self.allow_log_errors may get set after proactive log checking has been enabled
--- a/dtest.py
+++ b/dtest.py
@@ -62,6 +62,7 @@ REUSE_CLUSTER = os.environ.get('REUSE_CLUSTER', '').lower() in ('yes', 'true')
 SILENCE_DRIVER_ON_SHUTDOWN = os.environ.get('SILENCE_DRIVER_ON_SHUTDOWN', 'true').lower() in ('yes', 'true')
 IGNORE_REQUIRE = os.environ.get('IGNORE_REQUIRE', '').lower() in ('yes', 'true')
 DATADIR_COUNT = os.environ.get('DATADIR_COUNT', '3')
+ENABLE_ACTIVE_LOG_WATCHING = os.environ.get('ENABLE_ACTIVE_LOG_WATCHING', '').lower() in ('yes', 'true')
 
 CURRENT_TEST = ""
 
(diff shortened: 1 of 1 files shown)


knifewine:
depends on https://github.com/pcmanus/ccm/pull/455

knifewine:
currently running trial on cassci http://cassci.datastax.com/view/Upgrades/job/scratch_upgrade_with_log_watching/2/

ptnapoleon:
Should the fact that the cassci build you linked, aborted and has 51 failures worry me?

dtest.py
```diff
@@ -346,6 +353,58 @@ def setUp(self):
...
+        for nodename, errors in reportable_errordata.items():
+            message += "\n{nodename}: {errors}".format(nodename=nodename, errors=errors)
+
+        # thread.interrupt_main will SIGINT in the main thread, which we can
+        # catch to raise an exception with useful information
+        debug('Errors were just seen in logs, ending test!')
+        self.exit_with_exception = AssertionError(message)
+        thread.interrupt_main()
+
+    def _catch_interrupt(self, signal, frame):
```
ptnapoleon: Could you add a docstring here too?

dtest.py
```diff
@@ -346,6 +353,58 @@ def setUp(self):
...
+
+        In the event that errors are seen in logs, ccm will call back to _log_error_handler.
+        """
+        # log watching happens in another thread, but we want it to halt the main
+        # thread's execution, which we have to do by registering a signal handler
+        signal.signal(signal.SIGINT, self._catch_interrupt)
+        self.cluster.actively_watch_logs_for_error(self._log_error_handler, interval=0.25)
+
+    def _log_error_handler(self, errordata):
+        """
```
ptnapoleon: Could you add `@param errordata` line to this docstring that explains the format errordata should be in?

ptnapoleon:
Code looks good to me, but I'm not sure if that cassci result means it is working well or not

knifewine:
retrying since last build aborted, unsure what exactly the problem was http://cassci.datastax.com/view/Upgrades/job/scratch_upgrade_with_log_watching/3/

dtest.py
knifewine: pushed.

dtest.py
knifewine: pushed.

knifewine:
finally got a decent CI run here http://cassci.datastax.com/view/Upgrades/job/scratch_upgrade_with_log_watching/7/testReport/

knifewine:
going to try a dtest job with the log interrupt enabled as well

knifewine:
trunk dtest run with log watching enabled seems ok https://cassci.datastax.com/view/Dev/view/knifewine/job/rhatch_scratch_trunk_dtest_log_interrupt/1/testReport/

ptnapoleon:
+1 then

PR #810, opened 2016-02-16
Set test_mutation_v2 to only run on 2.2.x

mshuler:


ptnapoleon:
Shouldn't this also run on 2.1.x?

mshuler:
The entire test class has `@since('2.2')` so I was just going off of that.

ptnapoleon:
Makes sense then. +1. Merge away

PR #811, opened 2016-02-16
Repair audit

ptnapoleon:
Looking for review on the work so far, which is mostly just comments.

@knifewine and @mambocab


repair_tests/deprecated_repair_test.py
```diff
@@ -172,7 +149,9 @@ def _deprecated_repair_jmx(self, method, arguments):
         # wait for log to start
         node1.watch_log_for("Starting repair command")
         # get repair parameters from the log
-        l = node1.grep_log("Starting repair command #1, repairing keyspace ks with repair options \(parallelism: (?P<parallelism>\w+), primary range: (?P<pr>\w+), incremental: (?P<incremental>\w+), job threads: (?P<jobs>\d+), ColumnFamilies: (?P<cfs>.+), dataCenters: (?P<dc>.+), hosts: (?P<hosts>.+), # of ranges: (?P<ranges>\d+)\)")
+        l = node1.grep_log(("Starting repair command #1, repairing keyspace ks with repair options \(parallelism: (?P<parallelism>\w+), primary range: (?P<pr>\w+), "
```
mambocab: Side note: this is a kinda awesome way to extract values from executables' output, we could learn from this elsewhere.

As far as style goes, it might be even more readable to break the string up further, like, putting each field on its own line or something. Up to you, I don't feel strongly, except that the existing change is definitely better than it was before.

repair_tests/incremental_repair_test.py
```diff
@@ -44,7 +52,6 @@ def sstable_marking_test(self):
         else:
             node3.nodetool("repair -par -inc")
 
-        output = ""
         with open('sstables.txt', 'w') as f:
```
mambocab: So, `run_sstablemetadata` appends to `output_file`? I'm suspicious.

repair_tests/incremental_repair_test.py
```diff
@@ -58,6 +65,23 @@ def sstable_marking_test(self):
         os.remove('sstables.txt')
 
     def multiple_repair_test(self):
+        """
+        * Launch a three node cluster
+        * Create a keyspace with RF 3 and a table
+        * Insert 50 rows
```
mambocab: If I understand what `insert_c1c2(session, keys=range(1, 50), consistency=ConsistencyLevel.ALL)` means, this only inserts 49 keys:

```python
>>> len(list(range(1, 50)))
49
>>> 
```

mambocab: `s/range(1, 50)/range(50)/` would make it actually insert 50 rows.

repair_tests/incremental_repair_test.py
```diff
@@ -58,6 +65,23 @@ def sstable_marking_test(self):
...
+        * Stop node3
+        * Insert 100 rows
```
mambocab: This is incorrect -- the statement inserts rows at keys 51-100, inclusive, so there are a total of 99 rows.

repair_tests/incremental_repair_test.py
```diff
@@ -58,6 +65,23 @@ def sstable_marking_test(self):
...
+        * Restart node3
+        * Issue an incremental repair
```
mambocab: on node3.

repair_tests/incremental_repair_test.py
```diff
@@ -58,6 +65,23 @@ def sstable_marking_test(self):
...
+        * Stop node2
+        * Insert 50 rows
+        * Restart node2
+        * Issue an incremental repair
```
mambocab: on node2.

repair_tests/incremental_repair_test.py
```diff
@@ -58,6 +65,23 @@ def sstable_marking_test(self):
...
+        * Replace node3 with a new node
+        * Verify data integrity
+        # TODO: Several more verifications of data need to be interspersed throughout the test. The final assertion is insufficient.
```
mambocab: Yeah, this test isn't all that helpful; I'll understand if you don't address comments on this docstring until after the test gets redone.

repair_tests/incremental_repair_test.py
```diff
@@ -118,10 +142,24 @@ def multiple_repair_test(self):
...
+        * Start node1
+        * Issue an incremental repair
+        * Run sstablemetadata on both nodes again, pipe to a new file
+        * Verify repairs occurred and repairedAt was updated
+        """
         cluster = self.cluster
         cluster.populate(2).start()
         node1, node2 = cluster.nodelist()
-        node1.stress(['write', 'n=10000', '-schema', 'replication(factor=2)'])
+        node1.stress(['write', 'n=10K', '-schema', 'replication(factor=2)'])
```
mambocab: Should we add a `-rate` to this so the operation only runs once?

repair_tests/incremental_repair_test.py
```diff
@@ -135,8 +173,28 @@ def sstable_repairedset_test(self):
             node2.run_sstablemetadata(output_file=f, keyspace='keyspace1')
             node1.run_sstablemetadata(output_file=f, keyspace='keyspace1')
```
mambocab: I have the same suspicion here as I do above.

repair_tests/incremental_repair_test.py
```diff
@@ -170,12 +228,26 @@ def sstable_repairedset_test(self):
 
         self.assertGreaterEqual(max(matchcount), 2)
 
-        self.assertNotIn('repairedAt: 0', finaloutput)
+        self.assertNotIn('Repaired at: 0', finaloutput)
```
mambocab: We need a better assertion for this -- it'll be trivially true if the `Repaired at:` format changes. It's fine above:

```python
self.assertIn('Repaired at: 0', initialoutput)
```

Since that's a positive assertion that it's in there. Here, we can get false positives.

repair_tests/incremental_repair_test.py
```diff
@@ -118,10 +142,24 @@ def multiple_repair_test(self):
         assert_one(session, "SELECT COUNT(*) FROM ks.cf LIMIT 200", [149])
 
     def sstable_repairedset_test(self):
+        """
+        * Launch a two node cluster
+        * Insert data with stress
+        * Stop node2
+        * Run sstablerepairedset against node2
+        * Start node2
+        * Run sstablemetadata on both nodes, pipe to a file
```
mambocab: Should you mention the subsequent assertions here?

repair_tests/incremental_repair_test.py
```diff
@@ -269,10 +348,14 @@ def multiple_subsequent_repair_test(self):
     def sstable_marking_test_not_intersecting_all_ranges(self):
```
mambocab: This test has the same problem mentioned above; `self.assertNotIn('Repaired at: 0', output)` is asking for trouble.

repair_tests/repair_test.py
```diff
@@ -42,6 +50,11 @@ def check_rows_on_node(self, node_to_check, rows, found=None, missings=None, res
     @since('2.2.1')
     def no_anticompaction_after_dclocal_repair_test(self):
         """
+        * Launch a four node, two DC cluster
+        * Start a -local repair on node1 in dc1
+        * Assert that the dc1 nodes see repair messages
+        * Assert that the dc2 nodes do not see repair messages
+        * Assert no nodes anticompact
```
mambocab: Nit: anticompact_ed_

repair_tests/repair_test.py
```diff
@@ -62,6 +77,10 @@ def no_anticompaction_after_dclocal_repair_test(self):
     @since('2.2.1')
     def no_anticompaction_after_hostspecific_repair_test(self):
         """
+        * Launch a four node, two DC cluster
+        * Start a repair on all nodes, by enumerating with -hosts
+        * Assert all nodes see a repair messages
+        * Assert no nodes anticompact
```
mambocab: This assertion is implemented as `self.assertFalse(node.grep_log("Starting anticompaction"))`. Is there another way to get this information? As I mentioned earlier, I'm a little worried about negative assertions depending on string matching like this.

repair_tests/repair_test.py
```diff
@@ -78,6 +97,10 @@ def no_anticompaction_after_hostspecific_repair_test(self):
     @since('2.2.4')
     def no_anticompaction_after_subrange_repair_test(self):
         """
+        * Launch a four node, two DC cluster
```
mambocab: Copy-paste error; this test launches a 3-node cluster.

repair_tests/repair_test.py
```diff
@@ -78,6 +97,10 @@ def no_anticompaction_after_hostspecific_repair_test(self):
     @since('2.2.4')
     def no_anticompaction_after_subrange_repair_test(self):
```
mambocab: This test might also benefit from adding a `-rate` option to stress.

repair_tests/repair_test.py
```diff
@@ -144,6 +193,24 @@ def _repair_options(self, ks='', cf=None, sequential=True):
...
+        * Insert one thousand more rows at CL ALL
+        * Flush all nodes
+        * Check node3 only has 2000 keys
+        * Check node1 and node2 have 2001 keys
+        * Perform the repair type specified by the parent test
+        * Assert the appropriate messages are logged
+        * Assert node3 now has all data
+
+        @jira_ticket CASSANDRA-4373
+        """
```
mambocab: At the end of this test, do you have any idea what

```python
valid.remove((m.group(1), m.group(2)))
valid.remove((m.group(2), m.group(1)))
```

is for?

repair_tests/repair_test.py
```diff
@@ -210,7 +277,7 @@ def _empty_vs_gcable_no_repair(self, sequential):
         """
         Repairing empty partition and tombstoned partition older than gc grace
         should be treated as the same and no repair is necessary.
```
mambocab: Should this be fleshed out more?

repair_tests/repair_test.py
```diff
@@ -313,6 +385,11 @@ def local_dc_repair_test(self):
         self.check_rows_on_node(node2, 2001, found=[1000], restart=False)
 
     def dc_repair_test(self):
+        """
+        * Set up a multi DC cluster
+        * Perform a -dc repair on two dc's
+        * Assert only nodes on those dcs were repaired
```
mambocab: Should explain that this uses log checking

mambocab:
Most of my nits are of the form "please add this" or "consider adding this"; very few are "this is wrong". If you fix the "this is wrong" ones, I'm ok with the "plz add" ones being addressed in another PR. That might actually be preferable to letting this PR hang out forever.

repair_tests/incremental_repair_test.py
ptnapoleon: I can confirm it definitely appends

repair_tests/incremental_repair_test.py
ptnapoleon: Since they're in the same test, can't we just be sure that if the previous assertion passed, that this one is using the correct format for `Repaired at:`? It isn't going to change mid-test. I'd be fine breaking this string out into a variable.

repair_tests/incremental_repair_test.py
ptnapoleon: Could make the string discussed above a global? And have a smoke test in here that sanity checks that it is the current format.

repair_tests/repair_test.py
ptnapoleon: Checking on IRC, but maybe via sstablemetadata.

repair_tests/repair_test.py
ptnapoleon: Nope. I'll look

repair_tests/repair_test.py
ptnapoleon: Why?

repair_tests/incremental_repair_test.py
mambocab: Good insight there, but I don't like the fact that the check is incidental -- maybe pull out an `assert_` method that checks that `Repaired at:` is in there, but that the subsequent number isn't `0`. Using regexes like some of the code does elsewhere could work.

repair_tests/repair_test.py
mambocab: Eh, you're right, never mind.

mambocab:
+1 for the changes here, as long as there's a follow-up PR :+1:

PR #813, opened 2016-02-19
Add version check when creating second index on map column

beobal:
For CASSANDRA-11186

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -3712,8 +3712,14 @@ class TestCQL(UpgradeTester):
 
             # since 3.0, multiple indexes per-column are supported, so we can
             # create a value index even though we already have one on the keys
+            # note: creating the second index would be legal before upgrade if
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

ptnapoleon:
-1. The value of `self.UPGRADE_PATH.upgrade_version` can be in the form `git:cassandra-2.2`, and thus the comparison logic there doesn't work. You will need to first make sure it is in the right form.

beobal:
ah yes, are there util methods anywhere for normalizing versions that way? Or if not, a way to test the version of the specific session (rather than the cluster)?

mambocab:
Apologies for the really roundabout API, but here's an example:

https://github.com/riptano/cassandra-dtest/blob/master/upgrade_tests/cql_tests.py#L377

```
if self.get_node_version(is_upgraded) >= '2.2':
```

beobal:
Thanks Jim. I've made that change, rebased, squashed/merged & pushed to master.

mambocab:
Awesome, thank you :tada:

PR #815, opened 2016-02-22
Cassandra 6377

blerer:
Fix the test for the CASSANDRA-6377 patch and add extra tests

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -5,10 +5,11 @@ from assertions import assert_invalid
 from cassandra import ConsistencyLevel as CL
 from cassandra import InvalidRequest, ReadFailure, ReadTimeout
 from cassandra.policies import FallthroughRetryPolicy
-from cassandra.query import SimpleStatement, dict_factory, named_tuple_factory
+from cassandra.query import SimpleStatement, dict_factory, named_tuple_factory, tuple_factory
+from assertions import assert_invalid
 from datahelp import create_rows, flatten_into_set, parse_data_into_dicts
 from dtest import Tester, run_scenarios
-from tools import known_failure, since
+from tools import known_failure, rows_to_list, since
 
 
 class Page(object):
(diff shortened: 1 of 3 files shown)


secondary_indexes_test.py
```diff
@@ -476,11 +476,7 @@ def test_tuple_indexes(self):
 
         # no index present yet, make sure there's an error trying to query column
         stmt = ("SELECT * from simple_with_tuple where single_tuple = (1)")
-
-        if self.cluster.version() < "3":
-            assert_invalid(session, stmt, 'No secondary indexes on the restricted columns support the provided operators')
-        else:
-            assert_invalid(session, stmt, 'No supported secondary index found for the non primary key columns restrictions')
+        assert_invalid(session, stmt)
```
mambocab: In [your comment on JIRA](https://issues.apache.org/jira/browse/CASSANDRA-6377?focusedCommentId=15157160&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15157160) you note that this is checked in the unit tests, so I can see why you removed these. However, we also want to make sure the statement isn't failing for some unrelated reason. Is there a code you can check for? Unless that's available to us, checking for these messages needs to stay.

mambocab:
I'm seeing the following failure on blerer/6377-3.0: https://gist.github.com/mambocab/0f7e48e84c02ee5e8dda

Any thoughts on this @blerer?

paging_test.py
```diff
@@ -928,6 +928,155 @@ def static_columns_with_empty_non_static_columns_paging_test(self):
...
+                                   [2, 2],
+                                   [3, 3],
+                                   [4, 4]])
+
+            # Multi-partitions query with DISTINCT and LIMIT
+            res = rows_to_list(session.execute("SELECT DISTINCT a, s FROM test WHERE a IN (0, 1, 2, 3, 4) AND s >= 1 LIMIT 2 ALLOW FILTERING"))
+            self.assertEqual(res, [[1, 1],
+                                   [2, 2]])
+
+
```
mambocab: This makes 3 blank lines before the next class; there should be only 2. This seems to be what the linter is flagging.

secondary_indexes_test.py
blerer: For 2 error messages that I changed, I had to fix 14 unit test methods and 5 Dtests. I had to trigger 1 extra run for the Dtests do discover the validation checks and another one as I had missed one of the test.
I am really in favor of having a lot of tests but duplicating this type of checks does not bring more safety. Only frustration for the person that has to modify the error messages. :-(

blerer:
I did not had that error on my CI branches. Sorry, I have no clue from where it can come.

secondary_indexes_test.py
mambocab: Yeah, there's definitely a bunch of unnecessary duplication here -- once we merge this I'll pull out this logic so there isn't so much explicit version-checking. Sorry for the frustration about that :/

I realized I may have misread you -- do these tests duplicate unit tests? If so, we should remove them.

If we keep these tests, however, we need to continue checking that the error messages contain what we expect. The purpose is not to assert that the error contains the correct message -- the purpose is to assert that the erroneous `SELECT` raises the correct error. If we don't check the message, a (for example) syntax error raised by the statement will cause the test to pass, even though that behavior is incorrect.

mambocab:
For whatever reason, I'm not seeing that error anymore. Not sure what was happening before, but it doesn't seem to be a problem.

Thanks for addressing the style complaint. If you could remove the changes to the `assert_invalid` calls and squash into a single commit, I'll merge. Sorry again for the frustration with the code duplication.

secondary_indexes_test.py
blerer: > I realized I may have misread you -- do these tests duplicate unit tests? If so, we should remove them.

The error message checks are already performed in the unit tests. It is why I consider them as duplicate. Not because they are done at several places in the DTests.

It is clear that we need to check the error messages. Doing it in the unit tests should be enough for those ones as the checks are performed on the coordinator.

mambocab:
I think I understand now -- sorry, I didn't understand that your changes for 6377 introduced new messages and made existing tests fail.

I still think that removing the message-checking logic could introduce false positives in these tests, and I won't merge a change that does so. I don't like making this more complicated for you when you don't think it's necessary, but it seems we disagree about whether or not making these checks more lax is acceptable.

Would rebasing on top of this:

https://github.com/mambocab/cassandra-dtest/commit/ac536f9433906415a30fdb34c4b8629c0b316ad3

and adding another message to `assertions.assert_invalid_with_no_secondary_index` address the problems you're having? I think you'd wind up with something like:

```python
post_cassandra_6377 = ('2.2.5' < version < '3') or ('3' < version < '3.0.4') or ('3.4' < version)
if post_cassandra_6377:
    expected_msg = '<Post-6377 2i message>'
elif version < '3':
    expected_msg = 'No secondary indexes on the restricted columns support the provided operators'
else:
    expected_msg = 'No supported secondary index found for the non primary key columns restrictions'
```

I've tested my changes on the affected tests:

```
secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_list_indexes
secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_map_indexes
secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_set_indexes
secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_tuple_indexes
user_types_test.py:TestUserTypes.test_type_secondary_indexing
```

locally on `trunk` and `cassandra-2.2`, and they all pass.

blerer:
Unfortunatly, the problem is a bit more complex than that.
1. The patch for 6377 introduces a new message for 2.2 and and another one for 3.x (which means that we will have to support 4 different type of messages).
2. Some of those message depends on the query as the name of the column is specified in it (for example: "Predicates on non-primary-key columns (name) are not yet supported for non secondary index queries"). By consequence, your change does not really help.

The advantage of testing the coordinator error handling in unit tests is that each C* version has its own set of tests and that by consequence there is not need for version checks.

I am not removing tests. I have just moved some tests in the unit tests.

Have a look there: https://github.com/apache/cassandra/compare/trunk...blerer:6377-2.2#diff-7f5981228f9d9428fb164aa91316aa85R1348
If you believe that some unit tests are missing I am perfectly fine in adding them.

mambocab:
I'm not sure I'm making my objection clear -- my objection is that, when a statement is executed in the dtest and fails, a bare `assert_invalid` doesn't verify that the failure happened because of a 2i-related problem, only that the underlying error generated an `InvalidRequest` exception. Unless the unit tests logically duplicate the dtests, they can't verify that failures generated in the dtests happened for the correct reason; that has to happen in the dtest itself, as far as I can tell.

In any event, does the improved assertion from this commit work?:

https://github.com/mambocab/cassandra-dtest/commit/e038bdfa753a1537365ef311d07b8c113ea9ed87

It seems to work locally; I ran the following tests:

```
tests="nosetests secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_list_indexes secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_map_indexes secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_set_indexes secondary_indexes_test.py:TestSecondaryIndexesOnCollections.test_tuple_indexes user_types_test.py:TestUserTypes.test_type_secondary_indexing"

CASSANDRA_VERSION=github:blerer/6377-2.2 $tests --with-id --verbosity=2 --stop --nocapture
CASSANDRA_VERSION=github:blerer/6377-3.0 nosetests $tests --with-id --verbosity=2 --stop --nocapture
CASSANDRA_VERSION=github:blerer/6377-trunk nosetests $tests --with-id --verbosity=2 --stop --nocapture
```

And they passed, so they seem to work with the new messages introduced in CASSANDRA-6377.

mambocab:
Discussed off-GitHub -- the `assert_invalid` checks in question can be removed entirely. While the rest of each of these tests may be useful, those particular parts of the tests duplicate assertions that, since they only happen on the local coordinator, are more appropriate for the unit tests, and are already implemented there. The confusion there was all mine. In our discussion, @blerer agreed to change the PR to remove the `assert_invalid` checks entirely. :+1:

After that, @blerer, you should feel free to merge once CASSANDRA-6377 is closed.

mambocab:
6377 was merged to trunk, and this new test now passes on trunk. Merging, thanks a lot @blerer

PR #816, opened 2016-02-22
cut known_failure from index_summary_upgrade_test

mambocab:
@mshuler Could you have a look at this? If I understand correctly, #801 should  have fixed [CASSANDRA-11127](https://issues.apache.org/jira/browse/CASSANDRA-11127).

With the changes from #801, this test passes locally. There is a [flap on CassCI](http://cassci.datastax.com/job/trunk_dtest/1014/testReport/junit/index_summary_upgrade_test/TestUpgradeIndexSummary/test_upgrade_index_summary/) but it seems unrelated/looks like another flap I've seen elsewhere. Otherwise [it looks pretty clean the past couple weeks](http://cassci.datastax.com/job/trunk_dtest/1014/testReport/junit/index_summary_upgrade_test/TestUpgradeIndexSummary/test_upgrade_index_summary/history/).

If this is good to go, then #801 obviated the need for #775 as well, and we should close that without merge.

### DIFF
--- a/index_summary_upgrade_test.py
+++ b/index_summary_upgrade_test.py
@@ -2,13 +2,10 @@ from cassandra.concurrent import execute_concurrent_with_args
 
 from dtest import OFFHEAP_MEMTABLES, Tester, debug
 from jmxutils import JolokiaAgent, make_mbean, remove_perf_disable_shared_mem
-from tools import known_failure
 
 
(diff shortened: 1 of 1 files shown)


mambocab:
Never mind; I didn't read the JIRA ticket thoroughly enough to see that it was talking exclusively about jobs with vnodes disabled.

PR #822, opened 2016-02-23
Cql tests cleanup

mambocab:
Looking for feedback on TODOs in docstrings and (in one or 2 places) comments. I'm going to avoid expanding the scope of this PR if I can -- dealing with the TODOs will happen in a followup PR. @ptnapoleon, got time for this?

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -688,9 +688,9 @@ class AbortedQueriesTester(CQLTester):
             session.execute("INSERT INTO test3 (id, col, val) VALUES ({}, {}, 'foo')".format(i, i // 10))
 
         mark = node.mark_log()
-        statement = session.prepare("SELECT * from test3 WHERE col < ? ALLOW FILTERING",
-                                    consistency_level=ConsistencyLevel.ONE,
-                                    retry_policy=FallthroughRetryPolicy())
+        statement = session.prepare("SELECT * from test3 WHERE col < ? ALLOW FILTERING")
+        statement.consistency_level = ConsistencyLevel.ONE
+        statement.retry_policy = FallthroughRetryPolicy()
         assert_unavailable(lambda c: debug(c.execute(statement, [50])), session)
         node.watch_log_for("operations timed out", from_mark=mark, timeout=60)
 
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -57,13 +57,24 @@ class StorageProxyCQLTester(CQLTester):
     """
     Each CQL statement is exercised at least once in order to
     ensure we execute the code path in StorageProxy.
+    # TODO This probably isn't true anymore?
     Note that in depth CQL validation is done in Java unit tests,
     see CASSANDRA-9160.
+
+    # TODO I'm not convinced we need these. Seems like all the functionality
+    #      is covered in greater detail in other test classes.
     """
 
     def keyspace_test(self):
         """
-        CREATE KEYSPACE, USE KEYSPACE, ALTER KEYSPACE, DROP KEYSPACE statements
+        Smoke test that basic keyspace operations work:
+
+        - create a keyspace
+        - USE that keyspace
+        - ALTER it
+        - #TODO: assert the ALTER worked
+        - DROP it
+        - attempt to USE it again, asserting it raises an InvalidRequest exception
         """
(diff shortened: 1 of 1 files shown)


cql_tests.py
```diff
@@ -113,18 +143,26 @@ def table_test(self):
...
         session.execute("CREATE INDEX testidx ON test3 (v1)")
 
-        for i in xrange(0, 10):
-            session.execute("INSERT INTO test3 (k, v1, v2) VALUES (%d, %d, %d)" % (i, i, i))
+        for i in range(0, 10):
+            session.execute("INSERT INTO test3 (k, v1, v2) VALUES ({i}, {i}, {i})".format(i=i))
 
         res = session.execute("SELECT * FROM test3 WHERE v1 = 0")
-        assert rows_to_list(res) == [[0, 0, 0]], res
+        self.assertEqual(rows_to_list(res), [[0, 0, 0]])
```
ptnapoleon: This, and others, are good candidates for being replaced with the assertions from assertions.py

cql_tests.py
```diff
@@ -322,6 +416,12 @@ def rename_test(self):
     def invalid_string_literals_test(self):
         """
         @jira_ticket CASSANDRA-8101
+
+        - assert INSERTing into a nonexistent table fails
+        # TODO why?
```
ptnapoleon: Why would it not fail?

cql_tests.py
```diff
@@ -374,7 +488,17 @@ def prepared_statement_invalidation_test(self):
...
+        """
+        Regression test for CASSANDRA-1337:
+
+        - CREATE a table
+        - INSERT 2 rows
+        - SELECT * from the table
+        - assert 2 rows were returned
+
+        @jira_ticket CASSANDRA-1337
+        # TODO I don't see how this is an interesting test or how it tests 1337.
```
ptnapoleon: Well it's a range scan of a small table, certainly.

ptnapoleon:
Other than the asserts, nothing sticks out in a skim. I'll go commit by commit next.

cql_tests.py
mambocab: I wasn't clear -- I understand why it should fail, but not why we have that assertion.

cql_tests.py
```diff
@@ -322,6 +416,12 @@ def rename_test(self):
     def invalid_string_literals_test(self):
         """
         @jira_ticket CASSANDRA-8101
+
+        - assert INSERTing into a nonexistent table fails
+        # TODO why?
+        - create a table with ascii and text columns
+        - assert that trying to execute an insert statement with non-UTF8 contents raises a ProtocolException
+            - tries to insert into a nonexistent column to make sure the ProtocolException is raised over other errors
```
ptnapoleon: Looks like you tabbed too far here.

cql_tests.py
ptnapoleon: Are we maybe checking that Invalid is thrown instead of Protocol?

cql_tests.py
```diff
@@ -431,17 +573,33 @@ def local_query_test(self):
...
+
+        - populate the cluster with 2 nodes
+        - set a 1-second read timeout
+        - start one node without having it join the ring
+        - start the other node with read_iteration_delay set to 1.5 seconds
+            - (this will cause read queries to take longer than the read timeout)
+        - CREATE a table
+        - INSERT 5000 rows on a session on the node that is not a member of the ring
+        - run SELECT statements and assert they fail
+        # TODO why do queries fail? idgi
```
ptnapoleon: Not sure what you mean. We expect them to time out because the delay is longer than the timeout

cql_tests.py
mambocab: No, this is intentionally a nested bullet.

cql_tests.py
mambocab: Yep, that'd do it. I think this is a stale comment, I figured it out eventually.

ptnapoleon:
I am +1

cql_tests.py
mambocab: Is it worth keeping around? I don't get it.

mambocab:
Ok, all todos are encoded as `TODO`s in the docstrings. I'm gonna merge and then work on the todos.

PR #823, opened 2016-02-23
mark test failure

knifewine:


mambocab:
Looks like the `isort` was done incorrectly; `nose` should be imported above the local imports.

mambocab:
+1 on the rest, though.

knifewine:
fixed isort thing

PR #825, opened 2016-02-24
Test for CASSANDRA-10392

tjake:


cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+        node1 = cluster.nodelist()[0]
+
+        session = self.patient_cql_connection(node1, protocol_version=protocol_version)
+        if create_keyspace:
+            if self._preserve_cluster:
+                session.execute("DROP KEYSPACE IF EXISTS ks")
+            self.create_ks(session, 'ks', rf)
+        return session
+
+    def run_cqlsh(self, node, cmds, cqlsh_options=[]):
```
mambocab: `ccmlib.node.Node` objects already implement this as a method; could you please use that?

cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+from ccmlib import common
+from dtest import Tester, debug
+
+
+class TestCqlTracing(Tester):
+
+    def prepare(self, create_keyspace=True, nodes=3, rf=3, protocol_version=3, jvm_args=[], **kwargs):
+        cluster = self.cluster
+        cluster.populate(nodes).start(wait_for_binary_proto=True, jvm_args=jvm_args)
+        cluster.nodelist()[0].watch_log_for(' listening ')
```
mambocab: I believe this is redundant with `wait_for_binary_proto=True` in the `start()` args.

ptnapoleon: Yes, it is.

cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+import os
+import subprocess
+import sys
+from ccmlib import common
+from dtest import Tester, debug
+
+
+class TestCqlTracing(Tester):
+
+    def prepare(self, create_keyspace=True, nodes=3, rf=3, protocol_version=3, jvm_args=[], **kwargs):
```
mambocab: Could you please use the following convention for mutable function arguments?:

```python
def prepare(jvm_args=None):
    if jvm_args is None:
        jvm_args = []
```

Mutable default arguments are instantiated at function-definition time, and changes to it will persist across calls to the function. See [this guie](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) for a more thorough explanation.

cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+        port = node.network_interfaces['binary'][1]
+        args = cqlsh_options + [host, str(port)]
+        sys.stdout.flush()
+        p = subprocess.Popen([cli] + args, env=env, stdin=subprocess.PIPE, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+        for cmd in cmds.split(';'):
+            p.stdin.write(cmd + ';\n')
+        p.stdin.write("quit;\n")
+        return p.communicate()
+
+    def trace(self, cursor):
```
mambocab: We favor `session` over `cursor` in dtest (see [CONTRIBUTING](https://github.com/riptano/cassandra-dtest/blob/master/CONTRIBUTING.md#style)). Could you please change that in this method?

mambocab: Could you also add a docstring explaining each step of the test? See, e.g., @ptnapoleon's docstrings here:

https://github.com/riptano/cassandra-dtest/pull/811/files#diff-96b58282ba6bb7a35c82b35685ccb710R123

cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+        out, err = self.run_cqlsh(node1, 'TRACING ON')
+        self.assertIn('Tracing is enabled', out)
+
+        out, err = self.run_cqlsh(node1, 'TRACING ON; SELECT * from system.peers')
+        self.assertIn('Tracing session: ', out)
+        self.assertIn('Request complete ', out)
+
+        # Inserts
+        out, err = self.run_cqlsh(node1, "CONSISTENCY ALL; INSERT INTO ks.users (userid, firstname, lastname, age) VALUES (550e8400-e29b-41d4-a716-446655440000, 'Frodo', 'Baggins', 32)")
```
mambocab: Do we not want any assertions about the output of tracing here?

tjake:
I posted this on behalf of the author. I'll do my best to incorporate your suggested changes...

cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertIn('Request complete ', out)
+        self.assertIn(" Frodo |  Baggins", out)
+
+    def tracing_simple_test(self):
+        """ Test Simple Tracing """
+        cursor = self.prepare()
+        self.trace(cursor)
+
+    def tracing_unknown_impl_test(self):
+        """ Test unknown Tracing class """
```
mambocab: It's not clear to me why this and the test below pass. I guess because it falls back to using the default tracing class? If that's the case, could you document that?

cql_tracing_test.py
```diff
@@ -0,0 +1,90 @@
...
+        cursor = self.prepare(jvm_args=['-Dcassandra.custom_tracing_class=junk'])
+        self.ignore_log_patterns = ["Cannot use class junk for tracing"]
```
mambocab: Could you `grep_log_for_errors` for this, and in the test below? Seems like that's part of C*'s correct behavior here.

mambocab:
Discussed off GH: I'll make the changes I've requested since tjake isn't the original author.

mambocab:
@tjake It'd be nice to test an example of `custom_tracing_class` being successfully specified. Is there anywhere, maybe in the `test` package somewhere, where we could grab an implementation of `Tracing` that we can actually instantiate?

tjake:
There is one in the unit tests https://github.com/apache/cassandra/commit/bf25e668f416b1c279986d1b23fee2a0192d8022#diff-407fc187162f715ffe8a71877abd5244R138 and there is an impl for zipkin https://github.com/thelastpickle/cassandra-zipkin-tracing
But I'm not sure if the dtests could use it? Maybe write one that appends to a local file?

mambocab:
@tjake Thanks; we don't currently have anything set up to allow us to define Java classes ourselves, but I've added a TODO. Could you please review, or find someone to review, that this correctly tests the new behavior we expect after my changes?

@ptnapoleon Could you please review this for style?

ptnapoleon:
You forgot to mention the jira ticket in the docstrings.

ptnapoleon:
Otherwise, lgtm

mambocab:
@ptnapoleon Fixed. Thanks.

tjake:
Ship it!

mambocab:
Thanks all! Merging.

PR #826, opened 2016-02-24
Add additional repair coverage

ptnapoleon:
Made one bug fix, one style fix, and added three tests

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -280,8 +280,8 @@ class TestRepair(Tester):
         for line, m in out_of_sync_logs:
             self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
             self.assertIn({m.group(1), m.group(2)}, valid_out_of_sync_pairs, str((m.group(1), m.group(2))))
-            valid_out_of_sync_pairs.remove((m.group(1), m.group(2)))
-            valid_out_of_sync_pairs.remove((m.group(2), m.group(1)))
+            valid_out_of_sync_pairs.remove({m.group(1), m.group(2)})
+            valid_out_of_sync_pairs.remove({m.group(2), m.group(1)})
 
         # Check node3 now has the key
         self.check_rows_on_node(node3, 2001, found=[1000], restart=False)
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -9,6 +9,34 @@ from dtest import Tester, debug
 from tools import insert_c1c2, known_failure, no_vnodes, query_c1c2, since
 
 
+def _repair_options(version, ks='', cf=None, sequential=True):
+    """
+    Function for assembling appropriate repair CLI options,
+    based on C* version, as defaults have changed.
+    @param ks The keyspace to repair
+    @param cf The table to repair
+    @param sequential If the repair should be a sequential repair [vs parallel]
+    """
+    opts = []
+
+    # since version 2.2, default is parallel, otherwise it's sequential
+    if sequential:
+        if version >= '2.2':
+            opts += ['-seq']
+    else:
+        if version < '2.2':
+            opts += ['-par']
+
+    # test with full repair
+    if version >= '2.2':
+        opts += ['-full']
+    if ks:
+        opts += [ks]
+    if cf:
+        opts += [cf]
+    return opts
+
+
 class TestRepair(Tester):
 
     def check_rows_on_node(self, node_to_check, rows, found=None, missings=None, restart=True):
@@ -174,35 +202,6 @@ class TestRepair(Tester):
         """
         self._simple_repair(order_preserving_partitioner=True)
 
-    def _repair_options(self, ks='', cf=None, sequential=True):
-        """
-        Function for assembling appropriate repair CLI options,
-        based on C* version, as defaults have changed.
-        @param ks The keyspace to repair
-        @param cf The table(s) to repair
-        @param sequential If the repair should be a sequential repair [vs parallel]
-        """
-        if cf is None:
-            cf = []
-        opts = []
-        version = self.cluster.version()
-        # since version 2.2, default is parallel, otherwise it's sequential
-        if sequential:
-            if version >= '2.2':
-                opts += ['-seq']
-        else:
-            if version < '2.2':
-                opts += ['-par']
-
-        # test with full repair
-        if version >= '2.2':
-            opts += ['-full']
-        if ks:
-            opts += [ks]
-        if cf:
-            opts += [cf]
-        return opts
-
     def _simple_repair(self, order_preserving_partitioner=False, sequential=True):
         """
         * Configure a three node cluster to not use hinted handoff, and to use batch commitlog
@@ -265,7 +264,7 @@ class TestRepair(Tester):
         # Run repair
         start = time.time()
         debug("starting repair...")
-        node1.repair(self._repair_options(ks='ks', sequential=sequential))
+        node1.repair(_repair_options(self.cluster.version(), ks='ks', sequential=sequential))
         debug("Repair time: {end}".format(end=time.time() - start))
 
         # Validate that only one range was transfered
@@ -273,13 +272,16 @@ class TestRepair(Tester):
 
         self.assertEqual(len(out_of_sync_logs), 2, "Lines matching: " + str([elt[0] for elt in out_of_sync_logs]))
 
-        valid = [(node1.address(), node3.address()), (node3.address(), node1.address()), (node2.address(), node3.address()), (node3.address(), node2.address())]
+        valid_out_of_sync_pairs = [{node1.address(), node3.address()},
+                                   {node3.address(), node1.address()},
+                                   {node2.address(), node3.address()},
+                                   {node3.address(), node2.address()}]
 
         for line, m in out_of_sync_logs:
-            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
-            self.assertIn((m.group(1), m.group(2)), valid, str((m.group(1), m.group(2))))
-            valid.remove((m.group(1), m.group(2)))
-            valid.remove((m.group(2), m.group(1)))
+            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+            self.assertIn({m.group(1), m.group(2)}, valid_out_of_sync_pairs, str((m.group(1), m.group(2))))
+            valid_out_of_sync_pairs.remove((m.group(1), m.group(2)))
+            valid_out_of_sync_pairs.remove((m.group(2), m.group(1)))
 
         # Check node3 now has the key
         self.check_rows_on_node(node3, 2001, found=[1000], restart=False)
@@ -352,7 +354,7 @@ class TestRepair(Tester):
 
         # bring up node2 and repair
         node2.start(wait_for_binary_proto=True, wait_other_notice=True)
-        node2.repair(self._repair_options(ks='ks', sequential=sequential))
+        node2.repair(_repair_options(self.cluster.version(), ks='ks', sequential=sequential))
 
         # check no rows will be returned
         for cf in ['cf1', 'cf2']:
@@ -383,18 +385,18 @@ class TestRepair(Tester):
 
         debug("starting repair...")
         opts = ["-local"]
-        opts += self._repair_options(ks="ks")
+        opts += _repair_options(self.cluster.version(), ks="ks")
         node1.repair(opts)
 
         # Verify that only nodes in dc1 are involved in repair
         out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
         self.assertEqual(len(out_of_sync_logs), 1, "Lines matching: {}".format(len(out_of_sync_logs)))
         line, m = out_of_sync_logs[0]
-        self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
-        valid = [node1.address(), node2.address()]
-        self.assertIn(m.group(1), valid, "Unrelated node found in local repair: " + m.group(1))
-        valid.remove(m.group(1))
-        self.assertIn(m.group(2), valid, "Unrelated node found in local repair: " + m.group(2))
+        self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+        valid_out_of_sync_pairs = [node1.address(), node2.address()]
+        self.assertIn(m.group(1), valid_out_of_sync_pairs, "Unrelated node found in local repair: " + m.group(1))
+        valid_out_of_sync_pairs.remove(m.group(1))
+        self.assertIn(m.group(2), valid_out_of_sync_pairs, "Unrelated node found in local repair: " + m.group(2))
         # Check node2 now has the key
         self.check_rows_on_node(node2, 2001, found=[1000], restart=False)
 
@@ -411,17 +413,19 @@ class TestRepair(Tester):
 
         debug("starting repair...")
         opts = ["-dc", "dc1", "-dc", "dc2"]
-        opts += self._repair_options(ks="ks")
+        opts += _repair_options(self.cluster.version(), ks="ks")
         node1.repair(opts)
 
         # Verify that only nodes in dc1 and dc2 are involved in repair
         out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
         self.assertEqual(len(out_of_sync_logs), 2, "Lines matching: " + str([elt[0] for elt in out_of_sync_logs]))
-        valid = [(node1.address(), node2.address()), (node2.address(), node1.address()),
-                 (node2.address(), node3.address()), (node3.address(), node2.address())]
+        valid_out_of_sync_pairs = [{node1.address(), node2.address()},
+                                   {node2.address(), node1.address()},
+                                   {node2.address(), node3.address()},
+                                   {node3.address(), node2.address()}]
         for line, m in out_of_sync_logs:
-            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
-            self.assertIn((m.group(1), m.group(2)), valid, str((m.group(1), m.group(2))))
+            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+            self.assertIn({m.group(1), m.group(2)}, valid_out_of_sync_pairs, str((m.group(1), m.group(2))))
 
         # Check node2 now has the key
         self.check_rows_on_node(node2, 2001, found=[1000], restart=False)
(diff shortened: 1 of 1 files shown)


repair_tests/repair_test.py
```diff
@@ -415,6 +415,37 @@ def dc_repair_test(self):
...
+        * Assert only nodes on those dcs were repaired
+        """
+        cluster = self._setup_multi_dc()
+        node1 = cluster.nodes["node1"]
+        node2 = cluster.nodes["node2"]
+        node3 = cluster.nodes["node3"]
+
+        debug("starting repair...")
+        opts = ["-dc", "dc1", "-dc", "dc2", "-dcpar"]
+        opts += self._repair_options(ks="ks", sequential=False)
```
mambocab: Refactoring opportunity: `_repair_options` should be a bare function that takes a version. It doesn't need to be a method in the first place and can be inaccurate on mixed-version clusters since it uses `cluster.version()`. This may actually be something to add to `ccmlib.node.Node.repair`.

repair_tests/repair_test.py
```diff
@@ -415,6 +415,37 @@ def dc_repair_test(self):
...
+        node1.repair(opts)
+
+        # Verify that only nodes in dc1 and dc2 are involved in repair
+        out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
+        self.assertEqual(len(out_of_sync_logs), 2, "Lines matching: " + str([elt[0] for elt in out_of_sync_logs]))
+        valid = [(node1.address(), node2.address()), (node2.address(), node1.address()),
```
mambocab: `valid` what? Could you either give this a more descriptive name or add a comment? I think it'd be more readable with a line break between each tuple as well. Lotta parens in here.

repair_tests/repair_test.py
```diff
@@ -415,6 +415,37 @@ def dc_repair_test(self):
...
+                 (node2.address(), node3.address()), (node3.address(), node2.address())]
+        for line, m in out_of_sync_logs:
```
mambocab: I don't like the "access random elements of the match object" thing here -- takes some serious tracing/reading to figure out what's being compared. My gut says to construct some kind of lil data structure with named fields from `out_of_sync_logs`, so, like

```python
OutOfSyncLogData = namedtuple('OutOfSyncLogData', ['num_ranges', 'nodes'])
out_of_sync_nodes = [(line, OutOfSyncLogData(num_ranges=m.group(3), nodes={m.group(1), m.group(2)}),
                     for line, m in node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")]
```

repair_tests/repair_test.py
mambocab: Looking at how this is used, this should be something like `valid_out_of_sync_pairs = [{node1.address(), node2.address()}, {node2.address(), node3.address()}]`. These are unordered pairs of nodes, so sets are more appropriate than tuples.

repair_tests/repair_test.py
```diff
@@ -415,6 +415,37 @@ def dc_repair_test(self):
...
+            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
+            self.assertIn((m.group(1), m.group(2)), valid, str((m.group(1), m.group(2))))
```
mambocab: The error message could be a little better here, or it might not be necessary with a data structure with a self-describing `repr` like the `OutOfSyncLogData` described above.

mambocab: Ah, never mind, you'll still need it you're not passing the full data structure to the `.assert` method. Anyway, a more verbose message here would be good.

repair_tests/repair_test.py
```diff
@@ -454,6 +485,122 @@ def _setup_multi_dc(self):
...
+    def token_range_repair_test(self):
+        """
+        Test repair using the -st and -et options
+        * Launch a three node cluster
+        * Insert some data at RF 2
+        * Shut down node2, insert more data, restore node2
+        * Issue a repair on a range that only belongs to node1
+        * Verify that nodes 1 and 2, and only nodes 1+2, are repaired
+        """
+        cluster = self.cluster
```
mambocab: Why not add this preparation code to the `_range_repair` method? It looks like it's repeated across the methods that use it.

repair_tests/repair_test.py
```diff
@@ -454,6 +485,122 @@ def _setup_multi_dc(self):
...
+        cluster.set_configuration_options(values={'hinted_handoff_enabled': False}, batch_commitlog=True)
+        debug("Starting cluster..")
+        cluster.populate(3).start(wait_for_binary_proto=True)
+
+        node1, node2, node3 = cluster.nodelist()
+
+        self._range_repair(repair_opts=['-pr'])
+
+    def _range_repair(self, repair_opts):
```
mambocab: This should be renamed to make it clearer that it basically is a parameterized test; the current name makes it look like a utility method.

repair_tests/repair_test.py
```diff
@@ -454,6 +485,122 @@ def _setup_multi_dc(self):
...
+        job_thread_count = '2'
+        opts = ['-tr', '-j', job_thread_count]
+        opts += self._repair_options(ks='keyspace1', cf='standard1', sequential=False)
+        node1.repair(opts)
+
+        time.sleep(5)  # Give the trace table some time to populate
+
+        session = self.patient_cql_connection(node1)
+        rows = list(session.execute("SELECT activity FROM system_traces.events"))
+        self.assertGreater(len(rows), 2500, "We expected more trace data from repair. Found {} rows.".format(len(rows)))
```
mambocab: This weirds me out. Could something else populate the `events` table? If that's not the case now, is there anything preventing devs from making something else populate the `events` table in the future? Why do you expect 2500 or more? This just doesn't seem like a very strong or robust condition to verify.

repair_tests/repair_test.py
```diff
@@ -454,6 +485,122 @@ def _setup_multi_dc(self):
...
+        self.assertIn('job threads: {}'.format(job_thread_count),
+                      rows[0][0],
```
mambocab: How do you know that `rows[0][0]` will contain this information? Are there any guarantees about that? Same question as above -- if the devs add some feature that populates `events`, will your assumptions still hold here?

repair_tests/repair_test.py
```diff
@@ -454,6 +485,122 @@ def _setup_multi_dc(self):
...
+        node2.stop(wait_other_notice=True)
+
+        node1.stress(['write', 'n=20K', 'cl=ONE', '-schema', 'replication(factor=2)', '-rate', 'threads=30', '-pop', 'seq=20..40K'])
+        node2.start(wait_for_binary_proto=True, wait_other_notice=True)
+
+        node1.stress(['write', 'n=20K', 'cl=ALL', '-schema', 'replication(factor=2)', '-rate', 'threads=30', '-pop', 'seq=40..60K'])
+
+        cluster.flush()
+
+        job_thread_count = '2'
```
mambocab: Do we run any other tests to make sure the `-j` flag is respected? Might be worth adding a test for, or noting in Michael's spreadsheet.

mambocab:
Done reviewing. The test logic looks ok, though I haven't run the tests yet.

repair_tests/repair_test.py
```diff
@@ -454,6 +485,122 @@ def _setup_multi_dc(self):
...
+        # Repair only the range node 1 owns
+        opts = repair_opts
+        opts += self._repair_options(ks='keyspace1', cf='standard1', sequential=False)
+        node1.repair(opts)
+
+        self.assertEqual(len(node1.grep_log('are consistent for standard1')), 0, "Nodes 1 and 2 should not be consistent.")
+        self.assertEqual(len(node1.grep_log("/127.0.0.2 and /127.0.0.1 have ([0-9]+) range\(s\) out of sync")), 1)
+        self.assertEqual(len(node3.grep_log('Repair command')), 0, "Node 3 should not have been involved in the repair.")
+
+    def trace_threads_repair(self):
```
mambocab: Is this supposed to be a `_test` method?

repair_tests/repair_test.py
ptnapoleon: We discussed this already. I need to start the cluster in the main test so that in `token_range_repair_test` I can get the node's tokens, to pass into `_range_repair`

mambocab: Ah cool. And looks like I already forgot that conversation :grimacing:

mambocab:
Still running tests, but I got the following failure with vnodes on HEAD on 2.1:

```
FAIL: repair_tests/repair_test.py:TestRepair.dc_parallel_repair_test
  File "/home/mambocab/cstar_src/cassandra-dtest/repair_tests/repair_test.py", line 447, in dc_parallel_repair_test
    self.assertEqual(len(node1.grep_log('parallelism: dc_parallel')), 1)
AssertionError: 0 != 1
```

mambocab:
Seeing the following failure on `trunk` with `DISABLE_VNODES=true` 

```
FAIL: repair_tests/repair_test.py:TestRepair.trace_threads_repair_test
  File "/home/mambocab/cstar_src/cassandra-dtest/repair_tests/repair_test.py", line 598, in trace_threads_repair_test
    self.assertGreater(len(rows), 2500, "We expected more trace data from repair. Found {} rows.".format(len(rows)))
AssertionError: We expected more trace data from repair. Found 555 rows.
-------------------- >> begin captured logging << --------------------
```

repair_tests/repair_test.py
ptnapoleon: Is this not the test we check for that in? I was double dipping here

repair_tests/repair_test.py
ptnapoleon: I know that `rows[X][0]` contains what I want, because I'm only selecting the relevant column, `activity'`. And I know that `X = 0` because the first thing to be logged is the "Repair triggered with these options" message.

repair_tests/repair_test.py
ptnapoleon: I could compare line counts? I'm not sure how else to check the repair table grew, other than that.

repair_tests/repair_test.py
ptnapoleon: I disagree on putting it in CCM, but will pull it out for easier future use by mixed version clusters.

repair_tests/repair_test.py
mambocab: My objection is not "find a different way to assert that `system_traces.events` grew" -- it's "can we trust that `system_traces.events` only grows because of repair events?".

And one of the failures I commented with earlier (the one on `trunk` with `DISABLE_VNODES=true`) fails because repair didn't generate enough messages.

repair_tests/repair_test.py
mambocab: a) I'm not much for double-dipping. b) this only uses one case (where you're passing `-j 2`), so it can't repudiate, e.g., this assertion: `the -j flag is a noop and the number of threads is hardcoded to 2`.

ptnapoleon:
Okay, ready for re-review.

repair_tests/repair_test.py
ptnapoleon: Will extract then.

repair_tests/repair_test.py
ptnapoleon: Right, thus why I pulled the check. I cannot trust that system_traces.events only grew because of repair, because you shouldn't trust anything. The rows[0][0] check at least asserts that system_traces.events WAS written to by repair? And asserting that the table grew is better than nothing.

repair_tests/repair_test.py
```diff
@@ -265,13 +264,16 @@ def _simple_repair(self, order_preserving_partitioner=False, sequential=True):
 
         self.assertEqual(len(out_of_sync_logs), 2, "Lines matching: " + str([elt[0] for elt in out_of_sync_logs]))
 
-        valid = [(node1.address(), node3.address()), (node3.address(), node1.address()), (node2.address(), node3.address()), (node3.address(), node2.address())]
+        valid_out_of_sync_pairs = [{node1.address(), node3.address()},
```
mambocab: `{node1.address(), node3.address()}` is redundant with `{node3.address(), node1.address()}`, and the same goes for the 2/3 ones below. That was part of why I proposed using `set`s.

repair_tests/repair_test.py
```diff
@@ -265,13 +264,16 @@ def _simple_repair(self, order_preserving_partitioner=False, sequential=True):
...
 
         for line, m in out_of_sync_logs:
-            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
-            self.assertIn((m.group(1), m.group(2)), valid, str((m.group(1), m.group(2))))
-            valid.remove((m.group(1), m.group(2)))
-            valid.remove((m.group(2), m.group(1)))
+            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+            self.assertIn({m.group(1), m.group(2)}, valid_out_of_sync_pairs, str((m.group(1), m.group(2))))
+            valid_out_of_sync_pairs.remove({m.group(1), m.group(2)})
+            valid_out_of_sync_pairs.remove({m.group(2), m.group(1)})
```
mambocab: This is already much clearer to me, but I still don't like the repeated use of `m.group(X)`. If nothing else, pull them out into named variables in the loop so the reader doesn't have to scan back up to the regex.

After addressing [this comment](https://github.com/riptano/cassandra-dtest/pull/826/files#r54474820), you'll want to remove one of the `.remove` lines.

repair_tests/repair_test.py
```diff
@@ -372,18 +374,18 @@ def local_dc_repair_test(self):
...
-        self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
-        valid = [node1.address(), node2.address()]
-        self.assertIn(m.group(1), valid, "Unrelated node found in local repair: " + m.group(1))
-        valid.remove(m.group(1))
-        self.assertIn(m.group(2), valid, "Unrelated node found in local repair: " + m.group(2))
+        self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+        valid_out_of_sync_pairs = [node1.address(), node2.address()]
+        self.assertIn(m.group(1), valid_out_of_sync_pairs, "Unrelated node found in local repair: " + m.group(1))
+        valid_out_of_sync_pairs.remove(m.group(1))
+        self.assertIn(m.group(2), valid_out_of_sync_pairs, "Unrelated node found in local repair: " + m.group(2))
```
mambocab: Same objection as above -- the reader has to scan up to the regex to figure out what `m.group(X)` is supposed to be.

repair_tests/repair_test.py
```diff
@@ -400,21 +402,59 @@ def dc_repair_test(self):
...
+        # Verify that only nodes in dc1 and dc2 are involved in repair
+        out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
+        self.assertEqual(len(out_of_sync_logs), 2, "Lines matching: " + str([elt[0] for elt in out_of_sync_logs]))
+        valid_out_of_sync_pairs = [{node1.address(), node2.address()},
+                                   {node2.address(), node1.address()},
+                                   {node2.address(), node3.address()},
+                                   {node3.address(), node2.address()}]
+        for line, m in out_of_sync_logs:
+            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+            self.assertIn({m.group(1), m.group(2)}, valid_out_of_sync_pairs, str((m.group(1), m.group(2))))
```
mambocab: Same redundancy in `valid_out_of_sync_pairs` and same objection to `m.group(X)` as noted above.

repair_tests/repair_test.py
```diff
@@ -400,21 +402,59 @@ def dc_repair_test(self):
...
-                 (node2.address(), node3.address()), (node3.address(), node2.address())]
+        valid_out_of_sync_pairs = [{node1.address(), node2.address()},
+                                   {node2.address(), node1.address()},
+                                   {node2.address(), node3.address()},
+                                   {node3.address(), node2.address()}]
         for line, m in out_of_sync_logs:
-            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync, got " + m.group(3))
-            self.assertIn((m.group(1), m.group(2)), valid, str((m.group(1), m.group(2))))
+            self.assertEqual(int(m.group(3)), 1, "Expecting 1 range out of sync for {} and {}, but saw {}".format(m.group(1), m.group(2), line))
+            self.assertIn({m.group(1), m.group(2)}, valid_out_of_sync_pairs, str((m.group(1), m.group(2))))
```
mambocab: Same as above.

repair_tests/repair_test.py
```diff
@@ -400,21 +402,59 @@ def dc_repair_test(self):
...
 
         # Check node2 now has the key
         self.check_rows_on_node(node2, 2001, found=[1000], restart=False)
 
+        # Check the repair was a dc parallel repair
+        if self.cluster.version() >= '2.2':
```
mambocab: Is this done with any frequency? I think some of this logic should be pulled out. At the very least, have the version logic pick an argument for `grep_log` and deduplicate the rest.

repair_tests/repair_test.py
```diff
@@ -454,6 +494,166 @@ def _setup_multi_dc(self):
...
+            debug("Inserting data...")
+            node1.stress(['write', 'n=2K', 'cl=ALL', '-schema', 'replication(factor=2)', '-rate', 'threads=30', '-pop', 'seq={}..{}K'.format(2 * (i - 1), 2 * i)])
+
+            node2.flush()
+            node2.stop(wait_other_notice=True)
+
+            node1.stress(['write', 'n=2K', 'cl=ONE', '-schema', 'replication(factor=2)', '-rate', 'threads=30', '-pop', 'seq={}..{}K'.format(2 * (i), 2 * (i + 1))])
+            node2.start(wait_for_binary_proto=True, wait_other_notice=True)
+
+            node1.stress(['write', 'n=2K', 'cl=ALL', '-schema', 'replication(factor=2)', '-rate', 'threads=30', '-pop', 'seq={}..{}K'.format(2 * (i + 1), 2 * (i + 2))])
```
mambocab: Is this much setup necessary to trigger the behavior we want to see? Seems like we could probably remove a `stress write` or 2.

repair_tests/repair_test.py
```diff
@@ -454,6 +494,166 @@ def _setup_multi_dc(self):
...
+
+            cluster.flush()
+            session = self.patient_cql_connection(node1)
+            session.execute("TRUNCATE system_traces.events")
+
+            job_thread_count = i
```
mambocab: I think it'd be best to remove this line and rewrite the `for` to be `for job_thread_count in ...`.

repair_tests/repair_test.py
mambocab: > Right, thus why I pulled the check.

Sorry, I must've been looking at a stale version of this PR when I left that comment.

> The rows[0][0] check at least asserts that system_traces.events WAS written to by repair? And asserting that the table grew is better than nothing.

Yeah, this is better than nothing. This is fine for merge, but could you file a ticket to talk with a dev about finding a reliable/meaningful assertion here?

repair_tests/repair_test.py
```diff
@@ -454,6 +494,166 @@ def _setup_multi_dc(self):
...
+        job_thread_count = '2'
+        opts = ['-tr', '-j', job_thread_count]
+        opts += _repair_options(self.cluster.version(), ks='keyspace1', cf='standard1', sequential=False)
+        node1.repair(opts)
+
+        time.sleep(5)  # Give the trace table some time to populate
+
+        session = self.patient_cql_connection(node1)
+        rows = list(session.execute("SELECT activity FROM system_traces.events"))
+        self.assertIn('job threads: {}'.format(job_thread_count),
```
mambocab: This (and the similar `assertIn` below) should (as you mentioned) have some comments. Just documenting the assumptions we talked about should be fine -- `this check assumes that the only (or at least first) thing to write to `system_traces.events.activity` is the repair task triggered in the test. That line reports X information`.

repair_tests/repair_test.py
```diff
@@ -454,6 +494,166 @@ def _setup_multi_dc(self):
...
+                      rows[0][0],
+                      'Expected {} job threads in repair options. Instead we saw {}'.format(job_thread_count, rows[0][0]))
+
+    def thread_count_repair_test(self):
```
mambocab: This test may belong in the nodetool tests. It might even be appropriate as a unit test; idk if the utests test `nodetool` features.

mambocab: I'm not dead set on it though, just worth giving a lil thought to it;

mambocab:
Just realized I made this comment:

https://github.com/riptano/cassandra-dtest/pull/826#discussion_r54476880

on an outdated change. Please address it before merge.

mambocab:
Good changes. Biggest remaining objection is that `match.group(X)` is not very readable compared to something like 

```python
num_out_of_sync_ranges, out_of_sync_nodes = m.group(1), {m.group(2), m.group(3)}
```

mambocab:
Done with this pass. Good stuff.

ptnapoleon:
@mambocab This is ready for review again.

mambocab:
LGTM. +1.

PR #827, opened 2016-02-25
Cql tests audit followup

mambocab:
This is ready for review. I'm going to make another PR pulling out the `SELECT` assertions to use functions from the `assertions` module. @knifewine and/or @ptnapoleon to review.

Much of this uses checks over the driver's cluster metadata instead of the previous assertions -- for example, rather than adding a column to a table, then asserting that a query has the result we expect, we directly check `<metadata object>'.columns`. To make this easier, I've added a collection of objects that, when accessing items or attributes, refresh the appropriate metadata and update themselves. The API there is up for grabs, but the idea is that it basically looks like a `cassandra.metadata.Metadata` object, except you don't have to refresh the metadata before accessing its attributes.

I'm still running these locally and I'll have to make changes if stuff fails. I'm just looking to get discussion started.

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -5,6 +5,7 @@ import struct
 import time
 
 from cassandra import ConsistencyLevel, InvalidRequest
+from cassandra.metadata import NetworkTopologyStrategy, SimpleStrategy
 from cassandra.policies import FallthroughRetryPolicy
 from cassandra.protocol import ProtocolException
 from cassandra.query import SimpleStatement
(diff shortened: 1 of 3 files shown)


ptnapoleon:
I like the changes so far. Should we move the metadata object out into its own namespace? We should start re-using it everywhere else we use the metadata api

mambocab:
> Should we move the metadata object out into its own namespace

Like, `utils/metadata_wrapper.py` or something?

ptnapoleon:
> > Should we move the metadata object out into its own namespace

> Like, utils/metadata_wrapper.py or something?
Yes

mambocab:
> > > Should we move the metadata object out into its own namespace

> > Like, utils/metadata_wrapper.py or something?

> Yes

Good idea; will do.

mambocab:
Ok, I've simplified the metadata wrapper objects. @ptnapoleon can you have a look at the changes please?

mambocab:
Just pushed a small refactoring for the metadata dealios. All the affected tests still pass locally. Can squash after merge.

mambocab:
By "can squash after merge" I mean "before merge".

ptnapoleon:
The new API looks good. +1

mambocab:
Squashed, merging. Thanks for the review @ptnapoleon

PR #828, opened 2016-02-26
Use metadata wrapper

mambocab:
This targets master after merging #827. Just wanted to start some discussion about the API used in `auth_tests`. Because of the way the wrapper works, You have to do this:

```python
ks_meta = UpdatingClusterMetadataWrapper(session.cluster).keyspaces['system_auth']
self.assertEquals(3, ks_meta.replication_strategy.replication_factor)
```

I could make `replication_strategy` evaluate to a wrapped object, which would make the underlying metadata object "inaccessible" (i.e. hidden behind an underscore-prefixed property). The downside is you couldn't do something like an `isinstance` check against a driver replication strategy class, which is something I've actually already done elsewhere. The upside is that it would make it possible to do something like:

```python
auth_replication_meta = UpdatingClusterMetadataWrapper(session.cluster).keyspaces['system_auth'].replication_strategy
self.assertEquals(3, auth_replication_meta.replication_factor)
```

and `.replication_factor` would be updated on every access. Python isn't sufficiently Lispy to make

```python
auth_replication_factor_meta = UpdatingClusterMetadataWrapper(session.cluster).keyspaces['system_auth'].replication_strategy.replication_factor
self.assertEquals(3, auth_replication_factor_meta)
```

possible, so that's off the table.

So, how confusing is all this? It'd be pretty easy to keep going down the chain and wrap metadata objects that we want to keep updated, but I don't know where to draw the line. The current API is kind of confusing if you don't know that, e.g., the objects you get from `.keyspaces[ks_name]` and `.keyspaces[ks_name].user_types` keep themselves updated, but the object you get from `.keyspaces[ks_name].replication_strategy` doesn't. But, the chain of auto-updating objects _has to_ stop somewhere because of Python's limitations, and maybe _should_ stop somewhere to make `isinstance` checks work.

Thoughts?


mambocab:
@knifewine and @ptnapoleon, that's whose thoughts I want.

auth_test.py
```diff
@@ -65,7 +65,7 @@ def system_auth_ks_is_alterable_test(self):
             debug('Checking node: {i}'.format(i=i))
             node = self.cluster.nodelist()[i]
             session = self.patient_exclusive_cql_connection(node, user='cassandra', password='cassandra')
-            self.assertEquals(3, session.cluster.metadata.keyspaces['system_auth'].replication_strategy.replication_factor)
+            self.assertEquals(3, ks_meta.replication_strategy.replication_factor)
```
ptnapoleon: If this doesn't update itself, don't we need to force an update?

tools.py
```diff
@@ -436,3 +436,21 @@ def __init__(self, node):
...
+    return cluster.metadata.keyspaces[keyspace_name]
+
+
+def get_schema_metadata(session):
+    cluster = session.cluster
+    cluster.refresh_schema_metadata()
+    return cluster.metadata
+
+
+def get_table_metadata(session, keyspace_name, table_name):
```
ptnapoleon: We appear to be using none of these methods anymore.

auth_test.py
mambocab: This does update itself -- accessing any attribute on `ks_meta` results in an update. The sequence of actions that wouldn't trigger an update would be

```python
rs_meta = UpdatingClusterMetadataWrapper(session.cluster).keyspaces['system_auth'].replication_strategy
self.assertEquals(3, rs_meta.replication_factor)
```

ptnapoleon:
Okay, after offline discussion, I don't love the current API. Users should be able to create each `UpdatingNounMetadataWrapper` directly, so that there's no confusion as to when the accessing stops giving other Wrappers. I would be equally happy with the current API, and a list of methods that abstract that magic away, but removing it is just as good.

mambocab:
I'm gonna go ahead and close this PR and leave the discussion where it stands. I'll change up the metadata wrapper over on #827 and then file a ticket to use the new API elsewhere. Porting this stuff is not crucial now IMO.

PR #829, opened 2016-02-26
Tests for CASSANDRA-7423

thobbs:


user_functions_test.py
```diff
@@ -212,8 +212,6 @@ def udf_with_udt_test(self):
 
         session.execute("create type test (a text, b int);")
 
-        assert_invalid(session, "create table tab (key int primary key, udt test);")
```
mambocab: Why is this being removed? If this is now valid because of a new feature, we should version-gate this statement. The check would look something like

```python
post_cassandra_7423 = ('2.2.5' < version < '3') or ('3' < version < '3.0.4') or ('3.4' <= version)
```

thobbs: I removed it from here because it didn't make sense to test unrelated UDT functionality in a user-defined function test (and it was broken by the changes).  However, it's probably a good idea to make another version of this test that covers non-frozen UDTs, so I'll push a commit to do that.

mambocab:
Thanks for explaining the change and adding a new method. Could you make this a parameterized test to remove the duplication here? Other than that, looks good.

mambocab:
Looks like this has fallen out of sync with master and has to be rebased, fyi.

mambocab:
Once it's been rebased, should we hold merge on CASSANDRA-7423?

ptnapoleon:
Right, these shouldn't pass without 7423 being committed

tjake:
I think this breaks non trunk branches. missing a @since() tag?

ptnapoleon:
The first test has a gate inside the test based on version, but it does appear to be failing still. We'll check it out

ptnapoleon:
I found the problem. I'll open a PR

PR #831, opened 2016-03-01
Cassandra 11207

blerer:
Add 2 new tests to cover CASSANDRA 11207 and change the expectation of an existing one.

### DIFF
--- a/thrift_tests.py
+++ b/thrift_tests.py
@@ -105,7 +105,7 @@ class ThriftTester(BaseTester):
             Cassandra.CfDef('Keyspace1', 'Indexed1', column_metadata=[Cassandra.ColumnDef('birthdate', 'LongType', Cassandra.IndexType.KEYS, 'birthdate_index')]),
             Cassandra.CfDef('Keyspace1', 'Indexed2', comparator_type='TimeUUIDType', column_metadata=[Cassandra.ColumnDef(uuid.UUID('00000000-0000-1000-0000-000000000000').bytes, 'LongType', Cassandra.IndexType.KEYS)]),
             Cassandra.CfDef('Keyspace1', 'Indexed3', comparator_type='TimeUUIDType', column_metadata=[Cassandra.ColumnDef(uuid.UUID('00000000-0000-1000-0000-000000000000').bytes, 'UTF8Type', Cassandra.IndexType.KEYS)]),
-
+            Cassandra.CfDef('Keyspace1', 'Expiring', default_time_to_live=2)
         ])
 
         keyspace2 = Cassandra.KsDef('Keyspace2', 'org.apache.cassandra.locator.SimpleStrategy', {'replication_factor': '1'},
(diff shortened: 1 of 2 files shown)


ttl_test.py
```diff
@@ -210,7 +210,7 @@ def remove_column_ttl_test(self):
 
     def remove_column_ttl_with_default_ttl_test(self):
         """
-        Test that we cannot remove a column ttl when a default ttl is set.
+        Test that we can remove the default ttl by setting the ttl explicitly to zero.
```
ptnapoleon: This is only the expected behavior on 3.6+ right? You've changed it to test this on all versions.
You should restrict the existing test with `@since('2.1', max_version='3.5')` and add the new code as a separate test restricted with `@since('3.6')`

ptnapoleon:
Can you also edit the docstrings for these tests to include `@jira_ticket CASSANDRA-11207`? I mentioned this on the ttl_test, but do those thrift_tests also need version restrictions?

ttl_test.py
```diff
@@ -208,6 +208,30 @@ def remove_column_ttl_test(self):
...
+        CASSANDRA-11207
+        """
+
+        self.prepare(default_time_to_live=2)
+
+        start = time.time()
+        self.session1.execute("""
+            INSERT INTO ttl_table (key, col1, col2, col3) VALUES (%d, %d, %d, %d);
+        """ % (1, 1, 1, 1))
```
mambocab: Could you please use `.format` rather than `%` here and below?

thrift_tests.py
```diff
@@ -1892,6 +1892,27 @@ def test_simple_expiration(self):
...
+    def test_expiration_with_default_ttl_and_zero_ttl(self):
+        """
+        Test that we can remove the default ttl by setting the ttl explicitly to zero
+        CASSANDRA-11207
+        """
+        _set_keyspace('Keyspace1')
+        column = Column('cttl3', 'value1', 0, 0)
+        client.insert('key1', ColumnParent('Expiring'), column, ConsistencyLevel.ONE)
+        c = client.get('key1', ColumnPath('Expiring', column='cttl3'), ConsistencyLevel.ONE).column
+        assert c == Column('cttl3', 'value1', 0), c
```
mambocab: Could you please use `self.assertEqual` here? It gives better error messages and is consistent with our style elsewhere.

thrift_tests.py
```diff
@@ -1027,7 +1027,7 @@ def test_bad_calls(self):
         # start > finish, key version
         _expect_exception(lambda: get_range_slice(client, ColumnParent('Standard1'), SlicePredicate(column_names=['']), 'z', 'a', 1, ConsistencyLevel.ONE), InvalidRequestException)
         # ttl must be positive
```
mambocab: This comment is no longer consistent with the code after this change. Has something changed to make a negative TTL valid?

mambocab:
Other than my style nits and my confusion over the possibly-outdated comment, looks good.

thrift_tests.py
blerer: Before the change a TTL <= 0 was invalid for Thrift. Which was not making a lot of sense as a TTL of zero was valid in CQL. After the change we allow TTLs equals to zero but still reject negative TTLs. That is why I changed the TTL value from zero to -1 (like that the query is still rejected). I will update the comment.

mambocab:
Looks good, and passes locally with CASSANDRA-11207. Merging, thanks @blerer

PR #834, opened 2016-03-03
increase timeout likely causing ci test flap

knifewine:


knifewine:
https://issues.apache.org/jira/browse/CASSANDRA-11211

ptnapoleon:
I don't understand, in the CI failure, we had too many notifications, not too few, right? I don't think increasing the timeout would fix that.

Unless I'm unaware of the failure cause

knifewine:
You make a good point there. with the timeout bump I'm no longer repro-ing locally but maybe that's unrelated. Seems like the intent of the test was to wait for at least 3 notifications, so maybe it just needs to be relaxed to 3 or more.

ptnapoleon:
I disagree, based on initial analysis, that it was waiting for at least 3. I think we should be expecting an exact amount here.

knifewine:
ok, good thinking, closing this to look into a new fix.

PR #835, opened 2016-03-03
Expect non-matching repair values in repairedset_test

ptnapoleon:
Also add better debug output

@knifewine to review.


repair_tests/incremental_repair_test.py
```diff
@@ -185,8 +182,11 @@ def sstable_repairedset_test(self):
             initialoutput = r.read()
 
         matches = findall('(?<=Repaired at:).*', initialoutput)
+        debug(matches)
```
knifewine: is this debug useful to keep? if so, maybe a little text to give it context when it's output

knifewine:
+1

PR #837, opened 2016-03-04
add a forgotten since decorator

knifewine:
@tjake to review.

### DIFF
--- a/cql_tracing_test.py
+++ b/cql_tracing_test.py
@@ -1,7 +1,7 @@
 # coding: utf-8
 
 from dtest import Tester, debug
-from tools import known_failure
+from tools import since
 
 
 def post_cassandra_10392(version):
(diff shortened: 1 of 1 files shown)


mambocab:
See my note [here](https://issues.apache.org/jira/browse/CASSANDRA-11278?focusedCommentId=15183139&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15183139):

> I think a more complex `@since` is necessary – if I understand correctly, the change was merged to 3.0.4 on the 3.0 track, and 3.4 on the tick-tock track.
> Side note – my mistake, I'm the one who forgot to add the decorators.

knifewine:
ok, feel free to close this unless there's any reason to keep it out here. I was basically told 3.4+ only last week for this (but I could see in the code the check method was doing something a bit more involved, hence having a review at all).

mambocab:
Nope, I was wrong, my mistake. This is 3.4+ only.

mambocab:
Which means that the code check method is erroneously complex. Personally I kinda like pulling that logic out into a function with a name. In any event, the function should either be fixed or replaced with an inline comparison to `'3.4'`.

knifewine:
@mambocab, I force pushed changes to remove the extra logic.

knifewine:
with a proper since on 3.4, do we need the check at all?

mambocab:
Hah, no, we don't need the check. Awesome, +1.

knifewine:
just to be sure, I ran the tests locally on 3.4 and they pass. merging now.

PR #838, opened 2016-03-04
Cstar 343 offline tools

mshuler:


ptnapoleon:
We should also test the sstableupgrade tool in the situation where an upgrade does need to occur. We don't even need to generate them in the test. We could manually create them, and then bundle them in dtest. One row sstables shouldn't take too much space.

offline_tools_test.py
```diff
@@ -272,6 +273,54 @@ def sstableexpiredblockers_test(self):
...
+        session.execute("insert into ks.cf (key, val) values (1,1)")
+        node1.flush()
+        cluster.stop()
+        [(out, error, rc)] = node1.run_sstabledump(keyspace="ks", column_families=["cf"])
+        debug(out)
+        debug(error)
+        # Load the json output and check that it contains the inserted key=1
+        s = json.loads(out)
+        debug(s)
+        self.assertEquals('1', s[0]['partition']['key'][0])
```
mambocab: I think this would be more readable (and, as an added bonus, stricter) as

```python
self.assertEqual(len(s), 1)
dumped_row = s[0]
self.assertEqual(dumped_row['partition']['key'], '1')
```

At the very least we should definitely check the length.

offline_tools_test.py
```diff
@@ -272,6 +273,54 @@ def sstableexpiredblockers_test(self):
...
+        # Check that we only get the key back using the enumerate option
+        [(out, error, rc)] = node1.run_sstabledump(keyspace="ks", column_families=["cf"], enumerate_keys=True)
+        debug(out)
+        debug(error)
+        s = json.loads(out)
+        debug(s)
+        self.assertEquals('1', s[0][0])
```
mambocab: Similar thought here -- it'd be good to check the length of `s` and pull out values into named variables for readability.

mambocab:
After my readability and length-checking nits, I'm +1, go ahead and merge this. I'm +1 on @ptnapoleon's suggestion; file a CSTAR bug if we don't want do deal with it right now.

mambocab:
Also, could you please squash this into smaller commits? Maybe 1 per test, or even just 1 big commit would be fine.

mshuler:
Thanks for the feedback! I'll get the nits committed and I started working on upgrading from an older sstable, so I'll squash when that's ready to include.

mshuler:
Added the row length check and dumped_row vars and sstable upgrade steps, except for upgrading 2.0->2.1, since CCM has a problem with the jamm jar version update there.

offline_tools_test.py
```diff
@@ -272,6 +273,98 @@ def sstableexpiredblockers_test(self):
...
+        """
+        Test that sstableupgrade functions properly offline on a same-version Cassandra sstable, a
+        stdout message of "Found 0 sstables that need upgrading." should be returned.
+        """
+        # Set up original node version to test for upgrade
+        cluster = self.cluster
+        testversion = cluster.version()
+        original_install_dir = cluster.get_install_dir()
+        debug('Original install dir: {}'.format(original_install_dir))
+        # Set up last major version to upgrade from, assuming 2.0 branch is the oldest tested version
```
ptnapoleon: I don't think we need to test 1.2 -> 2.0

ptnapoleon:
This is very well commented, but could you possibly add line breaks to make chunks of code? It is a bit hard on the eyes like this.

mshuler:
I set the oldest version assumed for sstableupgrade to 2.1 and added some whitespace. Both tests pass locally for me.

ptnapoleon:
+1

PR #839, opened 2016-03-07
Remove blist

mambocab:
`blist` shouldn't be necessary anymore since the changes to the driver last year.

(this also sorts the bare package installs in `requirements.txt`, I didn't add `docopt` in the correct position in the file.)


ptnapoleon:
+1

mambocab:
Thanks, merging.

PR #840, opened 2016-03-07
don't manually calculate set() and Counter()

mambocab:
Small thing to improve that I found while filing a bug. @ptnapoleon and/or @knifewine to review.

### DIFF
--- a/repair_tests/incremental_repair_test.py
+++ b/repair_tests/incremental_repair_test.py
@@ -1,3 +1,4 @@
+from collections import Counter
 import os
 import time
 from re import findall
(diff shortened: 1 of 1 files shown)


knifewine:
hey, collections.Counter is neato.

looks there's another repeat of the same counting logic below (line ~220). if that one's easy to attack we may as well fix both?

i'm +1 on fixing just this one, and also +1 on fixing both.

knifewine:
looks like travis is upset about extra empty lines.

mambocab:
Good catches, thanks.

mambocab:
Can I get a re-review? I think travis should be happy this time.

knifewine:
+1

mambocab:
Thanks, merging.

PR #844, opened 2016-03-09
A new test for CASSANDRA-11053

stef1927:
This introduces a new test for CASSANDRA-11053, which was committed yesterday and added a new COPY FROM option for disabling prepared statements. 

I also noticed [one failure] (http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-11053-2.1-dtest/16/testReport/cqlsh_tests.cqlsh_copy_tests/CqlshCopyTest/test_bulk_round_trip_blogposts/) in the bulk copy tests due to a wrong element count, so I've increased the consistency level to ALL.


ptnapoleon:
+1

PR #845, opened 2016-03-09
Changes to formatting.format_value() for CASSANDRA-11274

stef1927:
CASSANDRA-11274 will change the signature of `formatting.format_value()` and this pull request addresses this, see comments in the patch.

This patch is backward compatible so we can merge at any time but once CASSANDRA-11274 is committed, some cqlsh copy tests will fail unless this pull request is merged.


ptnapoleon:
+1

PR #847, opened 2016-03-09
remove upgrade paths that won't change anymore

mambocab:
@knifewine could you have a look-see? This has to do with our recent talk about what to do with, e.g., the tests upgrading from 3.1 to 3.2.


ptnapoleon:
I'm +1. 3.1 to 3.2 will never change nor 3.2 to 3.3, and while technically 3.0 to 3.1/3.2 of will, as 3.0 is still changing, it's moot as the upgrades to 3.3 and 3.4 exist.

Make Travis less sad though, please

knifewine:
+1 after travis is happy.

mambocab:
Thanks y'all, merging.

PR #850, opened 2016-03-10
remove upgrade_through dep on schema_metadata_test

knifewine:
this has been the source of a number of problems and should be
removed for simplification.

each of these test modules is built with a different purpose in mind
and having them tangled has become an issue.

this needs to be followed up with proper expansion of schema coverage
within upgrade_through_versions or in another test built for the express
purpose of upgrade testing

### DIFF
--- a/upgrade_tests/upgrade_through_versions_test.py
+++ b/upgrade_tests/upgrade_through_versions_test.py
@@ -10,11 +10,10 @@ from multiprocessing import Process, Queue
 from Queue import Empty, Full
 
 import psutil
-from cassandra import ConsistencyLevel, WriteTimeout
-from cassandra.query import SimpleStatement
 from six import print_
 
-import schema_metadata_test
+from cassandra import ConsistencyLevel, WriteTimeout
+from cassandra.query import SimpleStatement
 from dtest import Tester, debug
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I'm not familiar with what this code does. Have you already created the ticket for replacing it?

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -377,18 +375,12 @@ def upgrade_scenario(self, populate=True, create_schema=True, rolling=False, aft
                 self._write_values()
                 self._increment_counters()
 
-                for completed_version in self.created_metadata_versions:
-                    self._check_metadata_schemas(completed_version[0], completed_version[1])
```
ptnapoleon: Should we remove `def _check_metadata`, as we're removing `def _create_metadata`?

ptnapoleon:
+1 on the code. If removing is the appropriate thing to do, go for it.

knifewine:
replacement will fall under the banner of https://datastax.jira.com/browse/CSTAR-294
but should ultimately be more robust all the same.

mambocab:
+1 if the CassCI job runs clean.

knifewine:
ci looks nice, mergin'

PR #853, opened 2016-03-11
isort and skip hanging test

mambocab:
@mshuler Still testing this locally, but does this look like it'll address CSTAR-347?

### DIFF
--- a/secondary_indexes_test.py
+++ b/secondary_indexes_test.py
@@ -3,14 +3,16 @@ import random
 import re
 import time
 import uuid
+from unittest import skipIf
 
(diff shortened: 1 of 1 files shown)


mambocab:
I guess I mean "address" pretty generously -- I just mean, will it keep stuff from hanging for the moment? It does correctly skip the test if `OFFHEAP_MEMTABLES` is set.

mshuler:
Couldn't hurt :)  I haven't fully researched why that test is hanging.

mambocab:
Cool cool, thank you. Merging.

PR #854, opened 2016-03-14
Monitor datadirectory during cleanup

krummas:
Make sure we don't get too many files during cleanup

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -524,7 +524,7 @@ class TestBootstrap(Tester):
         Make sure we remove processed files during cleanup
         """
         cluster = self.cluster
-        cluster.set_configuration_options(values={'concurrent_compactors': 1})
+        cluster.set_configuration_options(values={'concurrent_compactors': 4})
         cluster.populate(1)
         cluster.start(wait_for_binary_proto=True)
         node1, = cluster.nodelist()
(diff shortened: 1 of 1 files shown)


krummas:
Wrote this to rule out a bug in CASSANDRA-11179

bootstrap_test.py
```diff
@@ -518,6 +519,38 @@ def simultaneous_bootstrap_test(self):
...
+        thread = threading.Thread(target=self._monitor_datadir, args=(node1, event, len(node1.get_sstables("keyspace1", "standard1")), failed))
+        thread.start()
+        node1.cleanup()
+        event.set()
+        thread.join()
+        self.assertFalse(failed.is_set())
+
+    def _monitor_datadir(self, node, event, basecount, failed):
+        while True:
+            if len(node.get_sstables("keyspace1", "standard1")) > basecount + 1:
```
mambocab: About how many SSTables do we expect here? If there are relatively few, which is what it looks like based on my reading of CASSANDRA-11179, it'd be great to print the names of the SSTables each time. If this test ends up failing someday it'll be helpful for debugging in the future.

krummas:
just updated the branch - it now tests against the new --seq parameter in 11179 and prints sstables (which should always be 5 - we disable autocompaction and flush 5 times)

bootstrap_test.py
```diff
@@ -518,6 +519,42 @@ def simultaneous_bootstrap_test(self):
         for _ in xrange(5):
             assert_one(session, "SELECT count(*) from keyspace1.standard1", [500000], cl=ConsistencyLevel.ONE)
 
+    @require(11179)
+    def test_cleanup(self):
+        """
+        Make sure we remove processed files during cleanup
```
ptnapoleon: Can you add `@jira_ticket CASSANDRA-11179` to this docstring?

ptnapoleon:
Should we be asserting the number of sstables is always 5?

krummas:
No, I don't think so, if cassandra decides to flush more sstables that is fine, we just want to make sure that we process one file at a time

ptnapoleon:
+1

krummas:
I'll hold off merging - changing the parameter in 11179 to --jobs instead of --seq

PR #856, opened 2016-03-14
Make changes to tests based on CCM node start API changes

ptnapoleon:
https://github.com/pcmanus/ccm/pull/464. wait-other-notice now defaults to True, so we need
to set it to False in some places. Also, start is no longer
idempotent, so we can't double start() in some tests

Running here:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-cassandra-2.2-dtest/

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -156,7 +156,7 @@ class TestBootstrap(Tester):
         # keep timeout low so that test won't hang
         node3.set_configuration_options(values={'streaming_socket_timeout_in_ms': 1000})
         try:
-            node3.start()
+            node3.start(wait_other_notice=False)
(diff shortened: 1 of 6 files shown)


ptnapoleon:

ptnapoleon:
Okay, ready for review

mambocab:
How did you choose the `.start` calls to pass a non-default `wait_other_notice` value to?

ptnapoleon:
I ran CI with the CCM changes, then looked at the failing tests. These *should* be all the tests that start a node but expect it not to come up.

mambocab:
Got it, sounds reasonable. +1 from me then.

PR #857, opened 2016-03-14
whole_list_condition_test flakiness fix

knifewine:
write and read back at more appropriate levels
write at cl.all and read back at cl.one

it's possible that this test was failing previously
when a coordinator (and selfsame replica) would defer
to another node at cl.one, allowing a stale data read
and therefore an assertion failure

### DIFF
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -4267,6 +4267,8 @@ class TestCQL(UpgradeTester):
                 l frozen<list<text>>
             )""")
 
+        cl = getattr(self, 'CL', ConsistencyLevel.ONE)
+
         for is_upgraded, cursor in self.do_upgrade(cursor):
             debug("Querying %s node" % ("upgraded" if is_upgraded else "old",))
             cursor.execute("TRUNCATE tlist")
@@ -4275,11 +4277,11 @@ class TestCQL(UpgradeTester):
             for frozen in (False, True):
 
                 table = "frozentlist" if frozen else "tlist"
-                cursor.execute("INSERT INTO %s(k, l) VALUES (0, ['foo', 'bar', 'foobar'])" % (table,))
+                cursor.execute("INSERT INTO {}(k, l) VALUES (0, ['foo', 'bar', 'foobar'])".format(table))
 
                 def check_applies(condition):
-                    assert_one(cursor, "UPDATE %s SET l = ['foo', 'bar', 'foobar'] WHERE k=0 IF %s" % (table, condition), [True])
-                    assert_one(cursor, "SELECT * FROM %s" % (table,), [0, ['foo', 'bar', 'foobar']])
+                    assert_one(cursor, "UPDATE {} SET l = ['foo', 'bar', 'foobar'] WHERE k=0 IF {}".format(table, condition), [True], cl=cl)
+                    assert_one(cursor, "SELECT * FROM {}".format(table), [0, ['foo', 'bar', 'foobar']])  # read back at default cl.one
 
                 check_applies("l = ['foo', 'bar', 'foobar']")
                 check_applies("l != ['baz']")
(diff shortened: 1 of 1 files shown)


knifewine:
@mambocab to review.

knifewine:
I also snuck in some string formatting changes for readability.

I ran a slightly modified version of the test locally (with a loop going through thousands of repetitions) and it was able to pass without assertion errors.

upgrade_tests/cql_tests.py
```diff
@@ -4267,6 +4267,8 @@ def whole_list_conditional_test(self):
                 l frozen<list<text>>
             )""")
 
+        cl = getattr(self, 'CL', ConsistencyLevel.ONE)
```
mambocab: `self.CL` should always be set; there's a default set on the `UpgradeBase` class:

https://github.com/riptano/cassandra-dtest/blob/master/upgrade_tests/upgrade_base.py#L106

mambocab:
I think we want to run this by @thobbs or @pcmanus -- I thought we wrote and read at `ONE` to ensure we only read from a single node, so maybe the solution here is to restrict the query by some other means.

upgrade_tests/cql_tests.py
knifewine: ah, I thought it could be none. pushing a change to that momentarily.

knifewine:
based my a conversation yesterday I'm not certain we can prevent a coordinator who is a replica from going out to another replica in rare cases. 
read_from_others

mambocab:
Hmm, sure. I guess I don't understand the intent of using an exclusive connection, then. If we don't know that the query is going to be executed on particular nodes, then what are we testing with this setup?

ptnapoleon:
In the general case, exclusive guarantees the node will be coordinator, not handle the queries. You probably know that, but worth recording here.

ptnapoleon:
Is the exclusive connection in the `prepare()`? I don't see it in the test proper.

knifewine:
In this case the prepare connection isn't exclusive, but it's not actually
used for the majority of the test.

ptnapoleon:
Oh, it's coming from `do_upgrade` then?

knifewine:
yep

knifewine:
I didn't write this test, so I'm less clear on the unstated goals, but the impression I had was just that we're checking the conditional application of updates. I don't think we need a guarantee of exclusivity for that case anyway. If the test is not valid without an exclusivity guarantee, we can retire it as well.

ptnapoleon:
Okay. The point of the exclusivity is to ensure that in the mixed version cluster we set up, that a node from each version is tested as coordinator, to ensure that queries work in both directions, because bugs won't be symmetrical. 

As far as the conditional application of updates, I 100% think raising the CL is fine, based on my understanding of the issue.

knifewine:
We could keep the low CL and introduce repairs as part of the test as well. not sure if that's the right approach though.

ptnapoleon:
-1 on low CL and repairing.

mambocab:
> ensure that in the mixed version cluster we set up, that a node from each version is tested as coordinator

Seems odd to me that, with RF=ALL, the coordinator would query another node on a read with CL=ONE. As long as that's not a C* bug, I think raising the CL for writes should be fine.

I feel like we should file a ticket to write a different value with each of the updates in this test. Am I off-base?

knifewine:
> Seems odd to me that, with RF=ALL, the coordinator would query another node on a read with CL=ONE. As long as that's not a C* bug, I think raising the CL for writes should be fine

the all before wasn't being respected because assert_one, et al, create statements and apply a default consistency level of one to that statement, so this was superceding what was set on the session.

> I feel like we should file a ticket to write a different value with each of the updates in this test. Am I off-base?

seems like a good idea to me

thobbs:
It sounds like you guys sorted everything out, but just to confirm, the test is primarily concerned with ensuring the conditional updates apply correctly.  Specifically, making sure that all of the operators work correctly with different values and nulls.

> I feel like we should file a ticket to write a different value with each of the updates in this test. Am I off-base?

It would be nice to do this to ensure that all of the paxos writes are happening correctly.  That's not the original purpose of the test, but I think it's a good extension, and especially makes sense for an upgrade test.

ptnapoleon:
+1 to this code, and also to filing a ticket to extend as described by Jim and Tyler

mambocab:
@ptnapoleon I couldn't find the ticket -- could you assign me? i have Ideas.

PR #858, opened 2016-03-15
CASSANDRA-10990 dtests

pauloricardomg:
Tests for CASSANDRA-10990

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -17,6 +17,7 @@ import threading
 import time
 import traceback
 import types
+import glob
 from collections import OrderedDict
 from unittest import TestCase
 
@@ -462,6 +463,18 @@ class Tester(TestCase):
             # thread.interrupt_main will SIGINT in the main thread, which we can
             # catch to raise an exception with useful information
             thread.interrupt_main()
+
+    """
+    Finds files matching the glob pattern specified as argument on
+    the given keyspace in all nodes
+    """
+    def glob_data_dirs(self, path, ks="ks"):
+        result = []
+        for node in self.cluster.nodelist():
+            for data_dir in node.data_directories():
+                ks_dir = os.path.join(data_dir, ks, path)
+                result.extend(glob.glob(ks_dir))
+        return result
 
     def _catch_interrupt(self, signal, frame):
         """
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -37,6 +37,10 @@ def _repair_options(version, ks='', cf=None, sequential=True):
         opts += [cf]
     return opts
 
+LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
+                            "-Dcassandra.streamdes.max_mem_buffer_size=5",
+                            "-Dcassandra.streamdes.max_spill_file_size=16"]
+
 
 class TestRepair(Tester):
 
@@ -194,6 +198,24 @@ class TestRepair(Tester):
         """
         self._empty_vs_gcable_no_repair(sequential=False)
 
+    @since('3.0')
+    def repair_after_upgrade_test(self):
+        """
+        @jira_ticket CASSANDRA-10990
+        """
+        default_install_dir = self.cluster.get_install_dir()
+        cluster = self.cluster
+        cluster.set_configuration_options(values={'hinted_handoff_enabled': False})
+        cluster.set_batch_commitlog(enabled=True)
+
+        # Forcing cluster version on purpose
+        debug("Starting cluster..")
+        cluster.set_install_dir(version="2.1.9")
+        self._populate_cluster()
+
+        self._do_upgrade(default_install_dir)
+        self._repair_and_verify(True)
+
     @no_vnodes()
     def simple_repair_order_preserving_test(self):
         """
@@ -202,29 +224,8 @@ class TestRepair(Tester):
         """
         self._simple_repair(order_preserving_partitioner=True)
 
-    def _simple_repair(self, order_preserving_partitioner=False, sequential=True):
-        """
-        * Configure a three node cluster to not use hinted handoff, and to use batch commitlog
-        * Launch the cluster
-        * Create a keyspace at RF 3 and table
-        * Insert one thousand rows at CL ALL
-        * Flush on node3 and shut it down
-        * Insert one row at CL TWO
-        * Restart node3
-        * Insert one thousand more rows at CL ALL
-        * Flush all nodes
-        * Check node3 only has 2000 keys
-        * Check node1 and node2 have 2001 keys
-        * Perform the repair type specified by the parent test
-        * Assert the appropriate messages are logged
-        * Assert node3 now has all data
-
-        @jira_ticket CASSANDRA-4373
-        """
+    def _populate_cluster(self, start=True):
         cluster = self.cluster
-
-        if order_preserving_partitioner:
-            cluster.set_partitioner('org.apache.cassandra.dht.ByteOrderedPartitioner')
 
         # Disable hinted handoff and set batch commit log so this doesn't
         # interfere with the test (this must be after the populate)
@@ -248,6 +249,35 @@ class TestRepair(Tester):
         insert_c1c2(session, keys=range(1001, 2001), consistency=ConsistencyLevel.ALL)
 
         cluster.flush()
+
+    def _simple_repair(self, order_preserving_partitioner=False, sequential=True):
+        """
+        * Configure a three node cluster to not use hinted handoff, and to use batch commitlog
+        * Launch the cluster
+        * Create a keyspace at RF 3 and table
+        * Insert one thousand rows at CL ALL
+        * Flush on node3 and shut it down
+        * Insert one row at CL TWO
+        * Restart node3
+        * Insert one thousand more rows at CL ALL
+        * Flush all nodes
+        * Check node3 only has 2000 keys
+        * Check node1 and node2 have 2001 keys
+        * Perform the repair type specified by the parent test
+        * Assert the appropriate messages are logged
+        * Assert node3 now has all data
+
+        @jira_ticket CASSANDRA-4373
+        """
+        if order_preserving_partitioner:
+            cluster.set_partitioner('org.apache.cassandra.dht.ByteOrderedPartitioner')
+
+        self._populate_cluster()
+        self._repair_and_verify(sequential)
+
+    def _repair_and_verify(self, sequential=True):
+        cluster = self.cluster
+        node1, node2, node3 = cluster.nodelist()
 
         # Verify that node3 has only 2000 keys
         debug("Checking data on node3...")
@@ -756,6 +786,20 @@ class TestRepair(Tester):
         node3.stop(wait_other_notice=True)
         (stdout, stderr) = node2.stress(['read', 'n=1M', 'no-warmup', '-rate', 'threads=30', '-node', node2.address()], capture_output=True)
         self.assertTrue(len(stderr) == 0, stderr)
+
+    def _do_upgrade(self, default_install_dir):
+        cluster = self.cluster
+
+        for node in cluster.nodelist():
+            debug("Upgrading %s" % node.name)
+            if node.is_running():
+                node.flush()
+                time.sleep(1)
+                node.stop(wait_other_notice=True)
+            node.set_install_dir(install_dir=default_install_dir)
+            node.start(wait_other_notice=True, wait_for_binary_proto=True)
+            cursor = self.patient_cql_connection(node)
+        cluster.set_install_dir(default_install_dir)
 
 RepairTableContents = namedtuple('RepairTableContents',
                                  ['parent_repair_history', 'repair_history'])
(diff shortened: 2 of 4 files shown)


dtest.py
```diff
@@ -422,6 +423,14 @@ def _log_error_handler(self, errordata):
         self.exit_with_exception = AssertionError("Log error encountered during active log scanning, see stdout")
         thread.interrupt_main()
 
+    def glob_data_dirs(self, path, ks="ks"):
```
ptnapoleon: @mambocab, should we put this somewhere else?

sstable_generation_loading_test.py
```diff
@@ -222,3 +167,135 @@ def read_and_validate_data(session):
...
+        for data_dir in node1.data_directories():
+            basepath = os.path.join(data_dir, 'keyspace1')
+            for x in os.listdir(basepath):
+                if x.startswith("standard1"):
+                    path = os.path.join(basepath, x)
+
+            os.system('rm %s/*Index.db' % path)
+            os.system('rm %s/*Filter.db' % path)
+            os.system('rm %s/*Statistics.db' % path)
+            os.system('rm %s/*Digest.sha1' % path)
```
ptnapoleon: This is almost certainly not going to work on Windows.

ptnapoleon: Nevermind, I now see this is just relocated code.

dtest.py
mambocab: I don't have strong feelings; `dtest` is a bit of a dumping ground, so putting this here doesn't make the module appreciably worse. Maybe this is an opportunity to start something like a `./utils/data_dir_utils.py` or something, though. Or, this could live in CCM?

ptnapoleon: It could live in ccmlib.common, potentially. I don't really mind where it's put, as you've said, it doesn't make this module appreciably worse.

mambocab: Sorry, that was an incomplete thought. This is only used once, to find temporary files. Maybe there could be a `get_temp_files` or similar function on `ccmlib.node.Node`.

repair_tests/repair_test.py
```diff
@@ -36,6 +36,10 @@ def _repair_options(version, ks='', cf=None, sequential=True):
         opts += [cf]
     return opts
 
+LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
+                            "-Dcassandra.streamdes.max_mem_buffer_size=5  ",
```
mambocab: If the space at the end of this string is necessary, could you document it with a comment? and if it isn't, could you remove it?

dtest.py
ptnapoleon: Yes, there could.

mambocab: It's not worth blocking merge on this, but I'll file a ticket to look for other code that does similar stuff, or grabs temp files, to see if it's worth moving into CCM.

repair_tests/repair_test.py
```diff
@@ -202,6 +224,32 @@ def simple_repair_order_preserving_test(self):
...
 
+    def _populate_cluster(self, start=True):
+        cluster = self.cluster
+
+        if start:
+            # Disable hinted handoff and set batch commit log so this doesn't
+            # interfere with the test (this must be after the populate)
+            cluster.set_configuration_options(values={'hinted_handoff_enabled': False}, batch_commitlog=True)
+            debug("Starting cluster..")
+            cluster.populate(3).start()
```
mambocab: I believe this `.populate` call should be outside the `if start:` block. This would require messing with https://github.com/riptano/cassandra-dtest/pull/858/files#diff-370753b9f22e170af87c1cdeab3144ecR213 but if this is going to be named `_populate_cluster`, it shouldn't not populate the cluster for some values of `start`.

sstable_generation_loading_test.py
```diff
@@ -222,3 +167,135 @@ def read_and_validate_data(session):
...
+        Tester.__init__(self, *argv, **kwargs)
+        BaseSStableLoaderTest.__init__(self, upgrade_from='2.2.4', jvm_args=LEGACY_SSTABLES_JVM_ARGS)
+
+
+@since('3.0')
+class TestLoadLaCompactSStables(Tester, BaseSStableLoaderTest):
+
+    def __init__(self, *argv, **kwargs):
+        kwargs['cluster_options'] = {'start_rpc': 'true'}
+        Tester.__init__(self, *argv, **kwargs)
```
mambocab: Looks like each of these subclasses inherits from both `Tester` and `BaseSStableLoaderTest`, and each of their `__init__`s calls both of the superclass' `__init__`s. There are a couple improvements you could make here:

- If you call `Tester.__init__` in `BaseSStableLoaderTest.__init__`, you can avoid having to call both.
- If you take the values you're passing to `BaseSStableLoaderTest.__init__` and make them class variables instead, you can just override them in these subclasses rather than having to implement `__init__` at all.

pauloricardomg:
Oops, seems like I missed this notification from quite a while. Thanks for the review!

Addressed comments except  `Tester` and `BaseSStableLoaderTest` from @mambocab . The reason why I didn't make `BaseSStableLoaderTest` inherit from `Tester` is that I don't want it to run as a test class, since it will already be run by `TestSSTableGenerationAndLoading`. Is there anyway of making `BaseSStableLoaderTest` inherit from `Tester` without being executed as a test?

I don't know why travis checks are failing, any idea?

pauloricardomg:
I also submitted a [dtest run](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-newtrunk-10990-dtest/) to make sure tests are working after rebase.

ptnapoleon:
You can ignore Travis's current complaint

sstable_generation_loading_test.py
```diff
@@ -1,101 +1,30 @@
...
+                            "-Dcassandra.streamdes.max_spill_file_size=64"]
 
 
-class TestSSTableGenerationAndLoading(Tester):
+class BaseSStableLoaderTest():
 
-    def __init__(self, *argv, **kwargs):
-        kwargs['cluster_options'] = {'start_rpc': 'true'}
-        super(TestSSTableGenerationAndLoading, self).__init__(*argv, **kwargs)
+    def __init__(self, upgrade_from=None, compact=False, cluster_options=None, jvm_args=[]):
```
mambocab: Sorry I didn't notice this before -- because `jvm_args` is created at method-creation time and is mutable, changes to it will persist across calls to `__init__`, even on different objects (see [here](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) for a more detailed explanation). The pattern we use in this codebase to avoid this is

```python
def method(arg=None):
    if arg is None:
        arg = []
```

mambocab:
Ah, I see. I'm not sure, but I think `BaseSStableLoaderTest` will be picked up anyway, since its name ends in `Test`. You can check this using `nosetests --collect-only`. On my Ubuntu box I would use `nosetests --collect-only 2>&1 | grep BaseSST`, but iirc you run Windows, and I don't know the equivalent there.

@ptnapoleon How do we want to manage this code-sharing? The current implementation is basically a mixin (`BaseSStableLoaderTest`) used in 4 hand-created classes, with configuration in `__init__` methods. We definitely want to move that configuration to class variables, but do we want to do that generate-classes-in-a-loop thing we do elsewhere? Seems like a reasonable candidate for it.

mambocab:
@pauloricardomg Just realized I didn't answer your question about how to make the base class not run as a test. How we use it will depend on how we decide to share the code in that mixin, but the basic way we do it is: set a class variable named `__test__` on the base class to `False`, then override it as `True` on subclasses that we want to run as tests. `nose` will ignore objects where `object.__test__ == False`.

pauloricardomg:
Fixed `jvm_args`nit and checked with `nosetests --collect-only 2>&1 | grep BaseSST` (I don't use win btw :P) that `BaseSStableLoaderTest` is in fact not executed since it does not inherit from `Tester`. Given this, do you prefer keeping as it is, or make `BaseSStableLoaderTest` inherit from `Tester` and set `self.__test__ == False`, while subclasses make `self.__test__ == True` and override attributes?

Submitted tests seem to be [passing](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-newtrunk-10990-dtest/lastCompletedBuild/testReport/).

mambocab:
The more I think about it, the more I think I prefer using single inheritance and manually specifying values for `__test__`. Sorry to ask for a biggish refactor here, but I think it'll be easier to reason about in the future.

Also, I would prefer if all the logic in the subclass' `__init__` methods were be moved to `BaseSStableLoaderTest.__init__`, and the configuration were moved to class variables. That leaves us less duplication and more explicit  So, after these 2 changes, the code should look something like

```python
class BaseSStableLoaderTest(Tester):
    __test__ = False

    upgrade_from = None
    compact = False
    allow_log_errors = True
    jvm_args = None

    def __init__(self, *argv, **kwargs):
        if self.jvm_args is None:
            self.jvm_args = []
        kwargs['cluster_options'] = {'start_rpc': True}
        super(BaseSStableLoadertest, self).__init__(*args, **kwargs)

    ...


@since('3.0')
class TestLoadKaSStables(Tester, BaseSStableLoaderTest):
    __test__ = True
    upgrade_from = '2.1.6'
    jvm_args = LEGACY_SSTABLES_JVM_ARGS

...
```

(I haven't actually run this code, but this is the basic idea for moving configuration to class-level variables from initialization code. I think it's more readable and will make maintenance easier when making changes across test classes.)

Something I noted while writing this: I couldn't find anywhere that `cluster_options` was used, which is why I didn't reproduce it above.

Does that seem reasonable to you @pauloricardomg?

pauloricardomg:
Finally addressed suggestions. Can you check if it looks better now @mambocab ?

Also added new `storage_engine_upgrade_test:TestBootstrapAfterUpgrade` which I forgot to commit previously. Submitted new dtest run with new tests on trunk: http://cassci.datastax.com/job/pauloricardomg-trunkcopy-dtest/3/testReport/upgrade_tests.storage_engine_upgrade_test/

(the failing `upgrade_with_wide_partition_test` seems to be CASSANDRA-11663)

pauloricardomg:
friendly ping @mambocab :)

sstable_generation_loading_test.py
```diff
@@ -236,3 +184,123 @@ def read_and_validate_data(session):
...
+                for ddir in os.listdir(data_dir):
+                    keyspace_dir = os.path.join(data_dir, ddir)
+                    temp_files = self.glob_data_dirs(os.path.join(keyspace_dir, '*', "tmp", "*.dat"))
+                    debug("temp files: " + str(temp_files))
+                    self.assertEquals(0, len(temp_files), "Temporary files were not cleaned up.")
+
+
+@since('3.0')
+class TestLoadKaSStables(BaseSStableLoaderTest):
+    __test__ = True
```
mambocab: This configuration code is much more readable to me, and I think the single inheritance is much easier to reason about :+1:

upgrade_tests/storage_engine_upgrade_test.py
```diff
@@ -1,15 +1,23 @@
...
+import os
 
 from assertions import assert_all, assert_none, assert_one
-from dtest import Tester
-from tools import known_failure, since
+from dtest import Tester, debug
+from tools import known_failure, since, new_node
 
+LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
+                            "-Dcassandra.streamdes.max_mem_buffer_size=5  ",
```
mambocab: There's another unnecessary trailing space in this string.

mambocab:
Thanks for your patience on this PR. The code looks good, addressing my (tiny, tiny) nit. I'm running it locally now.

mambocab:
Oh, also, could you please rebase? Sorry to generate more work for you by sitting on this PR.

mambocab:
Tests pass locally. I'm +1, but @knifewine could you give this a quick once-over? I'm pretty confident in my assessment here but I've read it several times in the past few weeks and would like some fresh eyes on it. Don't spend too much time; just look for any glaring stuff I'm missing. Should be fine, just want a little insurance.

dtest.py
```diff
@@ -445,6 +446,14 @@ def _log_error_handler(self, errordata):
             # catch to raise an exception with useful information
             thread.interrupt_main()
 
+    def glob_data_dirs(self, path, ks="ks"):
```
knifewine: A brief docstring on this would be nice, it's purpose isn't quite obvious to me.

knifewine:
I'm not really familiar with these tests, but at a quick glance looks ok. One small nit is for a docstring, but that's it.

mambocab:
Great. @pauloricardomg we're done with our review -- go ahead and rebase, remove that unneccessary space in one of the strings, and add a docstring to your method, then feel free to merge. Thanks for the tests!

ptnapoleon:
Bump, see Jim's comment ^

pauloricardomg:
Addressed nits and triggered new dtest: http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/4/ (still running)

ptnapoleon:
Looks like you need to make a tiny change to your dtest, as it failed on CI

pauloricardomg:
Fixed nit and resubmitted :pray:

ptnapoleon:
+1!

pauloricardomg:
Thank you all! (I don't have commit rights I guess)

mambocab:
\o/

PR #859, opened 2016-03-16
Extended test for CASSANDRA-9043

stef1927:
I've added a text column to the primary key of `test_reading_counter`, which was introduced by CASSANDRA-9043. For counter tables we need to test values that require quotes as well: there was an issue with missing quotes that was introduced by CASSANDRA-9302 and fixed by CASSANDRA-11053. Unfortunately this problem found its way to a customer and so I would like to ensure this does not happen again in future.


ptnapoleon:
+1

PR #860, opened 2016-03-16
Use 127.0.0.1 instead of 'localhost' for dual-stacked systems

mshuler:
On systems that have dual-stack IPv4/IPv6, the name 'localhost' may
prefer IPv6 and resolve to ::1 instead of the expected 127.0.0.1.

### DIFF
--- a/pushed_notifications_test.py
+++ b/pushed_notifications_test.py
@@ -109,28 +109,29 @@ class TestPushedNotifications(Tester):
             self.assertEquals("MOVED_NODE", change_type)
             self.assertEquals(self.get_ip_from_node(node1), address)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11057')
     @no_vnodes()
     def move_single_node_localhost_test(self):
         """
         @jira_ticket  CASSANDRA-10052
         Test that we don't get NODE_MOVED notifications from nodes other than the local one,
-        when rpc_address is set to localhost.
+        when rpc_address is set to localhost (127.0.0.1).
 
-        To set-up this test we override the rpc_address to "localhost" for all nodes, and
+        To set-up this test we override the rpc_address to "localhost (127.0.0.1)" for all nodes, and
         therefore we must change the rpc port or else processes won't start.
         """
         cluster = self.cluster
         cluster.populate(3)
         node1, node2, node3 = cluster.nodelist()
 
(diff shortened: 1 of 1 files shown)


knifewine:
+1 if this passes on a ctool instance with whatever ipv6 sorcery is enabled there.

mshuler:
I'll run a test on my branch - good call!

pushed_notifications_test.py
```diff
@@ -109,27 +109,25 @@ def move_single_node_test(self):
...
-        To set-up this test we override the rpc_address to "localhost" for all nodes, and
+        To set-up this test we override the rpc_address to "localhost (127.0.0.1)" for all nodes, and
         therefore we must change the rpc port or else processes won't start.
         """
         cluster = self.cluster
         cluster.populate(3)
         node1, node2, node3 = cluster.nodelist()
 
-        # change node3 'rpc_address' from '127.0.0.x' to 'localhost', increase port numbers
+        # change node3 'rpc_address' from '127.0.0.x' to 'localhost (127.0.0.1)', increase port numbers
```
stef1927: This is an old typo, it should be `node's 'rpc_address'`, could you please fix it?

pushed_notifications_test.py
```diff
@@ -109,27 +109,25 @@ def move_single_node_test(self):
...
         i = 0
         for node in cluster.nodelist():
-            node.network_interfaces['thrift'] = ('localhost', node.network_interfaces['thrift'][1] + i)
-            node.network_interfaces['binary'] = ('localhost', node.network_interfaces['thrift'][1] + 1)
+            node.network_interfaces['thrift'] = ('127.0.0.1', node.network_interfaces['thrift'][1] + i)
```
stef1927: The thrift host is also used to set the `rpc_address` in the yaml file, see `__update_yaml()` in node.py.  In theory `127.0.0.1` should also work to test the patch of CASSANDRA-10052. However, we ship a default cassandra.yaml with `localhost`, not `127.0.0.1` and it would be preferable to continue testing with `localhost` in the yaml file. I think we just need to pass the `rpc_address` that we want to `node.set_configuration_options()`  before importing the config and it should work.

stef1927: Sorry I was a bit too quick and I wrongly assumed the address resolution was wrong in the test code, not the Cassandra process. Can you see if by any chance `-Djava.net.preferIPv6Addresses` was set to `true`? If it is `false` (the default) then we should not resolve to an IPV6 address and it looks like a bug to me.

PR #862, opened 2016-03-16
Don't specify the order of endpoints in repair log grep

ptnapoleon:
Fixes CASSANDRA-11359

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -615,8 +615,15 @@ class TestRepair(Tester):
         node1.repair(opts)
 
         self.assertEqual(len(node1.grep_log('are consistent for standard1')), 0, "Nodes 1 and 2 should not be consistent.")
-        self.assertEqual(len(node1.grep_log("/127.0.0.2 and /127.0.0.1 have ([0-9]+) range\(s\) out of sync")), 1)
         self.assertEqual(len(node3.grep_log('Repair command')), 0, "Node 3 should not have been involved in the repair.")
+
+        out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
+        _, matches = out_of_sync_logs[0]
+        out_of_sync_nodes = {matches.group(1), matches.group(2)}
+
+        valid_out_of_sync_pairs = [{node1.address(), node2.address()}]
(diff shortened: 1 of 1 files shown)


repair_tests/repair_test.py
```diff
@@ -615,9 +615,16 @@ def _parameterized_range_repair(self, repair_opts):
         node1.repair(opts)
 
         self.assertEqual(len(node1.grep_log('are consistent for standard1')), 0, "Nodes 1 and 2 should not be consistent.")
-        self.assertEqual(len(node1.grep_log("/127.0.0.2 and /127.0.0.1 have ([0-9]+) range\(s\) out of sync")), 1)
         self.assertEqual(len(node3.grep_log('Repair command')), 0, "Node 3 should not have been involved in the repair.")
 
+        out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
+        valid_out_of_sync_pairs = [{node1.address(), node2.address()}]
```
mambocab: Since there's only one valid pair here, you can just let this be a set and use `assertEqual` below instead of `assertIn`.

repair_tests/repair_test.py
```diff
@@ -615,9 +615,16 @@ def _parameterized_range_repair(self, repair_opts):
...
+
+        line, matches = out_of_sync_logs[0]
+        out_of_sync_nodes = {matches.group(1), matches.group(2)}
+
+        self.assertIn(out_of_sync_nodes, valid_out_of_sync_pairs, str(out_of_sync_nodes))
```
mambocab: is the `str` bit necessary?

mambocab:
Line ordering is a little weird; I'd do something more like

```python
out_of_sync_logs = node1.grep_log("/([0-9.]+) and /([0-9.]+) have ([0-9]+) range\(s\) out of sync")
_, matches = out_of_sync_logs[0]
out_of_sync_nodes = {matches.group(1), matches.group(2)}

valid_out_of_sync_pair = {node1.address(), node2.address()}
```

Maybe also check that `len(out_of_sync_logs) == 1`, if that's something we expect to be true.

mambocab:
Other than my lil comments, I'm +1. I haven't run the code, but I can if you want a more thorough review.

PR #865, opened 2016-03-17
Test and changes for CASSANDRA-11255

stef1927:
A new test for CASSANDRA-11255 and some changes due to a new default precision for double values in COPY TO.


ptnapoleon:
+1

PR #866, opened 2016-03-17
gc_test tag and improvments

mambocab:
I needed to tag this, and the docstring was old and sparse, so I figured I'd clean things up a bit. I've

- added a tag
- run isort
- fleshed out the docstring
- made stronger assertions using `assertEqual` instead of a bare `assert`

Runs fine for me locally.


knifewine:
+1

mambocab:
Cheers, merging.

PR #867, opened 2016-03-17
Fix auth race condition

mambocab:
@ptnapoleon to review. Main fix is to prevent flakes like this in the future:

http://cassci.datastax.com/job/cassandra-3.0_dtest/607/testReport/junit/auth_test/TestAuthRoles/grant_revoke_roles_test/

I also removed some unused variables that `flake8` caught. @ptnapoleon I pushed this to the `riptano` repo so you could rebase out the `flake8 fixes` commit and actually use those `usercount` variables if you want.

### DIFF
--- a/auth_test.py
+++ b/auth_test.py
@@ -3,12 +3,13 @@ import re
 import time
 from collections import namedtuple
 
-from assertions import (assert_all, assert_invalid, assert_one,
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

mambocab:
Cheers, merging.

PR #868, opened 2016-03-17
remove known_failure annotation

mambocab:
Removing the annotation since the ticket was closed. Do we know why we skip this pre-2.2? If it hangs or something we should skip it.

### DIFF
--- a/replace_address_test.py
+++ b/replace_address_test.py
@@ -226,7 +226,12 @@ class TestReplaceAddress(Tester):
                    flaky=False,
                    notes='fails hard on linux')
     def resumable_replace_test(self):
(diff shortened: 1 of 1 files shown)


mambocab:
We should *document* and skip it, that is.

mambocab:
Didn't investigate this enough before opening this PR; there's a new ticket in town. Question about why it's skipped still stands.

ptnapoleon:
This should mention `https://issues.apache.org/jira/browse/CASSANDRA-8838` in its docstring, which is also why we don't run pre 2.2

mambocab:
Thanks, I've added that information to the docstring and reordered the decorators so the tags take.

PR #869, opened 2016-03-18
Additional test for CASSANDRA-11053

stef1927:
A new test to cover a cqlsh COPY issue that was discovered on a machine with a single core. Please do not merge until CASSANDRA-11053 is resolved.

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -2373,6 +2373,18 @@ class CqlshCopyTest(Tester):
         self._test_bulk_round_trip(nodes=3, partitioner="murmur3", num_operations=10000,
                                    copy_from_options={'INGESTRATE': 1500})
 
+    @freshCluster()
+    def test_bulk_round_trip_with_single_core(self):
+        """
+        Perform a round trip on a simulated single core machine. When determining the number of cores,
+        copyutil.py will return the number carried by the environment variable CQLSH_COPY_TEST_NUM_CORES if it has
+        been set.
+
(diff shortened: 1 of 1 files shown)


mambocab:
Does this just assert that the operations don't fall over when running with a single core? That's fine, but if it'd be a good extension to the test I can file a followup ticket to somehow assert that the operations use the correct number of cores.

stef1927:
By setting the test environment variable, we can be reasonably confident that the number of cores used by CQLSH COPY is 1. We could verify that we use only 1 child process, but it's not exactly the same thing. To verify the number of cores used, we need to add a debug message printing the number of cores detected and then parse the output. The verification of the number of cores is definitely a good thing if you want to file a ticket. 

The purpose of the test is to really run CQLSH COPY with a boundary condition, 1 core, which has caused problems, basically it was not working at all on a single core VM.

Do we normally test CQLSH on very simple VMs before releasing? If so, then this test is perhaps redundant.

ptnapoleon:
I don't think the test is redundant. We do not test on one core VMs before release, or possibly ever.

mambocab:
I agree with Philip. I've filed https://issues.apache.org/jira/browse/CASSANDRA-11437 to get more visibility for the behavior of the test environment variable.

stef1927:
CASSANDRA-11053 has been resolved, shall we merge this pull request or shall we wait for the enhanced version delivered by CASSANDRA-11437?

mambocab:
Thanks for the ping -- nothing wrong with merging now. CASSANDRA-11437 will give us greater confidence, but this gives us some. Thanks!

PR #870, opened 2016-03-18
Fix node4 start in resumable_replace_address

ptnapoleon:
We expect it to fail to start up, and thus we need to
set wait_other_notice to false


mambocab:
Yep, +1.

PR #871, opened 2016-03-18
Fix node start in decommissioned_node_cant_rejoin_test

ptnapoleon:
Node is not expected to start up. Set wait_other_notice to False


mambocab:
If you remove the `known_failure`, I'm +1

PR #872, opened 2016-03-18
Fix node start in crash_during_decommission

ptnapoleon:
Node1 is being decommissioned while we restart node2. thus we
cannot guarantee that node1 will be alive to detect node2 is up.
So we must start node2 with wait_other_notice=False each time


mambocab:
+1. We should add this to a ticket somewhere to consider moving to Fallout.

PR #876, opened 2016-03-21
changes for 11256 stability

knifewine:


knifewine:
depends on https://github.com/pcmanus/ccm/pull/475 as well

knifewine:
I think these changes should be low impact to bring in, then I can do another repetition test before removing the known_failure tag.

PR #879, opened 2016-03-21
better assert error message

knifewine:
small change for more readable assert

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -62,7 +62,7 @@ class TestRepair(Tester):
 
         session = self.patient_cql_connection(node_to_check, 'ks')
         result = list(session.execute("SELECT * FROM cf LIMIT {}".format(rows * 2)))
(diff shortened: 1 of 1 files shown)


repair_tests/repair_test.py
```diff
@@ -62,7 +62,7 @@ def check_rows_on_node(self, node_to_check, rows, found=None, missings=None, res
 
         session = self.patient_cql_connection(node_to_check, 'ks')
         result = list(session.execute("SELECT * FROM cf LIMIT {}".format(rows * 2)))
-        self.assertEqual(len(result), rows, len(result))
+        self.assertEqual(len(result), rows, "expected {}, got {}".format(rows, len(result)))
```
ptnapoleon: If you just removed the message entirely, I am fairly sure this is the default output message. However, could your make it show `result` and not `len(result)`? That would help point us to what the extra or missing rows are.

PR #882, opened 2016-03-22
Set default CL to QUORUM in complex_mv_select_statements_test

ptnapoleon:
Seeing consistency issues with write at ONE then immediate
read at QUORUM

@knifewine or @mambocab to review


knifewine:
+1

PR #883, opened 2016-03-22
pause during do_upgrade

knifewine:
@mambocab to review

### DIFF
--- a/upgrade_tests/paging_test.py
+++ b/upgrade_tests/paging_test.py
@@ -1125,9 +1125,6 @@ class TestPagingDatasetChanges(BasePagingTester, PageAssertionMixin):
 
             self.assertEqualIgnoreOrder(pf.all_data(), expected_data)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11396',
-                   flaky=True)
     def test_data_change_impacting_later_page(self):
         cursor = self.prepare()
         cursor.execute("CREATE TABLE paging_test ( id int, mytext text, PRIMARY KEY (id, mytext) )")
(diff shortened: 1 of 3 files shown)


upgrade_tests/paging_test.py
```diff
@@ -6,7 +6,6 @@
 from cassandra import ConsistencyLevel as CL
 from cassandra import InvalidRequest, ReadFailure, ReadTimeout
 from cassandra.query import SimpleStatement, dict_factory, named_tuple_factory
-
```
mambocab: I don't think this line should come out.

upgrade_tests/upgrade_base.py
```diff
@@ -8,7 +8,6 @@
 from unittest import skipIf
 
 from ccmlib.common import get_version_from_build, is_win
-
```
mambocab: I don't think this line should come out either.

upgrade_tests/upgrade_base.py
```diff
@@ -199,7 +198,10 @@ def do_upgrade(self, session):
             for is_upgraded, session in sessions:
                 session.default_consistency_level = self.CL
 
-        return sessions
+        # let the upgraded node settle briefly, and help lower any perf noise from the previous iterations
```
mambocab: Since this behavior is sort of weird, I think a little more explanation, and maybe a reference to CASSANDRA-11396, is in order in this comment.

mambocab: In particular, an explanation of the bugs we ran into would be good.

mambocab:
Code is good; I can re-review after you expand the comment a bit.

knifewine:
@mambocab ready for another look

mambocab:
+1, nice

PR #887, opened 2016-03-23
Add test for CASSANDRA-10091

beobal:


jmx_auth_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+@since(3.6)
+class TestJMXAuth(Tester):
+
+    def __init__(self, *args, **kwargs):
+        self.ignore_log_patterns = [
+            # This one occurs if we do a non-rolling upgrade, the node
+            # it's trying to send the migration to hasn't started yet,
+            # and when it does, it gets replayed and everything is fine.
+            r'Can\'t send migration request: node.*is down',
```
mambocab: I can't find it where this message is generated on [the branch that I think you're testing](https://github.com/beobal/cassandra/tree/10091-trunk), but I'm guessing this could be a stricter regex, e.g., by replacing `node.*is` with `node\d+ is`. If you can make it stricter, that'd be nice.

jmx_auth_test.py
```diff
@@ -0,0 +1,67 @@
...
+from dtest import Tester
+from jmxutils import apply_jmx_authentication
+from tools import since
+
+
+@since(3.6)
+class TestJMXAuth(Tester):
+
+    def __init__(self, *args, **kwargs):
+        self.ignore_log_patterns = [
```
mambocab: I know there are a bunch of places where we use the `def __init__(...): self.ignore_log_patterns = ...` pattern, but if that's all you do in the method, it's unnecessary. You can just set `ignore_log_patterns` as a class-level variable, like [in `offline_tools_test`](https://github.com/riptano/cassandra-dtest/blob/master/offline_tools_test.py#L17). I'll file a ticket for one of us on the TE team to remove `__init__` methods that set a bad precedent this way.

mambocab: So, if you're so moved, could you please make that refactoring? If not, don't worry, we'll get it when we address the above-mentioned ticket.

jmx_auth_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # test doesn't yet have any privileges on the necessary JMX resources
+        with self.assertRaisesRegexp(NodetoolError, 'Access Denied'):
+            node.nodetool('-u test -pw abc123 gossipinfo')
+
+        session.execute("GRANT jmx_user TO test")
+        node.nodetool('-u test -pw abc123 gossipinfo')
+
+        # superuser status applies to JMX authz too
+        node.nodetool('-u cassandra -pw cassandra gossipinfo')
```
mambocab: Could you please add assertions that the `test` and `cassandra` users can't use `nodetool gossipinfo` with bad passwords? That increases coverage a bit, and makes it clearer that the "assertion" here is that `nodetool` runs without raising an error.

Could you also add an assertion that `nodetool gossipinfo` without a `-u` fails?

jmxutils.py
```diff
@@ -51,6 +51,28 @@ def make_mbean(package, type, **kwargs):
...
+    replacement = 'JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=true"'
+    pattern1 = 'JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.password.file=/etc/cassandra/jmxremote.password"'
+    replacement1 = '#JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.password.file=/etc/cassandra/jmxremote.password"'
+    pattern2 = '#JVM_OPTS="\$JVM_OPTS -Dcassandra.jmx.remote.login.config=CassandraLogin"'
+    replacement2 = 'JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.remote.login.config=CassandraLogin"'
+    pattern3 = '#JVM_OPTS="\$JVM_OPTS -Djava.security.auth.login.config=\$CASSANDRA_HOME/conf/cassandra-jaas.config"'
+    replacement3 = 'JVM_OPTS="$JVM_OPTS -Djava.security.auth.login.config=$CASSANDRA_HOME/conf/cassandra-jaas.config"'
+    pattern4 = '#JVM_OPTS="\$JVM_OPTS -Dcassandra.jmx.authorizer=org.apache.cassandra.auth.jmx.AuthorizationProxy"'
+    replacement4 = 'JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.authorizer=org.apache.cassandra.auth.jmx.AuthorizationProxy"'
+    replacement_list = [(pattern, replacement), (pattern1, replacement1), (pattern2, replacement2),
```
mambocab: This would be more readable in a multiline string literal:

```python
replacement_list = [
    ('JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=false"',
     'JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=true"'),
    ('JVM_OPTS="\$JVM_OPTS -Dcom.sun.management.jmxremote.password.file=/etc/cassandra/jmxremote.password"',
     '#JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.password.file=/etc/cassandra/jmxremote.password"'),
    ('#JVM_OPTS="\$JVM_OPTS -Dcassandra.jmx.remote.login.config=CassandraLogin"',
     'JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.remote.login.config=CassandraLogin"'),
    ('#JVM_OPTS="\$JVM_OPTS -Djava.security.auth.login.config=\$CASSANDRA_HOME/conf/cassandra-jaas.config"',
     'JVM_OPTS="$JVM_OPTS -Djava.security.auth.login.config=$CASSANDRA_HOME/conf/cassandra-jaas.config"'),
    ('#JVM_OPTS="\$JVM_OPTS -Dcassandra.jmx.authorizer=org.apache.cassandra.auth.jmx.AuthorizationProxy"',
     'JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.authorizer=org.apache.cassandra.auth.jmx.AuthorizationProxy"')
]
```

(Double-check that I did the correct substitutions before using, I may not have done it perfectly, but: I think that's more readable, personally.)

jmx_auth_test.py
```diff
@@ -0,0 +1,67 @@
+from ccmlib.node import NodetoolError
+from dtest import Tester
+from jmxutils import apply_jmx_authentication
+from tools import since
+
+
+@since(3.6)
```
mambocab: `since` takes a string.

jmxutils.py
```diff
@@ -51,6 +51,28 @@ def make_mbean(package, type, **kwargs):
     return rv
 
 
+def apply_jmx_authentication(node):
+    if common.is_win():
+        conf_file = os.path.join(node.get_conf_dir(), common.CASSANDRA_WIN_ENV)
+    else:
+        conf_file = os.path.join(node.get_conf_dir(), common.CASSANDRA_ENV)
```
mambocab: I've got a PR in flight on ccm that should make this code unnecessary:

https://github.com/pcmanus/ccm/pull/482

If it gets merged before this PR, I'll ping you to change this code.

mambocab:
Address my comments and I'm +1.

mambocab:
Ah, just realized that you're shepherding this PR, you're not its primary author @beobal. If you like, I can make some commits for you to cherry-pick, rather than making you do the work here. Just let me know.

jmx_auth_test.py
beobal: Done (I don't think it was necessary anyway, probably a vestigal artifact of copy+pasting from an existing test).

jmx_auth_test.py
beobal: There's already an assertion for `test`  with a bad password (line 42 above), do you think it's worth duplicating that with `cassandra` + bad password?

I've added the missing `-u` assertion, thanks for catching that one.

jmxutils.py
beobal: yep, looks more readable to me too, thanks.

jmx_auth_test.py
beobal: d'oh

jmxutils.py
beobal: sgtm

beobal:
> just realized that you're shepherding this PR

no worries, I've made the changes you suggested, modulo one which I think is redundant. Let me know what you think, and I'll squash and merge when you're happy. FTR, I'll hold off until 10091 is committed though.

jmx_auth_test.py
mambocab: The "`cassandra` with bad password" use case doesn't need to be tested, you're right. My thought about adding an assertion for the `test` user is to make sure it still doesn't allow use with a bad password after the change in permissions, but yeah, that's probably also redundant :+1:

mambocab:
Just to confirm, did you mean to remove `ignore_log_patterns` entirely?

beobal:
I did, I don't think it's necessary, Am I thinking about this wrong and this is required boilerplate for upgrade testing?

mambocab:
> Am I thinking about this wrong

Nope, just making sure I'm correctly interpreting what you wrote.

Looks good to me, squash away, and merge when ready. Thanks!

mambocab:
https://github.com/pcmanus/ccm/pull/482 got merged -- when you squash, could you also use the new `envfilename` method, rather than building the `cassandra.env` path yourself?

beobal:
CASSANDRA-10091 has been committed so I've rebased & merged, with @mambocab's suggestion to use `envfilename`.

mambocab:
:+1: Great, thanks!

PR #888, opened 2016-03-23
Change SCUpgrade tests to assume we are always on 2.1

ptnapoleon:
@mambocab to review.

We don't test on 2.0 anymore, so we should step always assume we're on 2.1. Easier than doing complicated version checking code.


mambocab:
+1, merge away

PR #892, opened 2016-03-27
join_ring=False also needs wait_other_notice=False

michaelsembwever:
A node started with "-Dcassandra.join_ring=false" will not get recognised as UP by the other nodes.

That is `node.start(join_ring=False)` needs to be `node.start(join_ring=False, wait_other_notice=False)`

ref:
 - https://github.com/pcmanus/ccm/pull/479
 - http://cassci.datastax.com/job/trunk_dtest/1076/testReport/

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -636,7 +636,7 @@ class AbortedQueriesTester(CQLTester):
         cluster.populate(2)
         node1, node2 = cluster.nodelist()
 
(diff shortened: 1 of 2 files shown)


ptnapoleon:
-1. See https://github.com/pcmanus/ccm/pull/479#issuecomment-202502244

PR #893, opened 2016-03-28
Add bootstrap tests with pre-3.0 sstables

pauloricardomg:
CASSANDRA-10990

### DIFF
--- a/upgrade_tests/storage_engine_upgrade_test.py
+++ b/upgrade_tests/storage_engine_upgrade_test.py
@@ -1,15 +1,23 @@
 import time
+import os
 
 from assertions import assert_all, assert_none, assert_one
-from dtest import Tester
-from tools import since
+from dtest import Tester, debug
+from tools import since, new_node
+
+
+LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
+                            "-Dcassandra.streamdes.max_mem_buffer_size=5  ",
+                            "-Dcassandra.streamdes.max_spill_file_size=128"]
 
 
 @since('3.0')
 class TestStorageEngineUpgrade(Tester):
-    def setUp(self):
+    def setUp(self, bootstrap=False, jvm_args=[]):
         super(TestStorageEngineUpgrade, self).setUp()
         self.default_install_dir = self.cluster.get_install_dir()
+        self.bootstrap = bootstrap
+        self.jvm_args = jvm_args
 
     def _setup_cluster(self, create_keyspace=True):
         cluster = self.cluster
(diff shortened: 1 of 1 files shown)


mambocab:
Sorry to neglect this for a week -- does this replace #858?

upgrade_tests/storage_engine_upgrade_test.py
```diff
@@ -1,15 +1,23 @@
...
 
 from assertions import assert_all, assert_none, assert_one
-from dtest import Tester
-from tools import since
+from dtest import Tester, debug
+from tools import since, new_node
+
+
+LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
+                            "-Dcassandra.streamdes.max_mem_buffer_size=5  ",
```
mambocab: Is there a reason for the whitespace at the end of this string? If there is, could you please document it, and if not, could you please remove it?

upgrade_tests/storage_engine_upgrade_test.py
```diff
@@ -357,3 +379,9 @@ def upgrade_with_range_and_collection_tombstones_test(self):
         self.cluster.compact()
 
         assert_one(session, "SELECT k FROM t", ['some_key'])
+
+
+@since('3.0')
+class TestBootstrapAfterUpgrade(TestStorageEngineUpgrade):
+    def setUp(self):
+        super(TestBootstrapAfterUpgrade, self).setUp(bootstrap=True, jvm_args=LEGACY_SSTABLES_JVM_ARGS)
```
mambocab: Since `bootstrap` and `jvm_args` are just setting `self.bootstrap` and `self.jvm_args`, you can just set `bootstrap` and `jvm_args` as class variables on this class and not worry about overriding the `setUp` method in this class.

upgrade_tests/storage_engine_upgrade_test.py
```diff
@@ -1,15 +1,23 @@
...
+
+LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
+                            "-Dcassandra.streamdes.max_mem_buffer_size=5  ",
+                            "-Dcassandra.streamdes.max_spill_file_size=128"]
 
 
 @since('3.0')
 class TestStorageEngineUpgrade(Tester):
-    def setUp(self):
+    def setUp(self, bootstrap=False, jvm_args=[]):
```
mambocab: Since `[]` is mutable, could you please change this to

```python
def setUp(self, bootstrap=False, jvm_args=None):
    ...
    if jvm_args is None:
        jvm_args = []
```

Basically, default argument values are created at function-definition time, and any changes to them will persist across calls. This is explained in more detail [in this article](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments).

pauloricardomg:
Closing this because it will be included on #858.

PR #894, opened 2016-03-28
Skip all 6+ node tests on CI

ptnapoleon:
@mambocab and @exlt to review.

Testing here:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-cassandra-2.1-dtest/12/console

All job configs will need changed to use the fast-select.cfg instead of test-select.cfg

### DIFF
--- a/conf/cassandra-2.1_test-select.cfg
+++ b/conf/cassandra-2.1_test-select.cfg
@@ -1,1 +1,14 @@
 [exclude]
+# Subsequent repair is a long test, ~40-50 minutes
+incremental_repair_test.py:TestIncRepair.multiple_subsequent_repair_test
+  tools.py:TestIncRepair.multiple_subsequent_repair_test
+consistency_test.py:TestAvailability.test_network_topology_strategy
+consistency_test.py:TestAvailability.test_network_topology_strategy_each_quorum
+  tools.py:TestAvailability.test_network_topology_strategy_each_quorum
+consistency_test.py:TestAccuracy.test_network_topology_strategy_users
+consistency_test.py:TestAccuracy.test_network_topology_strategy_each_quorum_users
+  tools.py:TestAccuracy.test_network_topology_strategy_each_quorum_users
+consistency_test.py:TestAccuracy.test_network_topology_strategy_each_quorum_counters
+  tools.py:TestAccuracy.test_network_topology_strategy_each_quorum_counters
+consistency_test.py:TestAccuracy.test_network_topology_strategy_counters
+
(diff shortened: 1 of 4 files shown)


conf/cassandra-2.1_fast-select.cfg
```diff
@@ -2,3 +2,12 @@
 # Subsequent repair is a long test, ~40-50 minutes
 incremental_repair_test.py:TestIncRepair.multiple_subsequent_repair_test
   tools.py:TestIncRepair.multiple_subsequent_repair_test
+consistency_test.py:TestAvailability.test_network_topology_strategy
+consistency_test.py:TestAvailability.test_network_topology_strategy_each_quorum
+  tools.py:TestAvailability.test_network_topology_strategy_each_quorum
```
mambocab: What's up with this?

mambocab: In particular, why are there tests in the `tools` module?

mambocab:
Based on my reading of the bucketizer output from the linked run, it looks fine. +1, though I could use an explanation of the "tests in the `tools` module" bit for my own edification.

conf/cassandra-2.1_fast-select.cfg
ptnapoleon: So the nose exclude mechanism is really, really smart and has great UX. It's so smart, that if a test is annotated, for example with `since`, or `require`, then you need this second line to specify the file containing the annotation, which here is `tools.py`.

mambocab:
Though, why not just add stuff to the `test-select` config files and avoid messing with existing jobs?

conf/cassandra-2.1_fast-select.cfg
mambocab: Gotcha.

ptnapoleon:
I guess I don't have a good reason? I just used `fast-select` since they seemed more appropriate based on their names. But if we aren't using test-select in any meaningful capacity, let's just add it there? What does @mshuler  think?

conf/cassandra-2.1_fast-select.cfg
ptnapoleon: For reference, I believe this behavior is a "bug", but we've never found or solved it in the exclude plugin.

mshuler:
I think *_fast-select.cfg is OK - it means changing up job templates from the "default" test-select configs to something else. My initial plan was to just add this to the test-select configs we're already using. I think fast-select was used to make some scratch test job runs quite a bit shorter in duration.

ptnapoleon:
Are you against just putting this in test-select? That's what I'm leaning towards as the better idea

mshuler:
I'm totally fine with using test-select for this and using test-select was my initial thought. My preference would be to just add these to the default `$branch_test-select.cfg` files we're already using, instead of using the fast-select configs (which are unused and were added to the repo as a quick hack to run some scratch jobs faster). No job template updates needed, existing dev jobs will pick up the exclude, etc.

ptnapoleon:
Okay, moved over to *test-select.cfg

mshuler:
+1!

PR #895, opened 2016-03-29
cl increase for cassandra-11395

knifewine:


knifewine:
@ptnapoleon to review. whole_map_conditional_test was the one still flaking a bit, but I think it may have been CI noise. I managed 500 runs without a single flap here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/49/testReport/

ptnapoleon:
+1

PR #896, opened 2016-03-29
Cassandra 11394

mambocab:
This won't break the existing test, but there's a companion PR over at https://github.com/pcmanus/ccm/pull/484. This is kind of a debugging step for some flaps we've seen.


mambocab:
Pushed with a new, more detailed, commit message.

ptnapoleon:
@mambocab and @knifewine is this something we should have merged, still merge, review, etc?

mambocab:
I don't know if this is useful any longer, this test hasn't flaked in 3 months. Thoughts?

ptnapoleon:
I'm okay closing then

mambocab:
Fab, thank you.

PR #897, opened 2016-03-29
remove known for CASSANDRA-11395

knifewine:
forgot to pull these in my pr with the fix.


mambocab:
Looks good, if a smoke test of just running the module successfully runs tests, then +1.

knifewine:
looks gtg, thanks @mambocab

PR #899, opened 2016-03-29
Support changed log units after CASSANDRA-9692 in compaction tests

jkni:
[CASSANDRA-9692](https://issues.apache.org/jira/browse/CASSANDRA-9692) normalizes logging output and pretty-prints rates/sizes in many places.

This breaks two tests in compaction_test.py, which match on the units with regexes. These have been extended to support both old and new-style rates, splitting behavior on version 3.6 (current trunk).

Please review but don't merge until CASSANDRA-9692 is commited; this will minimize the length of time for which the tests fail (since trunk is currently 3.6 and does not have these logging changes).

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -198,11 +198,12 @@ class TestCompaction(Tester):
 
         matches = block_on_compaction_log(node1)
         stringline = matches[0]
-        throughput_pattern = re.compile('''.*          # it doesn't matter what the line starts with
-                                           =           # wait for an equals sign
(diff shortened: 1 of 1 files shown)


jkni:
CASSANDRA-9692 was commited to [trunk](https://github.com/apache/cassandra/commit/db68ac9fd791d777df0241ef9bd381eca46bbd20), so this is ready for review and merge. (In fact, these dtests will fail on trunk until it is merged.)

ptnapoleon:
+1

PR #903, opened 2016-03-30
filter out annoying "connection failed" messages

mambocab:
I'm not sure whether or not to be concerned about false positives here -- will there be cases where we want this debug information? But, at the very least, this should remove a whole lot of useless error messages from test output.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -20,6 +20,7 @@ import types
 from collections import OrderedDict
 from unittest import TestCase
 
+from cassandra import cluster as cluster_module
 from cassandra import ConsistencyLevel
 from cassandra.auth import PlainTextAuthProvider
 from cassandra.cluster import Cluster as PyCluster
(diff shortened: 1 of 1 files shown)


mambocab:
(@ptnapoleon and @knifewine both to review, please -- like I said, I'm concerned about false positives and would like both of your insight.)

ptnapoleon:
I mean, the case *may* exist where the connection really is going to fail, and this debug info might be useful? But I think that's so rare, and we'll know that the connection failed anyway, that if it becomes a problem, we can deal with it then?

I don't expect false positives in terms of "we think connection worked but it didn't", this just might mask why it fails, which is almost always obvious anyway.

I'm +1

knifewine:
This looks ok to me. Another option would be just to adjust the log level up to something like critical, then bump back down after getting a connection. We do something similar over here: https://github.com/riptano/cassandra-dtest/blob/master/dtest.py#L257

I'm good with either approach. If we decide we like the one in this pr better, maybe we adjust the code I linked above to use a similar style.

mambocab:
> Another option would be just to adjust the log level up to something like critical, then bump back down after getting a connection... If we decide we like the version in this pr better, maybe we adjust the code I linked above to use a similar style.

My kneejerk reaction to this is that each is appropriate for its purpose -- in one case, we wanna just say "sssshhhhhhhhhh", in the other, we expect very specific messages and don't care when they're logged.

mambocab:
Ok, I've pushed a couple changes which I'd like re-reviewed, please @ptnapoleon @knifewine.

mambocab:
> ... we'll know that the connection failed anyway, that if it becomes a problem, we can deal with it then?... this just might mask why it fails, which is almost always obvious anyway.

My thoughts exactly.

knifewine:
+1

mambocab:
Thanks for the reviews, y'all. Merging.

PR #904, opened 2016-03-31
sleep before getting sstables after flush

mambocab:
This was added in the scrub method in `ee436c340946240ff4995277d5114db59b8bfd5d`, and maybe should be added here as well. This is to attempt to address https://issues.apache.org/jira/browse/CASSANDRA-11446.

### DIFF
--- a/scrub_test.py
+++ b/scrub_test.py
@@ -124,7 +124,7 @@ class TestHelper(Tester):
         in a dict keyed by the table or index name.
         """
         self.perform_node_tool_cmd('flush', table, indexes)
-        time.sleep(.1)
+        time.sleep(1)
         return self.get_sstables(table, indexes)
 
     def scrub(self, table, *indexes):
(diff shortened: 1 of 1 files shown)


ptnapoleon:
If sleep is the only way to make this work, I'd rather sleep an entire second, so we never have to come back and increment this.

mambocab:
Doing a debugging run here to see if `.1` might be sufficient: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/54/

ptnapoleon:
bump?

mambocab:
Rebased onto master, then kicked this off:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/104/

to see if a 1-second sleep fixes anything. If it does, I should add a comment.

mambocab:
Doesn't seem to fix. Closing. @ptnapoleon @knifewine Any ideas?

PR #905, opened 2016-03-31
get more debug info in tracing_unknown_impl_test

mambocab:
Just adds some debug printing for https://issues.apache.org/jira/browse/CASSANDRA-11465.

### DIFF
--- a/cql_tracing_test.py
+++ b/cql_tracing_test.py
@@ -114,8 +114,8 @@ class TestCqlTracing(Tester):
         @jira_ticket CASSANDRA-10392
         """
         expected_error = 'Cannot use class junk for tracing'
-        session = self.prepare(jvm_args=['-Dcassandra.custom_tracing_class=junk'])
         self.ignore_log_patterns = [expected_error]
+        session = self.prepare(jvm_args=['-Dcassandra.custom_tracing_class=junk'])
         self.trace(session)
 
         errs = self.cluster.nodelist()[0].grep_log_for_errors()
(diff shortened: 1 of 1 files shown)


mambocab:
Trying it here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/55/

mambocab:
I believe I have fixed the problems that made the tests explode -- the log watcher picked up errors in `prepare` calls because `ignore_log_errors` wasn't set.

@knifewine This might be a good debugging case for the log watcher -- looks like the dtests generally don't do teardown right on `SIGINT`?

ptnapoleon:
+1 on the debug changes. Shouldn't we SIGTERM, not SIGINT?

mambocab:
No, I do mean `SIGINT`. `unittest` has an option to finish the current test on `^C`:

https://docs.python.org/2/library/unittest.html#cmdoption-unittest-c

Doesn't look like we can use it directly through `nose` but we might look into it.

knifewine:
+1 on the changes

not certain I understand how the log watcher is misbehaving, but wonder if it could be forcing main thread termination before it's been able to complete all of the teardown.

mambocab:
Yeah, I think that's what's happening. I don't know if the failure results reflect this, but maybe the log-watching function is being called during teardown, since the thread doesn't stop running until the main thread exits?

mambocab:
But, thank you for the review. Merging.

PR #908, opened 2016-04-01
Move network_topology_test out of default CI jobs

ptnapoleon:
It starts 6 nodes and is way too flaky.

@mshuler or @mambocab to review


mambocab:
Have you run this anywhere? If it's doing the intended thing, +1.

mshuler:
Yep, if a quick test worked out, this is fine :)

PR #910, opened 2016-04-04
CASSANDRA-11225: hardened consistency_test test_simple_strategy_counters

stef1927:
CASSANDRA-11225: hardened consistency_test.TestAccuracy.test_simple_strategy_counters
by performing a read at CL.ALL before each iteration and using property file snitch
to ensure coordinator reads locally if it is a replica. Enhanced error reporting.

PR #911, opened 2016-04-04
remove known_failure...

mambocab:
I think we're good to close everything down for this:

https://github.com/riptano/cassandra-dtest/pull/910
https://issues.apache.org/jira/browse/CASSANDRA-11225


ptnapoleon:
If the ticket is resolved, I am 100% okay with just removing these. 

+2

mambocab:
Thanks!

PR #916, opened 2016-04-06
CASSANDRA-11225: fixed comment on default snitch

stef1927:
My comment on the default snitch is slightly incorrect, see discussion on CASSANDRA-11488.


knifewine:
+1

PR #918, opened 2016-04-08
name the tests so they get run by nose

krummas:


mambocab:
Good catch, thanks. Merging.

PR #919, opened 2016-04-11
Fixed test for CASSANDRA-11529

stef1927:
An existing test needs changing due to CASSANDRA-11529. 

Commit only when the C* ticket is also committed.


ptnapoleon:
+1

PR #920, opened 2016-04-11
fix drop user case sensitive test

krummas:
compare with the correct part of users list

### DIFF
--- a/auth_test.py
+++ b/auth_test.py
@@ -240,7 +240,7 @@ class TestAuth(Tester):
         assert_invalid(cassandra, "DROP USER test")
 
         cassandra.execute("DROP USER Test")
-        rows = list(cassandra.execute("LIST USERS"))
+        rows = [x[0] for x in list(cassandra.execute("LIST USERS"))]
         self.assertItemsEqual(rows, ['cassandra'])
 
         cassandra.execute("CREATE USER test WITH PASSWORD '12345'")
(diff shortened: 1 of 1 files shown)


auth_test.py
```diff
@@ -240,16 +240,16 @@ def drop_user_case_sensitive_test(self):
         assert_invalid(cassandra, "DROP USER test")
 
         cassandra.execute("DROP USER Test")
-        rows = list(cassandra.execute("LIST USERS"))
-        self.assertItemsEqual(rows, ['cassandra'])
+        rows = [x[0] for x in list(cassandra.execute("LIST USERS"))]
+        self.assertItemsEqual(rows, ['cassandra'], rows)
```
ptnapoleon: I don't think the change on this line is necessary, and in fact, I suspect it would reduce the amount of debug output we get. The original line should print out the content of both items being compared, and the diff. With the new output of just `rows`, someone would need to look in the test code to see what is wrong.

ptnapoleon:
Where is this test failing?

auth_test.py
krummas: yeah, added it when debugging and forgot to remove

krummas:
http://cassci.datastax.com/view/trunk/job/trunk_dtest/1122/testReport/auth_test/TestAuth/drop_user_case_sensitive_test/

ptnapoleon:
+1

PR #921, opened 2016-04-11
Add paging tests for filtering with clustering columns for 11310

ifesdjeen:
Add paging tests for filtering with clustering, static and regular columns for [#11310](https://issues.apache.org/jira/browse/CASSANDRA-11310)

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -1085,6 +1085,238 @@ class TestPagingData(BasePagingTester, PageAssertionMixin):
             self.assertEqual(res, [[1, 1],
                                    [2, 2]])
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
All of the since annotations need to be `@since('3.6')` not `@since('3.6.0')`. Trunk currently evaluates to `3.6`, which nose evaluates to being less than `3.6.0`, meaning these won't run on the correct versions.

ptnapoleon:
Otherwise, +1

ifesdjeen:
@ptnapoleon my bad, going to fix that right away.

ifesdjeen:
Fixed. I'll post here whenever the #11310 is merged.

blerer:
CASSANDRA-11310 has been committed.

PR #923, opened 2016-04-12
CASSANDRA-11549: extended null indicator tests 

stef1927:
I've extended the null indicator tests to catch COPY FROM problems with type conversions.

The enhanced tests will fail until CASSANDRA-11549 is committed.


ptnapoleon:
+1

stef1927:
Ticket committed, merging.

PR #925, opened 2016-04-12
Remove paging tests for queries containing DISTINCT and WHERE clause …

ifesdjeen:
…with keys other than Partition Key, since they're disallowed starting by [CASSANDRA-11339](https://issues.apache.org/jira/browse/CASSANDRA-11339)

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -989,18 +989,6 @@ class TestPagingData(BasePagingTester, PageAssertionMixin):
             self.assertEqual(res, [[1, 2, 1, 2, 3],
                                    [4, 2, 4, 2, 6]])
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Should we still test these on 2.1?

ifesdjeen:
No, they're marked for 3.0 only as they're filtering & paging-related. 
cc @blerer I hope I didn't misunderstand anything.

ptnapoleon:
Ahh I didn't realize these tests were already only running on 3.0+. 

+1

ifesdjeen:
After another discussion, closing as `DISTINCT` should work when restrictions are only on partition key and static columns.

PR #927, opened 2016-04-14
Fix CASSANDRA-11560 (user_types_test failure)

thobbs:
The ``listify()`` function had some strange behavior around single item
lists that resulted in an error.  I made it consistent and
non-surprising here, and updated other tests to match.


mshuler:
Thanks! LGTM

PR #928, opened 2016-04-15
Tests for CASSANDRA-11474

stef1927:
Some tests for CASSANDRA-11474, which disables COPY FROM batching if MAXBATCHSIZE is set to 1, but only for C* versions from 2.2 to 3.5, see discussion on the ticket for details. Since `test_reading_counter_without_batching` is only testing a different code path but no new functionality, I've disabled it for 3.6 onwards. 

As usual, we need to commit the C* ticket before merging this PR.


ptnapoleon:
+1

stef1927:
Ticket committed, merging.

PR #929, opened 2016-04-15
Fix 11563. Don't explicitly set vnodes in test

ptnapoleon:
We should let dtest handle when to run this with and without vnodes. It was originally explicitly set to True always, as that is a more rigorous test case than without vnodes. There is no chance of us ever stopping testing on vnodes by default, so we shouldn't override this in the test.

@knifewine to review


knifewine:
+1

PR #930, opened 2016-04-18
Fix 11564. Only run 2i vnode test with vnodes

ptnapoleon:
@mambocab to review


mambocab:
I'm rerunning Travis, we got a read error trying to pull from `riptano`? Are we hitting some kind of request limit on GH from CassCI?

+1 based on visual inspection and the assumption that you've run it locally with vnodes on and off..

PR #931, opened 2016-04-18
Fix 11476. Watch node logs instead of grepping.

ptnapoleon:
Solves problem where query is aborted but error hasn't been flushed
to log yet

### DIFF
--- a/pushed_notifications_test.py
+++ b/pushed_notifications_test.py
@@ -4,6 +4,7 @@ from threading import Event
 from cassandra import ConsistencyLevel as CL
 from cassandra import ReadFailure
 from cassandra.query import SimpleStatement
+from ccmlib.node import TimeoutError
 from nose.tools import timed
 
 from assertions import assert_invalid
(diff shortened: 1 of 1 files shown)


pushed_notifications_test.py
```diff
@@ -313,9 +313,9 @@ def read_failure_query():
 
         read_failure_query()
 
-        failure = (node1.grep_log(failure_msg) or
-                   node2.grep_log(failure_msg) or
-                   node3.grep_log(failure_msg))
+        failure = (node1.watch_log_for(failure_msg, timeout=5) or
```
mambocab: Does this work when we raise an exception? https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L425

ptnapoleon: Nope. But, if you look down ten lines, it wouldn't work there either. I'll grep after watching

mambocab: I don't follow.

mambocab:
Nice. +1 based on visual inspection; want me to run it?

ptnapoleon:
If you'd like. It passes locally.

mambocab:
Naw. +1.

PR #933, opened 2016-04-18
Properly set CL kwarg in cqlsh copy tests

ptnapoleon:
By using it as an arg, we were placing it were the retry policy
should be. Causing 11565 and 11494, amongst others.

@mambocab to review


ptnapoleon:
I didn't remove the known_failures in this, since it is several. I will do those after merge.

mambocab:
aaaahhhhhhhhhhhhhh

mambocab:
+1

PR #935, opened 2016-04-19
Make improvements to resumable_bootstrap_test

ptnapoleon:
@mambocab to review

Some light cleanup. I was going to make the stress larger, but I no longer think the size of the bootstrap is why the test fails.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -132,8 +132,9 @@ class TestBootstrap(Tester):
         new_rows = list(session.execute("SELECT * FROM %s" % (stress_table,)))
         self.assertEquals(original_rows, new_rows)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11414')
+    @known_failure(failure_source='cassandra',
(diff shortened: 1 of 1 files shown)


bootstrap_test.py
```diff
@@ -144,7 +145,7 @@ def resumable_bootstrap_test(self):
         cluster.populate(2).start(wait_other_notice=True)
 
         node1 = cluster.nodes['node1']
-        node1.stress(['write', 'n=100K', 'cl=TWO', '-schema', 'replication(factor=2)'])
+        node1.stress(['write', 'n=100K', 'cl=TWO', '-schema', 'replication(factor=2)', '-rate', 'threads=50'])
```
mambocab: What's this change for? I think it's probably good, but I don't know the purpose.

ptnapoleon: To limit the number of threads it uses, so as not to time out, or also melt my machine. And maybe it also pins it to running once? I can't confirm.

mambocab: Yeah, pretty sure it pins it to a single run. And sounds good.

mambocab:
+1

PR #939, opened 2016-04-20
CASSANDRA-11497: Fixed SSTableUtilTest.abortedcompaction_test

stef1927:
See discussion on CASSANDRA-11497.


ptnapoleon:
I'm +1, but I'm going to run this a few hundred times before merging.

ptnapoleon:
Looks like this fixes it. +1

PR #942, opened 2016-04-20
Fix 11608

ptnapoleon:
The problem is that when we don't gently stop the node, and we immediately start it right back up, the other nodes sometimes don't even notice it was down in the first place! This means the fact that we were waiting for other notice was bad, because that doesn't happen in that case.

@mambocab  or @knifewine to review. I ran this 500 times and it fixes the issue.

### DIFF
--- a/replace_address_test.py
+++ b/replace_address_test.py
@@ -5,7 +5,7 @@ from cassandra.query import SimpleStatement
 from ccmlib.node import Node, NodeError
 
 from dtest import DISABLE_VNODES, Tester, debug
-from tools import InterruptBootstrap, known_failure, since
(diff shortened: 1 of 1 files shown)


ptnapoleon:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/78/

mambocab:
Do you have a parameterized job that repro's on unpatched dtest?

ptnapoleon:
Of course, it's in the ticket. You want it here too?

mambocab:
No, I'm a dummy who doesn't read tickets is all

mambocab:
This looks great. +1.

PR #943, opened 2016-04-20
protect against active log scanning race conditions

knifewine:
don't merge just yet, still have some testing to do.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -434,12 +434,16 @@ class Tester(TestCase):
             for error in errors:
                 message += "\n{nodename}: {error}".format(nodename=nodename, error=error)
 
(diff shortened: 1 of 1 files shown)


knifewine:
@mambocab to review

I was able to do some basic testing with log scanning enabled/disabled, on tests which have 1 log failure seen, and multiple failures seen. I wasn't able to reproduce the problem of unclean exit with or without my code, so I can't be certain if it helps, but doesn't seem to do any harm.

The bad news is I found another pre-existing flaw with active log scanning, but that's something to tackle later.

mambocab:
Seems reasonable to me. I'm +1, squashing before merge would be good but not required. :+1:

knifewine:
squashed and merging, thanks!

PR #944, opened 2016-04-21
Adjust tests for the new json format that includes timezone (CASSANDRA-11137)

stef1927:
CASSANDRA-11137 added the timezone to JSON timestamps. If you are happy with the changes, please merge this PR since the ticket has already been committed.


stef1927:
I am not sure if we can fix the travis complaints, we just added a space or character to lines that were otherwise already longer that the max allowed. Also, I believe these lines must be formatted as the expected output but let me know if I misunderstood.

ifesdjeen:
`json_test.py` was introduced on 2015-01-30, and `.travis.yml` changes for `200` line length on `2015-06-26`, so looks like it went "under radar". 

Although these tests indeed seem to be "as intended" as they're comparing string output from cqlsh.

ptnapoleon:
+1

ifesdjeen:
thank 2ou @ptnapoleon

PR #945, opened 2016-04-21
Test for tables with thousands of columns for CASSANDRA-11621

ifesdjeen:
(currently [#11621](https://issues.apache.org/jira/browse/CASSANDRA-11621) awaits review)

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -553,6 +553,32 @@ class MiscellaneousCQLTester(CQLTester):
         res = list(session.execute("SELECT * FROM test"))
         self.assertEqual(len(res), 2, msg=res)
 
+    def many_columns_test(self):
+        """
+        Test for tables with thousands of columns.
+        For CASSANDRA-11621.
+        """
+
+        session = self.prepare()
+        width = 5000
+        cluster = self.cluster
+
+        session.execute("CREATE TABLE very_wide_table (pk int PRIMARY KEY, " +
+                        ",".join(map(lambda i: "c_{} int".format(i), range(width))) +
+                        ")")
+
+        session.execute("INSERT INTO very_wide_table (pk, " +
+                        ",".join(map(lambda i: "c_{}".format(i), range(width))) +
+                        ") VALUES (100," +
+                        ",".join(map(lambda i: str(i), range(width))) +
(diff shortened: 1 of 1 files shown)


ifesdjeen:
#11621 was merged as [3244774](https://github.com/apache/cassandra/commit/3244774572c56400ed96da4d57912779878c16e5)

cql_tests.py
```diff
@@ -553,6 +553,37 @@ def range_slice_test(self):
...
+        """
+
+        session = self.prepare()
+        width = 5000
+        cluster = self.cluster
+        table_create_statement = "CREATE TABLE very_wide_table (pk int PRIMARY KEY, "
+        table_create_statement += ",".join(map(lambda i: "c_{} int".format(i), range(width)))
+        table_create_statement += ")"
+
+        session.execute(table_create_statement)
```
mambocab: I'd rather see this as 

```python
session.execute(
    "CREATE TABLE very_wide_table (pk int PRIMARY KEY, " +
    ",".join(map(lambda i: "c_{} int".format(i), range(width))) +
    ")"
)
```

It's a bit more readable, and leaves fewer unnecessary intermediate variables. Same goes for the other `execute` statements below.

cql_tests.py
```diff
@@ -553,6 +553,37 @@ def range_slice_test(self):
         res = list(session.execute("SELECT * FROM test"))
         self.assertEqual(len(res), 2, msg=res)
 
+    @since('2.2')
```
ptnapoleon: I know the bug was only found on 2.2+, but I see no real reason to restrict the testing to 2.2+.

mambocab:
Other than my style concern, this looks fine Python-wise. I've run it locally and it passes on HEAD on 2.2 and later, and fails on 2.2.5. If you make the style change I've requested above, I'm +1.

cql_tests.py
ifesdjeen: True, tested it against 2.1, passes. I removed this annotation.

ifesdjeen:
@mambocab fixed the style. Indeed looks better now.

mambocab:
:+1: awesome, thank you!

PR #949, opened 2016-04-25
Tests for CASSANDRA-11631

stef1927:
Some additional tests for CASSANDRA-11631 (missing values without prepared statements). Please do not merge until ticket is committed.


ptnapoleon:
+1

stef1927:
Thanks for the quick review. I've added some more tests with prepared statements disabled. I've also had a problem with microseconds in the timestamp format. We need to round microseconds to milliseconds because server side we can only parse milliseconds.

ptnapoleon:
+1

stef1927:
Ticket committed, merging.

PR #951, opened 2016-04-25
Fix 11539

knifewine:


ptnapoleon:
Did you run to see how much this reduces the failure rate? +1 to the attempted fix though, but I don't want to remove the annotation until we're sure this nails it down.

knifewine:
yup, looks to be at zero or close to that with a larger dataset. mergin'

PR #955, opened 2016-04-27
Fix test failures caused by differences in output between 3.0 and 3.6…

ifesdjeen:
… after CASSANDRA-11137

[Issue link](https://issues.apache.org/jira/browse/CASSANDRA-11650)

To give a bit more background: main issue is the length of the strings containing `----+-----`. Ideally, we could just strip them from the output, although I found no simple way to do so. It's definitely possible of course although would require more effort. 

Simplest way is to track variable length columns and fill them with ellipsis in order to match additional "-" in output.


stef1927:
Please do not merge this PR, we are probably going to backport CASSANDRA-11137 to 2.2 and 3.0.

stef1927:
CASSANDRA-11137 has been backported to 2.2, this PR should no longer be available, closing.

PR #960, opened 2016-04-28
allow ambiguous tests

knifewine:
this is for special case tests that are difficult to repair, but generally harmless when they fail, provided they can be verified to pass the majority of the time.

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -20,6 +20,40 @@ from nose.plugins.attrib import attr
 from dtest import CASSANDRA_DIR, DISABLE_VNODES, IGNORE_REQUIRE, debug
 
 
+class RerunTestException(Exception):
+    """
+    This exception can be raised to signal a likely harmless test problem. If fixing a test is reasonable, that should be preferred.
+
+    Ideally this is used in conjunction with the 'flaky' decorator, allowing the test to be automatically re-run and passed.
+    When raising this exception in methods decorated with @flaky(rerun_filter=requires_rerun), do so carefully.
+    Avoid overly broad try/except blocks, otherwise real (intermittent) bugs could be masked.
+
+    example usage:
+
+    @flaky(rerun_filter=requires_rerun)  # see requires_rerun method below in this module
+    def some_flaky_test():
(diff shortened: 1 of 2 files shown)


knifewine:
@ptnapoleon and @mambocab for comments/suggestions/insults

knifewine:
Here's would look like on failure: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/86/testReport/node_0.upgrade_supercolumns_test/TestSCUpgrade/upgrade_with_counters_test_2/

ptnapoleon:
I..... don't like it. This doesn't solve the problem of being able to ignore this from the view of the cstar report page. You would need to click through every time. I would prefer to merge this with the flaky annotation, so that if a test failed with this exception, it is re-run automatically, and the better result is taken.

knifewine:
ooh, good idea

knifewine:
ok, I just pushed an update that will allow us to handle narrow failure cases as ambiguous with re-run. if you guys like that approach I'll do another bulk run so we can try it out in CI land.

ptnapoleon:
This looks perfect

knifewine:
ok, trying it out on the multiplexer. looks like if flaky is not enabled it will just bubble up as an exception, which should be easy to spot.
trying it now with flaky enabled here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/91/

knifewine:
All looks to work well, but the test improbably failed twice in a row, so there's a single failure. I'll bump it up to allow 2 reruns.

knifewine:
@ptnapoleon I made some docstring updates and squashed, can you have another look?

ptnapoleon:
I am +1

mambocab:
This looks excellent. I think I prefer the `RerunTestException` name you originally proposed in HC. So, consider a rename? but I'm +1 either way.

knifewine:
ok @mambocab rename done, I'll merge if that looks good to you.

mambocab:
Nice, +1

PR #966, opened 2016-05-02
CASSANDRA-11692: do not check final files after aborted compaction

stef1927:
In `SSTableUtilTest.abortedcompaction_test()`, we should not check final files after aborting compaction since we do not know how far it got. 

See the [C* ticket](https://issues.apache.org/jira/browse/CASSANDRA-11692) for more details.


ptnapoleon:
I'm running here: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/94/

PR #967, opened 2016-05-03
Ignore tmplink files

krummas:
CASSANDRA-11673


mambocab:
@krummas Where do these `tmplink` files come from? Why do we ignore them here? I'm wondering if other operations over SSTables need to do the same.

krummas:
tmplink files are created when doing early opening of sstables, so if we are counting files, yes we probably need to ignore them

PR #968, opened 2016-05-03
Fix test_paging_for_range_name_queries for 2.1

blerer:


blerer:
@ptnapoleon could you review? In case I missed something.

ptnapoleon:
Sorry for the delay, was out sick. I don't see a ticket prompting the change, but the code looks fine, and this is the only query of that type in this test. So +1, under the assumption this is a 2.2+ only feature being tested.

blerer:
Yes, I initially only tested my test on 2.2+ and forgot that IN restrictions were supported only on the last restricted column in 2.1.
Thanks for the review.

PR #969, opened 2016-05-05
Add test for Cassandra-11031 allow filtering on partition key

jasonstack:


cql_tests.py
```diff
@@ -298,6 +298,105 @@ def statements_test(self):
...
+        session.execute("INSERT INTO test_filter (k1, k2, ck1, v) VALUES (1, 1, 3, 0)")
+
+        # select test
+        res = session.execute("SELECT * FROM test_filter WHERE k1 = 0 ALLOW FILTERING")
+        res_list = rows_to_list(res)
+        res_list.sort()
+        expected = [[0, 0, 0, 0],[0, 0, 1, 0],[0, 0, 2, 0],[0, 0, 3, 0],
+            [0, 1, 0, 0],[0, 1, 1, 0],[0, 1, 2, 0],[0, 1, 3, 0]]
+        expected.sort()
+        self.assertEqual(res_list, expected)
```
mambocab: I know this is a large reformatting to ask for, but this would be better formatted as

```python
self.assertEqual(
    sorted(rows_to_list(session.execute("SELECT * FROM test_filter WHERE k1 = 0 ALLOW FILTERING"))),
    [[0, 0, 0, 0],
     [0, 0, 1, 0],
     [0, 0, 2, 0],
     [0, 0, 3, 0],
     [0, 1, 0, 0],
     [0, 1, 1, 0],
     [0, 1, 2, 0],
     [0, 1, 3, 0]]
)
```

I like to avoid assigning intermediate values to reused names; I think it invites mistakes. In addition, this gets rid of the missing whitespace after the commas in the lists here.

cql_tests.py
```diff
@@ -298,6 +298,105 @@ def statements_test(self):
...
+        res_list = rows_to_list(res)
+        res_list.sort()
+        expected = [[1, 0, 0, 0],[1, 0, 1, 0],[1, 0, 2, 0],[1, 0, 3, 0],
+            [1, 1, 0, 0],[1, 1, 1, 0],[1, 1, 2, 0],[1, 1, 3, 0]]
+        expected.sort()
+        self.assertEqual(res_list, expected)
+
+        res = session.execute("SELECT * FROM test_filter WHERE k1 = 2 ALLOW FILTERING")
+        res_list = rows_to_list(res)
+        self.assertEqual(len(res_list), 0)
```
mambocab: Could you please rewrite this (and similar blocks below) as

```python
self.assertEqual(
    rows_to_list(session.execute("SELECT * FROM test_filter WHERE k1 = 2 ALLOW FILTERING")),
    []
)
```

Comparing to `[]` gives us a more verbose and more helpful error. For example, this file:

```python
from unittest import TestCase

class TestAssertVerbosityDemo(TestCase):

    def test_check_length(self):
        self.assertEqual(len([1, 2, 3]), 0)

    def test_check_equality(self):
        self.assertEqual([1, 2, 3], [])
```

gives these errors:

```
FF
FAIL: test.py:TestAssertVerbosityDemo.test_check_equality
  File "/home/mambocab/tmp/scratch-1462456936/test.py", line 9, in test_check_equality
    self.assertEqual([1, 2, 3], [])
AssertionError: Lists differ: [1, 2, 3] != []

First list contains 3 additional elements.
First extra element 0:
1

- [1, 2, 3]
+ []

FAIL: test.py:TestAssertVerbosityDemo.test_check_length
  File "/home/mambocab/tmp/scratch-1462456936/test.py", line 6, in test_check_length
    self.assertEqual(len([1, 2, 3]), 0)
AssertionError: 3 != 0

Ran 2 tests in 0.001s

FAILED (failures=2)
```

cql_tests.py
```diff
@@ -298,6 +298,105 @@ def statements_test(self):
...
+        res = session.execute("SELECT count(*) FROM test_filter WHERE k2 = 1 ALLOW FILTERING")
+        self.assertEqual(rows_to_list(res), [[8]])
+
+        res = session.execute("SELECT count(*) FROM test_filter WHERE k2 = 2 ALLOW FILTERING")
+        self.assertEqual(rows_to_list(res), [[0]])
+
+        try:
+            session.execute("SELECT * FROM test_filter WHERE k1 = 0")
+            self.fail("Partition key parts: k2 must be restricted as other parts are")
+        except Exception as e:
```
mambocab: I want to say this will always pass -- I think `Tester.fail` raises some subclass of `Exception`. More generally, catching an undifferentiated `Exception` will allow errors to pass that we should fail on. This code should use `self.assertRaises` like [this code does](https://github.com/riptano/cassandra-dtest/blob/83615470198ab693d1547a88b10cff13ea175d39/commitlog_test.py#L271), and check for a more specific exception.

mambocab:
Thanks for the tests! I'd like to have a second look after my comments are addressed, but they look good so far other than my style comments.

jasonstack:
Thanks for your advice. I updated the tests.

jasonstack:
I fixed the indentation according to the output of `flake8 --ignore=E501,F811,F812,F821,F822,F823,F831,F841,N8,C9 --exclude=thrift_bindings,cassandra-thrift .`

ptnapoleon:
+1. Feel free to merge when the C* ticket is committed.

ifesdjeen:
should we squash commits here?

ptnapoleon:
Yes, absolutely. Thank you

ifesdjeen:
Just to give heads up: CASSANDRA ticket was not merged. Sorry I was not specific enough...

PR #971, opened 2016-05-09
new upgrade version logic

knifewine:
(mostly) unites test class generation between
the upgrade_through_version tests and the other upgrade
tests of mixed version clusters.

introduces clearer metaphors and makes test coverage
more explicit with the concept of indev, current, and next
for each version.

generative approach to tests using a manifest or starting
versions to all possible upgraded-to versions

### DIFF
--- a/upgrade_tests/upgrade_manifest.py
+++ b/upgrade_tests/upgrade_manifest.py
@@ -1,0 +1,116 @@
+from collections import namedtuple
+
+from dtest import debug
+
+# UpgradePath's contain data about upgrade paths we wish to test
+# They also contain VersionMeta's for each version the path is testing
+UpgradePath = namedtuple('UpgradePath', ('name', 'starting_version', 'upgrade_version', 'starting_meta', 'upgrade_meta'))
+
+# VersionMeta's capture data about version lines, protocols supported, and current version identifiers
+# they should have a 'variant' value of 'current', 'indev', or 'next':
+#    'current' means most recent released version, 'indev' means where changing code is found, 'next' means a tentative tag
+VersionMeta = namedtuple('VersionMeta', ('name', 'variant', 'version', 'min_proto_v', 'max_proto_v'))
+
+indev_2_0_x = None  # None if release not likely
+current_2_0_x = VersionMeta(name='current_2_0_x', variant='current', version='2.0.17', min_proto_v=1, max_proto_v=2)
+next_2_0_x = None  # None if not yet tagged
+
+indev_2_1_x = VersionMeta(name='indev_2_1_x', variant='indev', version='git:cassandra-2.1', min_proto_v=1, max_proto_v=3)
+current_2_1_x = VersionMeta(name='current_2_1_x', variant='current', version='2.1.14', min_proto_v=1, max_proto_v=3)
+next_2_1_x = None  # None if not yet tagged
+
+indev_2_2_x = VersionMeta(name='indev_2_2_x', variant='indev', version='git:cassandra-2.2', min_proto_v=1, max_proto_v=4)
+current_2_2_x = VersionMeta(name='current_2_2_x', variant='current', version='2.2.6', min_proto_v=1, max_proto_v=4)
+next_2_2_x = None  # None if not yet tagged
+
+indev_3_0_x = VersionMeta(name='indev_3_0_x', variant='indev', version='git:cassandra-3.0', min_proto_v=3, max_proto_v=4)
+current_3_0_x = VersionMeta(name='current_3_0_x', variant='current', version='3.0.5', min_proto_v=3, max_proto_v=4)
+next_3_0_x = None  # None if not yet tagged
+
+indev_3_x = VersionMeta(name='indev_3_x', variant='indev', version='git:cassandra-3.7', min_proto_v=3, max_proto_v=4)
+current_3_x = VersionMeta(name='current_3_x', variant='current', version='3.5', min_proto_v=3, max_proto_v=4)
+next_3_x = None  # None if not yet tagged
+
+head_trunk = VersionMeta(name='head_trunk', variant='indev', version='git:trunk', min_proto_v=3, max_proto_v=4)
+
+
+# maps an VersionMeta representing a line/variant to a list of other VersionMeta's representing supported upgrades
+MANIFEST = {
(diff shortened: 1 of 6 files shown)


knifewine:
@mambocab to review

mambocab:
Have you run this somewhere on CassCI? Looks good after a quick skim but I'd like to a) read more b) see the results of a test run.

upgrade_tests/upgrade_manifest.py
```diff
@@ -0,0 +1,116 @@
...
+            if not _is_targeted_variant_combo(origin_meta, destination_meta):
+                debug("skipping class creation, no testing of '{}' to '{}' (for {} upgrade to {})".format(origin_meta.variant, destination_meta.variant, origin_meta.name, destination_meta.name))
+                continue
+
+            if not _have_common_proto(origin_meta, destination_meta):
+                debug("skipping class creation, no compatible protocol version between {} and {}".format(origin_meta.name, destination_meta.name))
+                continue
+
+            valid_upgrade_pairs.append(
+                    UpgradePath(
```
mambocab: this is overindented.

mambocab:
I don't think I've ever understood why the protocol version is specified up front. Why doesn't the driver just negotiate through an upgrade as appropriate?

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -787,259 +786,75 @@ def create_upgrade_class(clsname, version_list, protocol_version,
...
+                     ('partitioner', 'org.apache.cassandra.dht.RandomPartitioner'),
+                 )),
+
+    # Proto v4 upgrades (v4 is supported on 2.2, 3.0, 3.1, trunk)
+    MultiUpgrade(name='ProtoV4Upgrade_AllVersions_EndsAt_Trunk_HEAD',
+                 version_metas=[current_2_2_x, current_3_0_x, head_trunk], protocol_version=4, extra_config=None),
+    MultiUpgrade(name='ProtoV4Upgrade_AllVersions_RandomPartitioner_EndsAt_Trunk_HEAD',
+                 version_metas=[current_2_2_x, current_3_0_x, head_trunk], protocol_version=4,
+                 extra_config=(
+                    ('partitioner', 'org.apache.cassandra.dht.RandomPartitioner'),
```
mambocab: I think this needs another space of indentation.

upgrade_tests/upgrade_manifest.py
```diff
@@ -0,0 +1,116 @@
...
+next_3_0_x = None  # None if not yet tagged
+
+indev_3_x = VersionMeta(name='indev_3_x', variant='indev', version='git:cassandra-3.7', min_proto_v=3, max_proto_v=4)
+current_3_x = VersionMeta(name='current_3_x', variant='current', version='3.5', min_proto_v=3, max_proto_v=4)
+next_3_x = None  # None if not yet tagged
+
+head_trunk = VersionMeta(name='head_trunk', variant='indev', version='git:trunk', min_proto_v=3, max_proto_v=4)
+
+
+# maps an VersionMeta representing a line/variant to a list of other VersionMeta's representing supported upgrades
```
mambocab: Took me a minute to remember what this was for/decipher this comment. I think this should expand on what "supported upgrades" means.

Has a dev reviewed this? Not sure who the right person for the job is; maybe Sylvain? Alternatively, is there documentation for supported upgrades?

upgrade_tests/upgrade_manifest.py
```diff
@@ -0,0 +1,116 @@
...
+    Takes two VersionMeta objects, in order of test from start version to next version.
+    Returns a boolean indicating if this is a test pair we care about.
+
+    for now we only test upgrades of these types:
+      current -> in-dev (aka: released -> branch)
+      current -> next (aka: released -> proposed release point)
+      next -> in-dev (aka: proposed release point -> branch)
+    """
+    return (origin_meta.variant == 'current' and destination_meta.variant == 'indev')\
+        or (origin_meta.variant == 'current' and destination_meta.variant == 'next')\
```
mambocab: Style nit: rather than using line continuations, just enclose this whole disjunction in parens.

ptnapoleon:
Because the driver doesn't negotiate in a mixed version cluster. It chooses the highest protocol version that can be used by the node it opens its first control connection to. If that isn't the lower of the two versions in the cluster, then you will only be able to connect to the upgraded nodes. Thus, we need to be explicit, which is SOP for users in the field in an upgrade.

knifewine:
I've tried this out for 3.0.6 and 3.6 tentative release testing, seemed to go ok (with some flakes, but I don't think anything related to this code). I've got a full run going now over here: http://cassci.datastax.com/view/Parameterized/job/upgrade_tests-all-custom_branch_runs/13/

upgrade_tests/upgrade_manifest.py
knifewine: The basic gist, as I understand it, is that you can jump from any version to a later version. This can be as wide a jump as 2.1 direct to trunk, or smaller jumps spanning the versions between. I'll update the comment accordingly.

knifewine: comments updated with more about upgrades allowed, and also with what "support" means since that's a pretty vague word.

upgrade_tests/upgrade_manifest.py
knifewine: nit approved and fixed.

upgrade_tests/upgrade_manifest.py
knifewine: fixed.

knifewine:
Yeah, ideally we could do away with the protocol version specification because it's awkward. But as I understand it we'd likely run into trouble without being specific there. Though, it's not something I've tried, so we can attempt it (would just rather hold off a change like that for another PR).

mambocab:
Specifying the protocol version seems fine, given what Philip said. I'm +1 to merge now, but I'd like it if a dev reviewed the changes and the upgrade paths chosen here.

knifewine:
ok, over chat I got basic validation of the version matrix (without subjecting anyone else to parsing the python version of that matrix).

knifewine:
a quick glance at travis makes me think it's not working properly for files within the upgrade_tests/ subdir.

PR #973, opened 2016-05-11
remove asserts that test vnode allocation distribution

krummas:


krummas:
this was CASSANDRA-11279 btw

PR #974, opened 2016-05-12
instructions and template-ish code for custom upgrade path testing

knifewine:
this is to make local testing easier for devs in their envs.
this was the simplest way I could think to help make local testing both flexible and not too awful.

### DIFF
--- a/upgrade_tests/upgrade_manifest.py
+++ b/upgrade_tests/upgrade_manifest.py
@@ -65,14 +65,14 @@ MANIFEST = {
 
 # Local env and custom path testing instructions. Use these steps to REPLACE the normal upgrade test cases with your own.
 # 1) Add a VersionMeta for each version you wish to test (see examples below). Update the name, version, and protocol restrictions as needed. Use a unique name for each VersionMeta.
-# 2) Update override_manifest (see example below).
+# 2) Update OVERRIDE_MANIFEST (see example below).
 # 3) If using ccm local: slugs, make sure you have LOCAL_GIT_REPO defined in your env. This is the path to your git repo.
 # 4) Run the tests! To run all, use 'nosetests -v upgrade_tests/'. To run specific tests, use 'nosetests -vs --collect-only' to preview the test names, then run nosetests using the desired test name.
 custom_1 = VersionMeta(name='custom_branch_1', variant='indev', version='local:some_branch', min_proto_v=3, max_proto_v=4)
 custom_2 = VersionMeta(name='custom_branch_2', variant='indev', version='git:trunk', min_proto_v=3, max_proto_v=4)
 custom_3 = VersionMeta(name='custom_branch_3', variant='indev', version='git:cassandra-3.5', min_proto_v=3, max_proto_v=4)
 custom_4 = VersionMeta(name='custom_branch_4', variant='indev', version='git:cassandra-3.6', min_proto_v=3, max_proto_v=4)
-override_manifest = {
+OVERRIDE_MANIFEST = {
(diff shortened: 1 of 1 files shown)


knifewine:
@mambocab to review
@thobbs I thought you might have some feedback here too, since you had to jump through some of these hoops earlier today.

upgrade_tests/upgrade_manifest.py
```diff
@@ -64,6 +63,21 @@
...
+# Local env and custom path testing instructions. Use these steps to REPLACE the normal upgrade test cases with your own.
+# 1) Add a VersionMeta for each version you wish to test (see examples below). Update the name, version, and protocol restrictions as needed. Use a unique name for each VersionMeta.
+# 2) Update override_manifest (see example below).
+# 3) If using ccm local: slugs, make sure you have LOCAL_GIT_REPO defined in your env. This is the path to your git repo.
+# 4) Run the tests! To run all, use 'nosetests -v upgrade_tests/'. To run specific tests, use 'nosetests -vs --collect-only' to preview the test names, then run nosetests using the desired test name.
+custom_1 = VersionMeta(name='custom_branch_1', variant='indev', version='local:some_branch', min_proto_v=3, max_proto_v=4)
+custom_2 = VersionMeta(name='custom_branch_2', variant='indev', version='git:trunk', min_proto_v=3, max_proto_v=4)
+custom_3 = VersionMeta(name='custom_branch_3', variant='indev', version='git:cassandra-3.5', min_proto_v=3, max_proto_v=4)
+custom_4 = VersionMeta(name='custom_branch_4', variant='indev', version='git:cassandra-3.6', min_proto_v=3, max_proto_v=4)
+override_manifest = {
```
mambocab: Will this ever change at runtime? I don't think it does. If not, its name should be all-caps so it's clearly a constant. I'm being picky about this because, well, it's me, but also, `_is_targeted_variant_combo` closes over it, I want it to be extra clear that the value defined here will be used in the closure at function-definition time.

upgrade_tests/upgrade_manifest.py
```diff
@@ -83,7 +97,8 @@ def _is_targeted_variant_combo(origin_meta, destination_meta):
       current -> next (aka: released -> proposed release point)
       next -> in-dev (aka: proposed release point -> branch)
     """
-    return ((origin_meta.variant == 'current' and destination_meta.variant == 'indev') or
+    return (override_manifest or  # if we're overriding the test manifest, we don't want to filter anything out
```
mambocab: This is the closure I'm talking about [in this comment](https://github.com/riptano/cassandra-dtest/pull/974/files#r63204462).

Can we change this to `bool(override_manifest)`? I know it doesn't really matter Because Ducks, but I think I'd prefer it return the same type all the time unless there's a compelling reason.

upgrade_tests/upgrade_manifest.py
knifewine: yup, will do.

upgrade_tests/upgrade_manifest.py
knifewine: yeah, that's clearer, will do.

knifewine:
tested locally using some local: slugs to make sure everything works if you follow the instructions.

upgrade_tests/upgrade_manifest.py
knifewine: done

mambocab:
+1, looks excellent

thobbs:
This is definitely an improvement, thanks for merging this.

PR #977, opened 2016-05-16
upgrades: add java version awareness

knifewine:
allows us to re-enable 2.0->2.1 upgrade testing

### DIFF
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -12,10 +12,19 @@ UPGRADE_TEST_RUN = os.environ.get('UPGRADE_TEST_RUN', '').lower() in {'true', 'y
 
 
 def switch_jdks(major_version_int):
+    """
+    Changes the jdk version globally, by setting JAVA_HOME = JAVA[N]_HOME.
+    This means the environment must have JAVA[N]_HOME set to switch to jdk version N.
+    """
     try:
-        os.environ['JAVA_HOME'] = os.environ['JAVA{}_HOME'.format(major_version_int)]
+        os.environ['JAVA{}_HOME'.format(major_version_int)]
     except KeyError:
         raise RuntimeError("You need to set JAVA{}_HOME to run these tests!".format(major_version_int))
+
+    # don't change if the same version was requested
+    if os.environ.get('JAVA_HOME') != os.environ['JAVA{}_HOME'.format(major_version_int)]:
+        debug("Switching jdk to {}".format(major_version_int))
+        os.environ['JAVA_HOME'] = os.environ['JAVA{}_HOME'.format(major_version_int)]
 
 
 @skipIf(not UPGRADE_TEST_RUN, 'set UPGRADE_TEST_RUN=true to run upgrade tests')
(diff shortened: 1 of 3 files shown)


knifewine:
@mambocab to review

mambocab:
Code looks good. I'll run a smoke test on my machine, but have you run this on CassCI somewhere?

mambocab:
Smoke test:

```sh
$ JAVA8_HOME=/usr/lib/jvm/java-8-oracle JAVA7_HOME=/usr/lib/jvm/java-7-oracle CASSANDRA_VERSION=git:trunk NOSE_PASTEABLE=true UPGRADE_TEST_RUN=true nosetests -vv upgrade_tests/upgrade_through_versions_test.py:TestUpgrade_current_2_0_x_To_indev_2_1_x.rolling_upgrade_with_internode_ssl_test
```

passes on my machine.

mambocab:
I'm +1 on the basis of code inspection and my smoke test. I'd be even more +1 with a successful CassCI build.

knifewine:
pushed one more new commit on top of this, to make sure we handle jdk switching properly in both u_t_v, and also in the mixed-version tests.

@mambocab I'd love another pass with those new changes.

In the meantime I'm running a small CI job here to try things out in a non-local env: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/98/

upgrade_tests/upgrade_base.py
```diff
@@ -12,11 +12,20 @@
...
+    This means the environment must have JAVA[N]_HOME set to switch to jdk version N.
+    """
     try:
-        os.environ['JAVA_HOME'] = os.environ['JAVA{}_HOME'.format(major_version_int)]
+        os.environ['JAVA{}_HOME'.format(major_version_int)]
     except KeyError:
         raise RuntimeError("You need to set JAVA{}_HOME to run these tests!".format(major_version_int))
 
+    # don't change if the same version was requested
+    if os.environ.get('JAVA_HOME') != os.environ['JAVA{}_HOME'.format(major_version_int)]:
```
mambocab: This logging could be a little more verbose -- maybe log the current `JAVA_HOME`, then log whether or not you changed that.

upgrade_tests/upgrade_base.py
```diff
@@ -32,6 +41,13 @@ class UpgradeTester(Tester):
     __metaclass__ = ABCMeta
     NODES, RF, __test__, CL, UPGRADE_PATH = 2, 1, False, None, None
 
+    def setUp(self):
+        debug("Upgrade test beginning, setting CASSANDRA_VERSION to {}, and jdk to {}. (Prior values will be restored after test)."
+              .format(self.UPGRADE_PATH.starting_version, self.UPGRADE_PATH.starting_meta.java_version))
+        switch_jdks(self.UPGRADE_PATH.starting_meta.java_version)
+        os.environ['CASSANDRA_VERSION'] = self.UPGRADE_PATH.starting_version
```
mambocab: I don't like this, but I acknowledge it might be necessary given the way dtest is currently configured. Why'd you chose to do it this way?

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -242,10 +242,13 @@ def __init__(self, *args, **kwargs):
         Tester.__init__(self, *args, **kwargs)
 
     def setUp(self):
+        debug("Upgrade test beginning, setting CASSANDRA_VERSION to {}, and jdk to {}. (Prior values will be restored after test)."
+              .format(self.test_version_metas[0].version, self.test_version_metas[0].java_version))
+        os.environ['CASSANDRA_VERSION'] = self.test_version_metas[0].version
```
mambocab: Same thought as above.

knifewine:
I also ran about 1k tests locally overnight and the results are looking pretty good.

upgrade_tests/upgrade_base.py
knifewine: this is the surest way I know to get ccm to use that version. not sure what an alternative might be.

knifewine: but yeah, I agree it's kinda gross. dtest does, however record and restore the pre-test env after tests run.

mambocab: Fair enough, leave it for now, then. This might be something to add to global configuration (https://github.com/riptano/cassandra-dtest/blob/master/dtest.py#L73) at some point, though there may be CCM changes necessary too.

mambocab:
How confident are you about the entire suite based on the success of that parameterized run? I'm at about 75%, if you're more confident than that, then merge away IMO.

knifewine:
my local 1k tests run overnight looked pretty good (but those were only mixed-version tests). I'm running some more u_t_v tests locally now and if those look ok, I'm decently confident.

mambocab:
Ok, cool -- when you feel good about it, merge away.

knifewine:
small logging change made as suggested above (sorry, missed it earlier).

PR #979, opened 2016-05-19
Don't wait notice on an already stopped node

ptnapoleon:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/103/

https://issues.apache.org/jira/browse/CASSANDRA-11837

Literally anyone to review, I promise it's fine.

### DIFF
--- a/topology_test.py
+++ b/topology_test.py
@@ -47,7 +47,7 @@ class TestTopology(Tester):
 
         # Decommission node and wipe its data
         node2.decommission()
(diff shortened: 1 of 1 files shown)


mambocab:
Why are we calling `stop` at all?

ptnapoleon:
I'm not sure that the process ends after decommission, only that the node leaves the cluster.

mambocab:
Ah, gotcha. Ok, +1.

PR #980, opened 2016-05-19
change list default arg with None then set in method

knifewine:


ptnapoleon:
+100

mambocab:
Oof, nice catch. +1.

knifewine:
thanks gents

PR #981, opened 2016-05-20
exclude some replication tests so they can run on the large dtest job

knifewine:
@ptnapoleon to review. tested cfgs locally with --collect-only, and looks to work just fine to leave those tests out.


ptnapoleon:
+1

PR #982, opened 2016-05-21
Always disable maxDiff

ptnapoleon:
@knifewine and @mambocab to review


knifewine:
Good idea. If this leads to any excessive console logs we can always toggle back off later.

+1

PR #983, opened 2016-05-23
Fixes for CASSANDRA-11731

stef1927:
Some fixes as discussed in CASANDRA-11731 and additional log messages. 

The restart node tests have been modified to take into account the fix that will be delivered by CASSANDRA-11038. Until this patch is delivered, these tests will fail if you merge this pull request.


ptnapoleon:
+1

PR #978, opened 2016-05-16
Fix auth test import

mambocab:
The first commit here (`don't use env vars for getting C* version`) is intended to fix #976, but does not on my machine. Regardless, I think it's a good thing to do. To demonstrate the problem in #976, run `auth_test.py:TestAuthRoles` with the `CASSANDRA_DIR` env var unset, and the `CASSANDRA_VERSION` var set.

I'm not sure why this doesn't fix the issue. Any thoughts? @ptnapoleon @knifewine

PR #984, opened 2016-05-23
Revert "Fix auth test import"

mambocab:
Reverts riptano/cassandra-dtest#978


mambocab:
I need to fix #978 and re-merge.

PR #986, opened 2016-05-23
Fix11857

c-kodman:
Please review this.  Just fixed so that "frozen" will not run on anything less than 2.1


c-kodman:
@ptnapoleon

ptnapoleon:
+1

PR #993, opened 2016-05-24
adjusted wait time for index creation

smccarthy788:
@ptnapoleon please review.


ptnapoleon:
We don't have a driver connection here, to block on index creation, so +1, once Travis is happy.

PR #994, opened 2016-05-24
make note of env var for local upgrade testing

knifewine:
@mambocab I forgot to mention this env var in my earlier pr (one more small thing I don't want people to trip over)


mambocab:
+1, good catch

PR #996, opened 2016-05-25
3.6 branch is still alive until 3.6 release

knifewine:
@ptnapoleon or @mambocab to review


ptnapoleon:
+1, I think. Assuming that 3.6 is what we want for 3.x indev upgrading

knifewine:
I think so? given we want a provisional look ahead at whatever may be changing prior to a tag.

knifewine:
(or at least based on discussions this morning)

PR #997, opened 2016-05-25
Fix CASSANDRA-11890

jkni:
CASSANDRA-11890 covers recent auth failures in tests on 2.2, 3.0, and 3.x. They appeared recently because the cassandra-test branch of the Python driver was updated for the 3.4.0 driver release.

These failures come from two commits in [datastax/python-driver](https://github.com/datastax/python-driver/commits/cassandra-test).

The failures in user_types_test came from this [commit](https://github.com/datastax/python-driver/commit/88b013daac765cec04626cc3d5fdad804fd0b19b), which adds error checking for positional retry_policy parameters. A consistency level was instead being passed; we now pass this as a labeled argument.

The failures in auth_test came from this [commit](https://github.com/datastax/python-driver/commit/da994dcd3ad76f3eacae367b87f345f2ccc7a1c4), which explicitly states when errors come from the server in error messages. I've updated the regex to match this new error format and added logging of the actual message when the regex doesn't match.

EDIT: Edited to add that I've confirmed failures prior to these fixes locally and ran the tests after these fixes to confirm they pass. I have not ran the fixes on CassCI.

### DIFF
--- a/auth_test.py
+++ b/auth_test.py
@@ -9,7 +9,7 @@ from cassandra.protocol import SyntaxException
 from assertions import (assert_all, assert_invalid, assert_one,
                         assert_unauthorized)
 from dtest import CASSANDRA_VERSION_FROM_BUILD, Tester, debug
-from tools import known_failure, since
+from tools import since
 
 
 class TestAuth(Tester):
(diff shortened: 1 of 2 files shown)


ptnapoleon:
Travis is upset that you didn't remove the imports of `known_failure`.

ptnapoleon:
I checked this works against all C* versions. +1, after linter fix. Tyvm

PR #998, opened 2016-05-25
Test or CASSANDRA-11891

thobbs:


commitlog_test.py
```diff
@@ -160,6 +161,77 @@ def _provoke_commitlog_failure(self):
...
+            if replay_complete:
+                break
+        else:
+            self.fail("Did not finish commitlog replay within 120 seconds")
+
+        debug("Reconnecting to node")
+        session = self.patient_cql_connection(node1)
+        debug("Make query to ensure data is present")
+        res = list(session.execute("SELECT * FROM Test.mytable"))
+        self.assertEqual(num_rows, len(res))
```
ptnapoleon: Can you add `res` as the third arg to this?

commitlog_test.py
```diff
@@ -160,6 +161,77 @@ def _provoke_commitlog_failure(self):
...
+        for i in xrange(num_rows):
+            session.execute("INSERT INTO Test.mytable (a, b, c) VALUES (0, %d, %d)" % (i, i))
+
+        node1.stop(gently=False)
+        node1.mark_log_for_errors()
+
+        debug("Verify commitlog was written before abrupt stop")
+        commitlog_dir = os.path.join(node1.get_path(), 'commitlogs')
+        commitlog_files = os.listdir(commitlog_dir)
+        self.assertTrue(len(commitlog_files) > 0)
```
ptnapoleon: Can you add `commitlog_files` as the second arg to this method?

ptnapoleon:
+1 after minor nits to improve debug output on assertion failure.

commitlog_test.py
mambocab: It'd be better to use `self.assertNotEqual(commitlog_files, [])`.

thobbs:
Looks like the flake8 failure is in an unrelated file, and the previous commits had no problems, so I'm going to merge this.  Thanks for the reviews.

PR #999, opened 2016-05-26
coordinator dies repair test

krummas:


repair_tests/repair_test.py
```diff
@@ -750,6 +750,35 @@ def test_multiple_concurrent_repairs(self):
...
+        cluster.populate(3).start(wait_for_binary_proto=True)
+        node1, node2, node3 = cluster.nodelist()
+        node1.stress(['write', 'n=100k', '-schema', 'replication(factor=3)', '-rate', 'threads=30'])
+        if cluster.version() >= "2.2":
+            t1 = threading.Thread(target=node1.repair)
+        else:
+            t1 = threading.Thread(target=node1.nodetool, args=('repair keyspace1 standard1 -inc -par',))
+        t1.start()
+        node1.watch_log_for('equesting merkle trees for')  # requesting in 2.1, Requesting in 2.2+
+        time.sleep(0.5)  # make sure the other nodes have time to mark as repairing
```
ptnapoleon: Do we want to watch logs here? Or increase this sleep? 0.5s is very brief.

ptnapoleon:
I don't see where we assert the failed repair session is cleared out. Does issue-ing the second repair (and it succeeding) guarantee that?

repair_tests/repair_test.py
krummas: We watch logs until the coordinator sends it out, then we wait 0.5 for the other nodes to receive and mark sstables as repairing.

I guess it makes more sense to watch the logs on the receiving nodes - will fix that

krummas:
the second repair fails unless the repair session is cleared out

ptnapoleon:
+1 then, after log watching fix

ptnapoleon:
@krummas, do you want us to make the log watching fix? This has been hanging for a while

krummas:
updated, needed to do the old check on 2.1 since we don't log as much there

ptnapoleon:
C* Ticket is merged, tests pass, and changes were made, I'm gonna merge this, +1!

PR #1000, opened 2016-05-26
Never purge tombstones longer test

krummas:


compaction_test.py
```diff
@@ -451,6 +451,28 @@ def disable_autocompaction_alter_and_nodetool_test(self):
...
+        """
+        @jira_ticket CASSANDRA-11831
+        Run a longer compaction with never_purge_tombstones to make sure we don't refresh the overlaps
+        """
+        cluster = self.cluster
+        cluster.populate(1).start(jvm_args=["-Dcassandra.never_purge_tombstones=true"], wait_for_binary_proto=True)
+        [node] = cluster.nodelist()
+        node.nodetool("setcompactionthroughput 1")
+        for i in xrange(0, 4):
+            node.stress(['write', 'n=300k', "no-warmup"])
```
ptnapoleon: This will write the same 300k rows each iteration of the loop. Is that okay?

krummas: yes, we just want to trigger a compaction that takes more than 1 minute

ptnapoleon:
+1

PR #1001, opened 2016-05-26
Remove index_summary_upgrade_test.py

thobbs:
See the rationale in CASSANDRA-11127.


ptnapoleon:
+1, based on persuasive rationale in that ticket. Can you please include the ticket number in the commit message though?

thobbs:
Good call.  I've force pushed a new commit message.

PR #1004, opened 2016-05-26
Changed % string formatting to .format() string formatting

smccarthy788:
@mambocab, please review.

Tested on cassci:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/124/


mambocab:
Looks good, and the only failures are known ones. +1, thanks.

PR #1005, opened 2016-05-26
Force single data dir for scrub tests

yukim:
This is for [CASSANDRA-11693](https://issues.apache.org/jira/browse/CASSANDRA-11693).

Since range-aware JBOD is introduced, there is a chance that zero byte SSTable is created (not actually on the disk, just internally) and generation number goes out of sync with what tests expect.

I propose to simply disable JBOD configuration for scrub tests, since it is not the target of the tests anyway.

### DIFF
--- a/scrub_test.py
+++ b/scrub_test.py
@@ -14,6 +14,17 @@ KEYSPACE = 'ks'
 
 
 class TestHelper(Tester):
+
+    def setUp(self):
+        """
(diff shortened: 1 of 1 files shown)


ptnapoleon:
The change looks fine, +1. If you're happy with the behavior of C* here, which doesn't sound too unreasonable.

PR #1008, opened 2016-05-27
8844 tests pr0

mambocab:
Incrementally pushing up tests for 8844. @ptnapoleon and/or @knifewine to review, please. @josh-mckenzie, could you please take a look as well? These pass locally running on your `8844_review` branch.

Please (please) don't merge until I've squashed this into something sensible. Don't even look at the commits. Please, I beg of you.

### DIFF
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -252,12 +252,9 @@ class TestCDC(Tester):
                           s=time.time() - start))
                 error_found = True
 
-        debug(
-            'number of rows inserted: ' +
-            str(len(list(
-                session.execute('SELECT * FROM ' + ks_name + '.' + full_cdc_table_name)
-            )))
-        )
+        self.assertLess(0, rows_loaded,
+                        'No CDC rows inserted. This may happen when '
+                        'cdc_total_space_in_mb > commitlog_segment_size_in_mb')
 
         commitlog_dir = os.path.join(node.get_path(), 'commitlogs')
         commitlogs_size = size_of_files_in_dir(commitlog_dir)
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -1,0 +1,356 @@
+from __future__ import division
+
+import os
+import time
+from itertools import izip as zip
+
+from cassandra import WriteFailure
+from cassandra.concurrent import (execute_concurrent,
+                                  execute_concurrent_with_args)
+
+from dtest import Tester, debug
+from tools import since
+from utils.fileutils import size_of_files_in_dir
+from utils.funcutils import get_rate_limited_function
+
+
+def _set_cdc_on_table(session, table_name, value, ks_name=None):
+    """
+    Uses <session> to set CDC to <value> on <ks_name>.<table_name>.
+    """
+    table_string = ks_name + '.' + table_name if ks_name else table_name
+    value_string = 'true' if value else 'false'
+    stmt = 'ALTER TABLE ' + table_string + ' WITH CDC = ' + value_string
+
+    debug(stmt)
+    session.execute(stmt)
+
+
+def _get_set_cdc_func(session, ks_name, table_name):
+    """
+    Close over a session, keyspace name, and table name and return a function
+    that takes enables CDC on that keyspace if its argument is truthy and
+    otherwise disables it.
+    """
+    def set_cdc(value):
+        return _set_cdc_on_table(
+            session=session,
+            ks_name=ks_name, table_name=table_name,
+            value=value
+        )
+    return set_cdc
+
+
+def _get_commitlog_files(node_path):
+    commitlog_dir = os.path.join(node_path, 'commitlogs')
+    return {
+        os.path.join(commitlog_dir, name)
+        for name in os.listdir(commitlog_dir)
+    }
+
+
+def _get_cdc_raw_files(node_path, cdc_raw_dir_name='cdc_raw'):
+    commitlog_dir = os.path.join(node_path, cdc_raw_dir_name)
+    return {
+        os.path.join(commitlog_dir, name)
+        for name in os.listdir(commitlog_dir)
+    }
+
+
+@since('3.8')
+class TestCDC(Tester):
+    """
+    @jira_ticket CASSANDRA-8844
+
+    Test the correctness of some features of CDC, Change Data Capture, which
+    provides a view of the commitlog on tables for which it is enabled.
+    """
+
+    def prepare(self, ks_name,
+                table_name=None, cdc_enabled_table=None,
+                data_schema=None,
+                configuration_overrides=None):
+        """
+        Create a 1-node cluster, start it, create a keyspace, and if
+        <table_name>, create a table in that keyspace. If <cdc_enabled_table>,
+        that table is created with CDC enabled. If <data_schema>, use that
+        string to specify the schema of the table -- for example, a valid value
+        is '(a int PRIMARY KEY, b int)'. The <configuration_overrides> is
+        treated as a dict-like object and passed to
+        self.cluster.set_configuration_options.
+        """
+        config_defaults = {
+            'cdc_enabled': True,
+            # we want to be able to generate new segments quickly
+            'commitlog_segment_size_in_mb': 2,
+        }
+        if configuration_overrides is None:
+            configuration_overrides = {}
+        self.cluster.populate(1)
+        self.cluster.set_configuration_options(dict(config_defaults, **configuration_overrides))
+        self.cluster.start(wait_for_binary_proto=True)
+        node = self.cluster.nodelist()[0]
+        session = self.patient_cql_connection(node)
+        self.create_ks(session, ks_name, rf=1)
+
+        if table_name is not None:
+            self.assertIsNotNone(cdc_enabled_table, 'if creating a table in prepare, must specify whether or not CDC is enabled on it')
+            self.assertIsNotNone(data_schema, 'if creating a table in prepare, must specify its schema')
+            stmt = ('CREATE TABLE ' + table_name +
+                    ' ' + data_schema + ' '
+                    'WITH CDC = ' + ('true' if cdc_enabled_table else 'false'))
+            debug(stmt)
+            session.execute(stmt)
+
+        return node, session
+
+    def _assert_cdc_data_readable_on_round_trip(self, start_with_cdc_enabled):
+        """
+        Parameterized test asserting that data written to a table is still
+        readable after flipping the CDC flag on that table, then flipping it
+        again. Starts with CDC enabled if start_with_cdc_enabled, otherwise
+        starts with it disabled.
+        """
+        ks_name, table_name = 'ks', 'tab'
+        sequence = [True, False, True] if start_with_cdc_enabled else [False, True, False]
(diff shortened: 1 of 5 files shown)


cdc_test.py
```diff
@@ -0,0 +1,284 @@
...
+            # number of successfully completed statements...
+            rows_loaded += len([br for br in batch_results if br[0]])
+            # then, we make sure that the only failures are the expected
+            # WriteFailures.
+            self.assertEqual([],
+                             [result for (success, result) in batch_results
+                              if not success and not isinstance(result, WriteFailure)])
+            # Finally, if we find a WriteFailure, that means we've inserted all
+            # the CDC data we can and so we flip error_found to exit the loop.
+            if any(type(result) == WriteFailure for (_, result) in batch_results):
```
mambocab: @josh-mckenzie Here, I'm basically asserting `the write raised a WriteFailure`, and treating that as a proxy for `the write was rejected because we've reached cdc_total_space_in_mb`. Is there a way to actually assert the latter? I don't think C\* propagates the information in a way that the driver consumes, and I couldn't find any debug logging about rejecting changes to CDC commitlogs in your the 8844 diff.

mambocab:
I've pushed up some more changes to the tests that take advantage of having already loaded up the cdc table.

cdc_test.py
```diff
@@ -284,20 +288,44 @@ def test_insertion_and_commitlog_behavior_after_reaching_cdc_total_space(self):
...
+               '\n  '.join(f for f in cdc_segments))
+        debug('commitlog files are:\n  ' +
+               '\n  '.join(f for f in _get_commitlog_files(node.get_path())))
+        node.flush()
+
+        start, files_equal = time.time(), False
+        while not files_equal:
+            try:
+                rate_limited_debug('these files are in cdc_raw: ' + str(set(os.listdir(os.path.join(node.get_path(), 'cdc_raw')))))
+                self.assertEqual(
```
mambocab: @josh-mckenzie This assertion is failing -- so, after the `flush` above, no files ever go into `cdc_raw`. Is that a bug, or are my expectations wrong here?

cdc_test.py
```diff
@@ -0,0 +1,349 @@
...
+    Return the size of all files found in a non-recursive ls of the argument.
+    Based on http://stackoverflow.com/a/1392549
+    """
+    files = [os.path.join(dir_name, f) for f in os.listdir(dir_name)]
+    if verbose:
+        debug('getting sizes of these files: {}'.format(files))
+    return sum(os.path.getsize(f) for f in files)
+
+
+def get_rate_limited_function(func, limit):
```
josh-mckenzie: This looks like it could/should be promoted into a generic utility file in python

cdc_test.py
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+
+@since('3.8')
+class TestCDC(Tester):
+    """
+    @jira_ticket CASSANDRA-8844
+
+    Test the correctness of some features of CDC, Change Data Capture, which
+    provides an view of the commitlog on tables for which it is enabled.
```
josh-mckenzie: "a view"

cdc_test.py
josh-mckenzie: Definitely nothing in the logs from this failure. Confirming it's a WriteTimeoutException, while not completely 1:1 representative of it being CDC that caused it, should *probably* be good enough for us to rely on that as a signal that the throw was us. It's not super common to get a WTE from C* during dtests, is it? If not, I say we trust it. If so... we can pursue other options.

cdc_test.py
```diff
@@ -0,0 +1,349 @@
...
+        # WriteFailure. This should happen when the CDC commitlogs take up 1MB
+        # or more.
+        start, rows_loaded, error_found = time.time(), 0, False
+        rate_limited_debug = get_rate_limited_function(debug, 5)
+        debug('beginning data insert to fill CDC commitlogs')
+        while not error_found:
+            # We want to fail if inserting data takes too long. Locally this
+            # takes about a minute and a half, but let's be generous.
+            self.assertLessEqual(
+                (time.time() - start), 600,
```
josh-mckenzie: It takes 1.5 minutes to fill 1MB of data? That seems kinda awful. =/

cdc_test.py
josh-mckenzie: As a side note - do you have access to the writeType field of WriteTimeoutException? Should be serialized as a string in there as CDC.

josh-mckenzie: Aaaand a final point - coordinators don't pass back exceptions, so I'm going to add a NoSpamLogger rate limited log warning about being at CDC capacity when a segment is allocated. Should be a push with that commit relatively shortly and you should be able to grep the log for that.

cdc_test.py
josh-mckenzie: errr... you're doing a nodetool flush on segments w/active CDC data in them, so they *should* be going into cdc_raw. You can throw a debug print into CommitLogSegmentManagerCDC.discard(...) to indicate whether the segment being removed knows it has CDC or not. This def. isn't behavior we're reproducing testing on the C* isolated node case.

cdc_test.py
mambocab: +1, good idea.

cdc_test.py
```diff
@@ -0,0 +1,349 @@
...
+    def set_cdc(value):
+        return set_cdc_on_table(
+            session=session,
+            ks_name=ks_name, table_name=table_name,
+            value=value
+        )
+    return set_cdc
+
+
+def size_of_files_in_dir(dir_name, verbose=True):
```
mambocab: @ptnapoleon Josh's comment below made me think -- should we pull this out into `utils` somewhere?

cdc_test.py
mambocab: @ptnapoleon @knifewine Hmm. Would this be better as a class?

```
class get_rate_limited_function(object):  # or some other name
    def __init__(self, func, limit):
        self.func, self.limit = func, limit
        self.last_called = time.time()

    def __call__(self, *args, **kwargs):
        ...
```

Might be a more sensible way to persist state. I dunno.

cdc_test.py
ptnapoleon: +1

cdc_test.py
mambocab: I'm still futzing with it, but I think the problem may only happen because I've set `cdc_total_space_in_mb` to 1. When it's set that low, it looks like nothing gets written at all. I'll bisect over values for that, but I can insert values just fine when it's set to 16.

mambocab: @josh-mckenzie ^ So do we treat that as a bug? Should we disallow setting it that low?

cdc_test.py
```diff
@@ -0,0 +1,349 @@
...
+        elapsed = time.time() - state['last_called']
+        if elapsed >= limit:
+            func(*args, **kwargs)
+            state['last_called'] = time.time()
+    f.__name__ = '{n}_with_{s}s_rate_limit'.format(n=func.__name__, s=limit)
+
+    return f
+
+
+def _get_commitlog_files(node_path):
```
mambocab: Having second thoughts about this API. Should this take a `Node` instead?

cdc_test.py
mambocab: > I'm going to add a NoSpamLogger rate limited log warning

Not seeing this in `CommitLogSegmentManagerCDC.allocate`, or in a quick grep for `logger` in the CDC commit. Am I missing it?

cdc_test.py
mambocab: Sped this process up by using wider rows.

mambocab:
I've been working towards 4aa694eecf359e7b75f8ad3b8be00d9bdc026552; rebasing and squashing onto master now.

mambocab:
@ptnapoleon [This comment](https://github.com/riptano/datastax-jenkins-util/pull/67#issuecomment-226868771) means we aren't blocked on figuring out the `mock` dep, right?

ptnapoleon:
Yes, I believe that is resolved now

ptnapoleon:
PS, what is the state of this and me needing to review it? Properly covered by Josh? Waiting for more changes? Waiting on me?

mambocab:
@ptnapoleon ready for re-review by you, waiting for feedback from Josh. Thanks for checking.

cdc_test.py
```diff
@@ -0,0 +1,353 @@
...
+                debug("write failed (presumably because we've overrun "
+                      'designated CDC commitlog space) after '
+                      'loading {r} rows in {s:.2f}s'.format(
+                          r=rows_loaded,
+                          s=time.time() - start))
+                error_found = True
+
+        self.assertLess(0, rows_loaded,
+                        'No CDC rows inserted. This may happen when '
+                        'cdc_total_space_in_mb > commitlog_segment_size_in_mb')
```
josh-mckenzie: I am not reproducing this problem locally. I have cdc_total_space_in_mb set to 8192 and commitlog_segment_size_in_mb set to 32, commitlog_total_space_in_mb to 4096 and commit log segments are being moved to cdc_raw correctly.

The above doesn't make sense to me. By definition our cdc_total_space_in_mb needs to be > commitlog_segment_size_in_mb (default 4096 vs. 32). Is there a typo above?

cdc_test.py
josh-mckenzie: I'm going to add this as part of https://issues.apache.org/jira/browse/CASSANDRA-12018. I'll get on that immediately and see if I can't get it in by Friday. In the interim, I'd say write the test to check for WTE and we can update it when 12018 goes in.

cdc_test.py
mambocab: Ah, I had it backwards -- I mean if `commitlog_segment_size_in_mb` > `cdc_total_space_in_mb`, CDC segments are never allocated.

I can write a dtest just to repro this if you like.

josh-mckenzie: Hm. What happens if commitlog_total_space_in_mb < commitlog_segment_size_in_mb? This sounds like a pathologically bad situation in either case. I'll poke around and see if we handle that.

ptnapoleon: Isn't there an assertNotZero, or something similar? Is that not better here? It doesn't really matter though

ptnapoleon:
Everything LGTM on first close pass

cdc_test.py
mambocab: > What happens if commitlog_total_space_in_mb < commitlog_segment_size_in_mb?

In my tests, the first external sign of the misconfiguration is that you can't write to CDC tables. There could be something logged at `WARN` or lower, I haven't checked. If it logged at `ERROR`, I'd have seen tests fail because of it.

mambocab:
I'm going to run them on CassCI here to double-check that CCM and the `mock` dependency are dealt with properly:

https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/130/

Once that's done, I got a +1 from Josh offline, so I'll merge.

mambocab:
Oh, and @ptnapoleon:

> Isn't there an `assertNotZero`, or something similar?

Not an `assertNotZero`, no. I could use `assertTrue`, but I think this is clearer.

mambocab:
Ran clean. Merging, thanks for the reviews.

PR #1011, opened 2016-05-31
annotated jira-11927

smccarthy788:
@ptnapoleon, please review.

### DIFF
--- a/replication_test.py
+++ b/replication_test.py
@@ -196,7 +196,7 @@ class ReplicationTest(Tester):
             print("-" * 40)
 
     @known_failure(failure_source='test',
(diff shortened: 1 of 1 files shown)


replication_test.py
```diff
@@ -195,6 +195,9 @@ def pprint_trace(self, trace):
                 print("%s\t%s\t%s\t%s" % (t.source, t.source_elapsed, t.description, t.thread_name))
             print("-" * 40)
 
+    @known_failure(failure_source='test',
+                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11439',
```
ptnapoleon: This ticket number does not match the PR name, so one of these is wrong

ptnapoleon:
-1

PR #1012, opened 2016-05-31
Cstar 67

ptnapoleon:
This isn't ready yet, but @mambocab and I want to offer early feedback in a useful place, so I'm opening this.

### DIFF
--- a/assertions.py
+++ b/assertions.py
@@ -121,7 +121,6 @@ def assert_one(session, query, expected, cl=None):
     list_res = rows_to_list(res)
     assert list_res == [expected], "Expected {} from {}, but got {}".format([expected], query, list_res)
 
-
 def assert_none(session, query, cl=None):
     """
     Assert query returns nothing
@@ -137,7 +136,6 @@ def assert_none(session, query, cl=None):
     res = session.execute(simple_query)
     list_res = rows_to_list(res)
     assert list_res == [], "Expected nothing from {}, but got {}".format(query, list_res)
-
 
 def assert_all(session, query, expected, cl=None, ignore_order=False):
     """
--- a/assertions.py
+++ b/assertions.py
@@ -1,85 +1,198 @@
-import re
+from nose.tools import assert_regexp_matches, assert_equal
 
-from cassandra import (ConsistencyLevel, InvalidRequest, ReadFailure,
+from cassandra import (InvalidRequest, ReadFailure,
                        ReadTimeout, Unauthorized, Unavailable, WriteFailure,
                        WriteTimeout)
 from cassandra.query import SimpleStatement
 
 from tools import rows_to_list
 
+"""
+The assertion methods in this file are used to structure, execute, and test different queries and scenarios. Use these anytime you are trying
+to check the content of a table, the row count of a table, if a query should raise an exception, etc. These methods handle error messaging
+well, and will help discovering and treating bugs.
 
-def assert_unavailable(fun, *args):
+An example:
+Imagine some table, test:
+
+    id | name
+    1  | John Doe
+    2  | Jane Doe
+
+We could assert the row count is 2 by using:
+    assert_row_count(session, 'test', 2)
+
+After inserting [3, 'Alex Smith'], we can ensure the table is correct by:
+    assert_all(session, "SELECT * FROM test", [[1, 'John Doe'], [2, 'Jane Doe'], [3, 'Alex Smith']])
+or we could check the insert was successful:
+    assert_one(session, "SELECT * FROM test WHERE id = 3", [3, 'Alex Smith'])
+
+We could remove all rows in test, and assert this was sucessful with:
+    assert_none(session, "SELECT * FROM test")
+
+Perhaps we want to assert invalid queries will throw an exception:
+    assert_invalid(session, "SELECT FROM test")
+or, maybe after shutting down all the nodes, we want to assert an Unavailable exception is raised:
+    assert_unavailable(session.execute, "SELECT * FROM test")
+    OR
+    assert_exception(session, "SELECT * FROM test", expected=Unavailable)
+
+"""
+
+def _assert_exception(fun, *args, **kwargs):
+    matching = kwargs.pop('matching', None)
+    expected = kwargs['expected']
     try:
         if len(args) == 0:
             fun(None)
         else:
             fun(*args)
-    except (Unavailable, WriteTimeout, WriteFailure, ReadTimeout, ReadFailure) as e:
-        pass
+    except expected as e:
+        if matching is not None:
+            assert_regexp_matches(str(e), matching)
     except Exception as e:
-        assert False, "Expecting unavailable exception, got: " + str(e)
+        raise e
     else:
-        assert False, "Expecting unavailable exception but no exception was raised"
+        assert False, "Expecting query to raise an exception, but nothing was raised."
+
+
+def assert_exception(session, query, matching=None, expected=None):
+    if expected is None:
+        assert False, "Expected exception should not be None. Your test code is wrong, please set `expected`."
+
+    _assert_exception(session.execute, query, matching=matching, expected=expected)
+
+
+def assert_unavailable(fun, *args):
+    """
+    Attempt to execute a function, and assert Unavailable, WriteTimeout, WriteFailure, ReadTimeout, or ReadFailure exception is raised.
+    @param fun Function to be executed
+    @param *args Arguments to be passed to the function
+
+    Examples:
+    assert_unavailable(session2.execute, "SELECT * FROM ttl_table;")
+    assert_unavailable(lambda c: debug(c.execute(statement)), session)
+    """
+    _assert_exception(fun, *args, expected=(Unavailable, WriteTimeout, WriteFailure, ReadTimeout, ReadFailure))
 
 
 def assert_invalid(session, query, matching=None, expected=InvalidRequest):
-    try:
-        res = session.execute(query)
-        assert False, "Expecting query to be invalid: got %s" % res
-    except AssertionError as e:
-        raise e
-    except expected as e:
-        msg = str(e)
-        if matching is not None:
-            assert re.search(matching, msg), "Error message does not contain " + matching + " (error = " + msg + ")"
+    """
+    Attempt to issue a query and assert that the query is invalid.
+    @param session Session to use
+    @param query Invalid query to run
+    @param matching Optional error message string contained within expected exception
+    @param expected Exception expected to be raised by the invalid query
+
+    Examples:
+    assert_invalid(session, 'DROP USER nonexistent', "nonexistent doesn't exist")
+    """
+    assert_exception(session, query, matching=matching, expected=expected)
 
 
 def assert_unauthorized(session, query, message):
     """
     Attempt to issue a query, and assert Unauthorized is raised.
-    @param message Expected error message
     @param session Session to use
     @param query Unauthorized query to run
+    @param message Expected error message
+
+    Examples:
+    assert_unauthorized(session, "ALTER USER cassandra NOSUPERUSER", "You aren't allowed to alter your own superuser status")
+    assert_unauthorized(cathy, "ALTER TABLE ks.cf ADD val int", "User cathy has no ALTER permission on <table ks.cf> or any of its parents")
     """
-    assert_invalid(session, query, message, Unauthorized)
+    assert_exception(session, query, matching=message, expected=Unauthorized)
 
 
-def assert_one(session, query, expected, cl=ConsistencyLevel.ONE):
+def assert_one(session, query, expected, cl=None):
+    """
+    Assert query returns one row.
+    @param session Session to use
+    @param query Query to run
+    @param expected Expected results from query
+    @param cl Optional Consistency Level setting. Default ONE
+
+    Examples:
+    assert_one(session, "LIST USERS", ['cassandra', True])
+    assert_one(session, query, [0, 0])
+    """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
     list_res = rows_to_list(res)
-    assert list_res == [expected], "Expected %s from %s, but got %s" % ([expected], query, list_res)
+    assert list_res == [expected], "Expected {} from {}, but got {}".format([expected], query, list_res)
 
 
-def assert_none(session, query, cl=ConsistencyLevel.ONE):
+def assert_none(session, query, cl=None):
+    """
+    Assert query returns nothing
+    @param session Session to use
+    @param query Query to run
+    @param cl Optional Consistency Level setting. Default ONE
+
+    Examples:
+    assert_none(self.session1, "SELECT * FROM test where key=2;")
+    assert_none(cursor, "SELECT * FROM test WHERE k=2", cl=ConsistencyLevel.SERIAL)
+    """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
     list_res = rows_to_list(res)
-    assert list_res == [], "Expected nothing from %s, but got %s" % (query, list_res)
+    assert list_res == [], "Expected nothing from {}, but got {}".format(query, list_res)
 
 
-def assert_all(session, query, expected, cl=ConsistencyLevel.ONE, ignore_order=False):
+def assert_all(session, query, expected, cl=None, ignore_order=False):
+    """
+    Assert query returns all expected items optionally in the correct order
+    @param session Session in use
+    @param query Query to run
+    @param expected Expected results from query
+    @param cl Optional Consistency Level setting. Default ONE
+    @param ignore_order Optional boolean flag determining whether response is ordered
+
+    Examples:
+    assert_all(session, "LIST USERS", [['aleksey', False], ['cassandra', True]])
+    assert_all(self.session1, "SELECT * FROM ttl_table;", [[1, 42, 1, 1]])
+    """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
     list_res = rows_to_list(res)
     if ignore_order:
         expected = sorted(expected)
         list_res = sorted(list_res)
-    assert list_res == expected, "Expected %s from %s, but got %s" % (expected, query, list_res)
+    assert list_res == expected, "Expected {} from {}, but got {}".format(expected, query, list_res)
 
 
 def assert_almost_equal(*args, **kwargs):
+    """
+    Assert variable number of arguments all fall within a margin of error.
+    @params *args variable number of numerical arguments to check
+    @params error Optional margin of error. Default 0.16
+    @params error_message Optional error message to print. Default ''
+
+    Examples:
+    assert_almost_equal(sizes[2], init_size)
+    assert_almost_equal(ttl_session1, ttl_session2[0][0], error=0.005)
+    """
     error = kwargs['error'] if 'error' in kwargs else 0.16
     vmax = max(args)
     vmin = min(args)
     error_message = '' if 'error_message' not in kwargs else kwargs['error_message']
-    assert vmin > vmax * (1.0 - error) or vmin == vmax, "values not within %.2f%% of the max: %s (%s)" % (error * 100, args, error_message)
+    assert vmin > vmax * (1.0 - error) or vmin == vmax, "values not within {.2f}% of the max: {} ({})".format(error * 100, args, error_message)
 
 
-def assert_row_count(session, table_name, expected):
-    """ Function to validate the row count expected in table_name """
+def assert_row_count(session, table_name, expected, where=None):
+    """
+    Assert the number of rows in a table matches expected.
+    @params session Session to use
+    @param table_name Name of the table to query
+    @param expected Number of rows expected to be in table
 
-    query = "SELECT count(*) FROM {};".format(table_name)
+    Examples:
+    assert_row_count(self.session1, 'ttl_table', 1)
+    """
+    if where is not None:
+        query = "SELECT count(*) FROM {} WHERE {};".format(table_name, where)
+    else:
+        query = "SELECT count(*) FROM {};".format(table_name)
     res = session.execute(query)
     count = res[0][0]
     assert count == expected, "Expected a row count of {} in table '{}', but got {}".format(
@@ -89,6 +202,16 @@ def assert_row_count(session, table_name, expected):
 
 def assert_crc_check_chance_equal(session, table, expected, ks="ks", view=False):
     """
+    Assert crc_check_chance equals expected for a given table or view
+    @param session Session to use
+    @param table Name of the table or view to check
+    @param ks Optional Name of the keyspace
+    @param view Optional Boolean flag indicating if the table is a view
+
+    Examples:
+    assert_crc_check_chance_equal(session, "compression_opts_table", 0.25)
+    assert_crc_check_chance_equal(session, "t_by_v", 0.5, view=True)
+
     driver still doesn't support top-level crc_check_chance property,
     so let's fetch directly from system_schema
     """
@@ -102,3 +225,15 @@ def assert_crc_check_chance_equal(session, table, expected, ks="ks", view=False)
                    "SELECT crc_check_chance from system_schema.tables WHERE keyspace_name = 'ks' AND "
                    "table_name = '{table}';".format(table=table),
                    [expected])
+
+
+def assert_length_equal(object_with_length, expected_length):
+    """
+    Assert an object has a specific length.
+    @param object_with_length The object whose length will be checked
+    @param expected_length The expected length of the object
+
+    Examples:
+    assert_length_equal(res, nb_counter)
+    """
+    assert_equal(len(object_with_length), expected_length, object_with_length)
--- a/assertions.py
+++ b/assertions.py
@@ -5,7 +5,7 @@ from cassandra import (InvalidRequest, ReadFailure,
                        WriteTimeout)
 from cassandra.query import SimpleStatement
 
-from tools import rows_to_list
+import tools
 
 """
 The assertion methods in this file are used to structure, execute, and test different queries and scenarios. Use these anytime you are trying
@@ -38,6 +38,7 @@ or, maybe after shutting down all the nodes, we want to assert an Unavailable ex
     assert_exception(session, "SELECT * FROM test", expected=Unavailable)
 
 """
+
 
 def _assert_exception(fun, *args, **kwargs):
     matching = kwargs.pop('matching', None)
@@ -118,8 +119,9 @@ def assert_one(session, query, expected, cl=None):
     """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
-    list_res = rows_to_list(res)
+    list_res = tools.rows_to_list(res)
     assert list_res == [expected], "Expected {} from {}, but got {}".format([expected], query, list_res)
+
 
 def assert_none(session, query, cl=None):
     """
@@ -134,8 +136,9 @@ def assert_none(session, query, cl=None):
     """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
-    list_res = rows_to_list(res)
+    list_res = tools.rows_to_list(res)
     assert list_res == [], "Expected nothing from {}, but got {}".format(query, list_res)
+
 
 def assert_all(session, query, expected, cl=None, ignore_order=False):
     """
@@ -152,7 +155,7 @@ def assert_all(session, query, expected, cl=None, ignore_order=False):
     """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
-    list_res = rows_to_list(res)
+    list_res = tools.rows_to_list(res)
     if ignore_order:
         expected = sorted(expected)
         list_res = sorted(list_res)
--- a/assertions.py
+++ b/assertions.py
@@ -237,4 +237,6 @@ def assert_length_equal(object_with_length, expected_length):
     Examples:
     assert_length_equal(res, nb_counter)
     """
-    assert_equal(len(object_with_length), expected_length, object_with_length)
+    assert_equal(len(object_with_length), expected_length,
+                 "Expected {} to have length {}, but instead is of length {}".format(object_with_length,
+                                                                                     expected_length, len(object_with_length)))
--- a/batch_test.py
+++ b/batch_test.py
@@ -185,7 +185,8 @@ class TestBatch(Tester):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner')
             APPLY BATCH
         """, consistency_level=ConsistencyLevel.ANY)
-        assert_all(session, "SElECT * FROM users", [[0, u'Jack', u'Sparrow'], [1, u'Will', u'Turner']])
+        session.execute(query)
+        assert_all(session, "SElECT * FROM users", [[1, u'Will', u'Turner'], [0, u'Jack', u'Sparrow']])
 
     def acknowledged_by_batchlog_not_set_when_batchlog_write_fails_test(self):
         """ Test that acknowledged_by_batchlog is False if batchlog can't be written """
--- a/batch_test.py
+++ b/batch_test.py
@@ -5,7 +5,7 @@ from unittest import skipIf
 from cassandra import ConsistencyLevel, Timeout, Unavailable
 from cassandra.query import SimpleStatement
 
-from assertions import assert_invalid, assert_one, assert_unavailable
+from assertions import assert_invalid, assert_one, assert_unavailable, assert_all
 from dtest import CASSANDRA_DIR, Tester, debug
 from tools import since
 
@@ -34,8 +34,7 @@ class TestBatch(Tester):
             UPDATE clicks SET total = total + 1 WHERE userid = 2 and url = 'http://baz.com'
             APPLY BATCH
         """)
-        rows = session.execute("SELECT total FROM clicks")
-        assert [list(rows[0]), list(rows[1]), list(rows[2])] == [[1], [1], [1]], rows
+        assert_all(session, "SELECT total FROM clicks", [[1], [1], [1]])
 
     def counter_batch_rejects_regular_mutations_test(self):
         """ Test that counter batch rejects non-counter mutations """
@@ -60,9 +59,7 @@ class TestBatch(Tester):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner')
             APPLY BATCH
         """)
-        rows = session.execute("SELECT * FROM users")
-        res = sorted(rows)
-        assert [list(res[0]), list(res[1])] == [[0, u'Jack', u'Sparrow'], [1, u'Will', u'Turner']], res
+        assert_all(session, "SELECT * FROM users", [[1, u'Will', u'Turner'], [0, u'Jack', u'Sparrow']])
 
     @since('3.0')
     def logged_batch_gcgs_below_threshold_single_table_test(self):
@@ -151,9 +148,7 @@ class TestBatch(Tester):
             INSERT INTO users (id, firstname, lastname) VALUES (2, 'Elizabeth', 'Swann')
             APPLY BATCH
         """)
-        rows = session.execute("SELECT * FROM users")
-        res = sorted(rows)
-        assert [list(res[0]), list(res[1])] == [[0, u'Jack', u'Sparrow'], [2, u'Elizabeth', u'Swann']], res
+        assert_all(session, "SELECT * FROM users", [[0, u'Jack', u'Sparrow'], [2, u'Elizabeth', u'Swann']])
 
     def unlogged_batch_rejects_counter_mutations_test(self):
         """ Test that unlogged batch rejects counter mutations """
@@ -190,8 +185,7 @@ class TestBatch(Tester):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner')
             APPLY BATCH
         """, consistency_level=ConsistencyLevel.ANY)
-        session.execute(query)
-        assert True
+        assert_all(session, "SElECT * FROM users", [[0, u'Jack', u'Sparrow'], [1, u'Will', u'Turner']])
 
     def acknowledged_by_batchlog_not_set_when_batchlog_write_fails_test(self):
         """ Test that acknowledged_by_batchlog is False if batchlog can't be written """
@@ -226,9 +220,8 @@ class TestBatch(Tester):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner')
             APPLY BATCH
         """)
-        rows = session.execute("SELECT id, writetime(firstname), writetime(lastname) FROM users")
-        res = sorted(rows)
-        assert [list(res[0]), list(res[1])] == [[0, 1111111111111111, 1111111111111111], [1, 1111111111111111, 1111111111111111]], res
+        query = "SELECT id, writetime(firstname), writetime(lastname) FROM users"
+        assert_all(session, query, [[1, 1111111111111111, 1111111111111111], [0, 1111111111111111, 1111111111111111]])
 
     def only_one_timestamp_is_valid_test(self):
         """ Test that TIMESTAMP must not be used in the statements within the batch. """
@@ -249,9 +242,9 @@ class TestBatch(Tester):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner') USING TIMESTAMP 1111111111111112
             APPLY BATCH
         """)
-        rows = session.execute("SELECT id, writetime(firstname), writetime(lastname) FROM users")
-        res = sorted(rows)
-        assert [list(res[0]), list(res[1])] == [[0, 1111111111111111, 1111111111111111], [1, 1111111111111112, 1111111111111112]], res
+
+        query = "SELECT id, writetime(firstname), writetime(lastname) FROM users"
+        assert_all(session, query, [[1, 1111111111111112, 1111111111111112], [0, 1111111111111111, 1111111111111111]])
 
     def multi_table_batch_for_10554_test(self):
         """ Test a batch on 2 tables having different columns, restarting the node afterwards, to reproduce CASSANDRA-10554 """
--- a/batch_test.py
+++ b/batch_test.py
@@ -186,7 +186,7 @@ class TestBatch(Tester):
             APPLY BATCH
         """, consistency_level=ConsistencyLevel.ANY)
         session.execute(query)
-        assert_all(session, "SElECT * FROM users", [[1, u'Will', u'Turner'], [0, u'Jack', u'Sparrow']])
+        assert_all(session, "SELECT * FROM users", [[1, u'Will', u'Turner'], [0, u'Jack', u'Sparrow']])
 
     def acknowledged_by_batchlog_not_set_when_batchlog_write_fails_test(self):
         """ Test that acknowledged_by_batchlog is False if batchlog can't be written """
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -227,9 +227,7 @@ class TestBootstrap(Tester):
 
         # check if 2nd bootstrap succeeded
         session = self.patient_exclusive_cql_connection(node3)
-        rows = list(session.execute("SELECT bootstrapped FROM system.local WHERE key='local'"))
-        assert len(rows) == 1
-        assert rows[0][0] == 'COMPLETED', rows[0][0]
+        assert_one(session, "SELECT bootstrapped FROM system.local WHERE key='local'", ['COMPLETED'])
 
     def manual_bootstrap_test(self):
         """
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -188,9 +188,8 @@ class TestBootstrap(Tester):
                                        'replication(factor=2)', '-rate', 'threads=8'],
                                       capture_output=True)
 
-        if stdout and "FAILURE" in stdout:
-            debug(stdout)
-            assert False, "Cannot read inserted data after bootstrap"
+        if stdout is not None:
+            self.assertNotIn("FAILURE", stdout)
 
     @since('2.2')
     def bootstrap_with_reset_bootstrap_state_test(self):
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -204,21 +204,21 @@ class TestCompaction(Tester):
         self.assertEqual(len(expired_sstables), expected_sstable_count)
         # write a new sstable to make DTCS check for expired sstables:
         for x in range(0, 100):
-            session.execute('insert into cf (key, val) values (%d, %d)' % (x, x))
+            session.execute('insert into cf (key, val) values ({}, {})'.format(x, x))
         node1.flush()
         time.sleep(5)
         # we only check every 10 minutes - sstable should still be there:
         for expired_sstable in expired_sstables:
-            assert expired_sstable in node1.get_sstables('ks', 'cf')
+            self.assertIn(expired_sstable, node1.get_sstables('ks', 'cf'))
 
         session.execute("alter table cf with compaction =  {'class':'DateTieredCompactionStrategy', 'max_sstable_age_days':0.00035, 'min_threshold':2, 'expired_sstable_check_frequency_seconds':0}")
         time.sleep(1)
         for x in range(0, 100):
-            session.execute('insert into cf (key, val) values (%d, %d)' % (x, x))
+            session.execute('insert into cf (key, val) values ({}, {})'.format(x, x))
         node1.flush()
         time.sleep(5)
         for expired_sstable in expired_sstables:
-            assert expired_sstable not in node1.get_sstables('ks', 'cf')
+            self.assertNotIn(expired_sstable, node1.get_sstables('ks', 'cf'))
 
     def compaction_throughput_test(self):
         """
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -6,8 +6,8 @@ import subprocess
 import tempfile
 import time
 
-from assertions import assert_none, assert_one
 from ccmlib import common
+from assertions import assert_none, assert_one, assert_length_equal
 from dtest import Tester, debug
 from nose.tools import assert_equal
 from tools import known_failure, since
@@ -319,7 +319,7 @@ class TestCompaction(Tester):
             session.execute("UPDATE ks.large SET properties[%i] = '%s' WHERE userid = 'user'" % (i, get_random_word(strlen)))
 
         ret = list(session.execute("SELECT properties from ks.large where userid = 'user'"))
-        assert len(ret) == 1
+        assert_length_equal(ret, 1)
         self.assertEqual(200, len(ret[0][0].keys()))
 
         node.flush()
@@ -330,8 +330,7 @@ class TestCompaction(Tester):
         node.watch_log_for('{} large partition ks/large:user \({}\)'.format(verb, sizematcher), from_mark=mark, timeout=180)
 
         ret = list(session.execute("SELECT properties from ks.large where userid = 'user'"))
-
-        assert len(ret) == 1
+        assert_length_equal(ret, 1)
         self.assertEqual(200, len(ret[0][0].keys()))
 
     def disable_autocompaction_nodetool_test(self):
--- a/configuration_test.py
+++ b/configuration_test.py
@@ -129,13 +129,10 @@ class TestConfiguration(Tester):
             if 'compression' in result:
                 params = result
 
-        assert params is not '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result)
-
+        self.assertNotEqual(params, '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result))
         chunk_string = "chunk_length_kb" if self.cluster.version() < '3.0' else "chunk_length_in_kb"
         chunk_length = int(re.search("{chunk}.*?:.*?'(\d*?)'".format(chunk=chunk_string), result).groups()[0])
-
-        assert chunk_length == value, "Expected chunk_length: %s.  We got: %s" % (value, chunk_length)
-
+        self.assertEqual(chunk_length, value, "Expected chunk_length: {}.  We got: {}".format(value, chunk_length))
 
 
 def write_to_trigger_fsync(session, ks, table):
--- a/configuration_test.py
+++ b/configuration_test.py
@@ -137,6 +137,7 @@ class TestConfiguration(Tester):
         assert chunk_length == value, "Expected chunk_length: %s.  We got: %s" % (value, chunk_length)
 
 
+
 def write_to_trigger_fsync(session, ks, table):
     """
     Given a session, a keyspace name, and a table name, inserts enough values
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -9,7 +9,7 @@ from cassandra import ConsistencyLevel
 from cassandra.query import SimpleStatement
 
 from assertions import assert_none, assert_unavailable, assert_length_equal
-from dtest import DISABLE_VNODES, Tester, debug
+from dtest import DISABLE_VNODES, Tester, debug, MultiError
 from tools import (create_c1c2_table, insert_c1c2, insert_columns,
                    known_failure, query_c1c2, rows_to_list, since)
 from nose.tools import assert_greater_equal
@@ -536,7 +536,8 @@ class TestAccuracy(TestHelper):
                 break
 
         if not exceptions_queue.empty():
-            raise exceptions_queue.get()[1]
+            _, exceptions, tracebacks = zip(*exceptions_queue.queue)
+            raise MultiError(exceptions=exceptions, tracebacks=tracebacks)
 
     def test_simple_strategy_users(self):
         """
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -197,7 +197,7 @@ class TestHelper(Tester):
         Return the current counter value. If we find no value we return zero
         because after the next update the counter will become one.
         """
-        statement = SimpleStatement("SELECT c from counters WHERE id = {}".format(id,), consistency_level=consistency)
+        statement = SimpleStatement("SELECT c from counters WHERE id = {}".format(id), consistency_level=consistency)
         ret = rows_to_list(session.execute(statement))
         return ret[0][0] if ret else 0
 
@@ -218,7 +218,7 @@ class TestAvailability(TestHelper):
 
         num_alive = nodes
         for node in xrange(nodes):
-            debug('Testing node {} in single dc with {} nodes alive'.format(node, num_alive,))
+            debug('Testing node {} in single dc with {} nodes alive'.format(node, num_alive))
             session = self.patient_exclusive_cql_connection(cluster.nodelist()[node], self.ksname)
             for combination in combinations:
                 self._test_insert_query_from_node(session, 0, [rf], [num_alive], *combination)
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -12,6 +12,7 @@ from assertions import assert_none, assert_unavailable, assert_length_equal
 from dtest import DISABLE_VNODES, Tester, debug
 from tools import (create_c1c2_table, insert_c1c2, insert_columns,
                    known_failure, query_c1c2, rows_to_list, since)
+from nose.tools import assert_greater_equal
 
 
 class TestHelper(Tester):
@@ -447,8 +448,8 @@ class TestAccuracy(TestHelper):
                 for s in sessions:
                     if outer.query_user(s, n, val, read_cl, check_ret=strong_consistency):
                         num += 1
-                self.assertGreaterEqual(num, write_nodes, "Failed to read value from sufficient number of nodes, required {} but got {} - [{}, {}]"
-                                        .format(write_nodes, num, n, val))
+                assert_greater_equal(num, write_nodes, "Failed to read value from sufficient number of nodes, required {} but got {} - [{}, {}]"
+                                     .format(write_nodes, num, n, val))
 
             for n in xrange(start, end):
                 age = 30
@@ -484,8 +485,8 @@ class TestAccuracy(TestHelper):
                 results = []
                 for s in sessions:
                     results.append(outer.query_counter(s, n, val, read_cl, check_ret=strong_consistency))
-                self.assertGreaterEqual(results.count(val), write_nodes, "Failed to read value from sufficient number of nodes, required {} nodes to have a counter "
-                                        "value of {} at key {}, instead got these values: {}".format(write_nodes, val, n, results))
+                assert_greater_equal(results.count(val), write_nodes, "Failed to read value from sufficient number of nodes, required {} nodes to have a counter "
+                                     "value of {} at key {}, instead got these values: {}".format(write_nodes, val, n, results))
 
             for n in xrange(start, end):
                 c = outer.read_counter(sessions[0], n, ConsistencyLevel.ALL)
@@ -535,7 +536,7 @@ class TestAccuracy(TestHelper):
                 break
 
         if not exceptions_queue.empty():
-            raise exceptions_queue.get()
+            raise exceptions_queue.get()[1]
 
     def test_simple_strategy_users(self):
         """
@@ -784,7 +785,7 @@ class TestConsistency(Tester):
 
             # value 0, 1 and 2 have been deleted
             for i in xrange(1, 4):
-                self.assertEqual(res[i - 1][1], 'value {}'.format(i + 2), 'Expecting value {}, got {} ({})'.format(i + 2, res[i - 1][1], str(res)))
+                self.assertEqual(res[i - 1][1], 'value{}'.format(i + 2), 'Expecting value{}, got {} ({})'.format(i + 2, res[i - 1][1], str(res)))
 
             truncate_statement = SimpleStatement('TRUNCATE cf', consistency_level=ConsistencyLevel.QUORUM)
             session.execute(truncate_statement)
@@ -820,7 +821,7 @@ class TestConsistency(Tester):
         # Query first column
         session = self.patient_cql_connection(node1, 'ks')
 
-        assert_none(sesion, "SELECT c, v FROM cf WHERE key=\'k0\' LIMIT 1", cl=ConsistencyLevel.QUORUM)
+        assert_none(session, "SELECT c, v FROM cf WHERE key=\'k0\' LIMIT 1", cl=ConsistencyLevel.QUORUM)
 
     def short_read_quorum_delete_test(self):
         """
@@ -862,7 +863,7 @@ class TestConsistency(Tester):
         # we read the first row in the partition (so with a LIMIT 1) and A and B answer first.
         node3.flush()
         node3.stop(wait_other_notice=True)
-        self.assert_none(session, "SELECT * FROM t WHERE id = 0 LIMIT 1", cl=ConsistencyLevel.QUORUM)
+        assert_none(session, "SELECT * FROM t WHERE id = 0 LIMIT 1", cl=ConsistencyLevel.QUORUM)
 
     def readrepair_test(self):
         cluster = self.cluster
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -2,14 +2,13 @@ import Queue
 import sys
 import threading
 import time
-import traceback
 from collections import OrderedDict
 from copy import deepcopy
 
 from cassandra import ConsistencyLevel
 from cassandra.query import SimpleStatement
 
-from assertions import assert_none, assert_unavailable
+from assertions import assert_none, assert_unavailable, assert_length_equal
 from dtest import DISABLE_VNODES, Tester, debug
 from tools import (create_c1c2_table, insert_c1c2, insert_columns,
                    known_failure, query_c1c2, rows_to_list, since)
@@ -185,10 +184,11 @@ class TestHelper(Tester):
         res = session.execute(statement)
         ret = rows_to_list(res)
         if check_ret:
-            assert ret[0][1] == val, "Got %s from %s, expected %s at %s" % (ret[0][1],
-                                                                            session.cluster.contact_points,
-                                                                            val,
-                                                                            self._name(consistency))
+            self.assertEqual(ret[0][1], val, "Got {} from {}, expected {} at {}".format(ret[0][1],
+                                                                                        session.cluster.contact_points,
+                                                                                        val,
+                                                                                        self._name(consistency)))
+
         return ret[0][1] if ret else 0
 
     def read_counter(self, session, id, consistency):
@@ -447,9 +447,8 @@ class TestAccuracy(TestHelper):
                 for s in sessions:
                     if outer.query_user(s, n, val, read_cl, check_ret=strong_consistency):
                         num += 1
-                assert num >= write_nodes, \
-                    "Failed to read value from sufficient number of nodes, required %d but  got %d - [%d, %s]" \
-                    % (write_nodes, num, n, val)
+                self.assertGreaterEqual(num, write_nodes, "Failed to read value from sufficient number of nodes, required {} but got {} - [{}, {}]"
+                                        .format(write_nodes, num, n, val))
 
             for n in xrange(start, end):
                 age = 30
@@ -485,9 +484,8 @@ class TestAccuracy(TestHelper):
                 results = []
                 for s in sessions:
                     results.append(outer.query_counter(s, n, val, read_cl, check_ret=strong_consistency))
-                assert results.count(val) >= write_nodes, \
-                    "Failed to read value from sufficient number of nodes, required %d nodes to have a counter " \
-                    "value of %s at key %d, instead got these values: %s" % (write_nodes, val, n, results)
+                self.assertGreaterEqual(results.count(val), write_nodes, "Failed to read value from sufficient number of nodes, required {} nodes to have a counter "
+                                        "value of {} at key {}, instead got these values: {}".format(write_nodes, val, n, results))
 
             for n in xrange(start, end):
                 c = outer.read_counter(sessions[0], n, ConsistencyLevel.ALL)
@@ -540,7 +538,8 @@ class TestAccuracy(TestHelper):
             err_type, err, err_traceback = exceptions_queue.get()
             debug("Failed with exception {}: {}".format(err_type, err.message))
             traceback.print_exception(err_type, err, err_traceback)
-            assert False, err.message
+
+            self.fail(err.message)
 
     def test_simple_strategy_users(self):
         """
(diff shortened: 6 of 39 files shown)


assertions.py
```diff
@@ -7,77 +8,165 @@
...
+assert_crc_check_chance_equal(session, "t_by_v", 0.5, view=True)
+"""
+
+
+def __assert_exception(fun, *args, **kwargs):
+    try:
+        matching = kwargs['matching']
+    except KeyError:
+        matching = None
+    expected = kwargs['expected']
```
mambocab: This code is, other than argument order, a less-readable implementation of

```python
def __assert_exception(fun, *args, expected, matching=None, **kwargs):
    ...
```

(Those optional and keyword arguments may be in an illegal order; I'm not sure. But, grabbing things out of `kwargs` is unnecessary here.)

mambocab: Also, it's more conventional to prefix with a single underscore.

mambocab:
:fireworks: Get ready for a proposed scope increeeeease :fireworks: 

In the long haul, imo, it'd be good to have tests for our utility functions, so we don't stop catching C\* bugs when we break our code. This might be a good opportunity to start introducing some meta-tests so we know these functions work correctly.

assertions.py
```diff
@@ -7,77 +8,165 @@
...
+def assert_exception(session, query, matching=None, expected=None):
+    if expected is None:
+        assert False, "Expected exception should not be None. Your test code is wrong, please set `expected`."
+
+    __assert_exception(session.execute, query, matching=matching, expected=expected)
+
+
+def assert_unavailable(fun, *args):
+    """
+    Attempt to execute a function, and assert Unavailable exception is raised.
```
mambocab: Not accurate -- any one of `(Unavailable, WriteTimeout, WriteFailure, ReadTimeout, ReadFailure)` is valid.

assertions.py
```diff
@@ -7,77 +8,165 @@
...
+    @param fun Function to be executed
+    @param *args Arguments to be passed to the function
+    """
+    __assert_exception(fun, *args, expected=(Unavailable, WriteTimeout, WriteFailure, ReadTimeout, ReadFailure))
```
mambocab: I don't like the how `assert_unavailable` asserts that `fun` raises one of 5 exceptions, 4 of which are not `Unavailable`.

I understand that breaking this API is probably out of scope right now, but maybe file a JIRA to rename this.

assertions.py
```diff
@@ -7,77 +7,164 @@
...
+            assert_regexp_matches(matching, str(e))
     except Exception as e:
-        assert False, "Expecting unavailable exception, got: " + str(e)
+        raise e
     else:
-        assert False, "Expecting unavailable exception but no exception was raised"
+        assert False, "Expecting query to raise an exception, but nothing was raised."
+
+
+def assert_exception(session, query, matching=None, expected=None):
```
mambocab: With the changes described above for `__assert_exception`, it might not be necessary to wrap it anymore; you might be able to consolidate all the logic into a single method.

read_repair_test.py
```diff
@@ -47,7 +47,7 @@ def alter_rf_and_run_read_repair_test(self):
             debug("Checking " + n.name)
             session = self.patient_exclusive_cql_connection(n)
             res = rows_to_list(session.execute(cl_one_stmt))
-            assert res == [[1, 1, 1]], res
+            self.assertEqual(res, [[1, 1, 1]], res)
```
ptnapoleon: Use assert_one here

paging_test.py
```diff
@@ -516,7 +516,7 @@ def handle_scenario(scenario):
                 )
             else:
                 # this should not happen
-                assert False
+                self.assertTrue(False, "This should not happen.")
```
ptnapoleon: @mambocab Should we self.fail() here, or do it like this?

native_transport_ssl_test.py
```diff
@@ -28,8 +28,7 @@ def connect_to_ssl_test(self):
         except NoHostAvailable:
             pass
 
-        assert len(node1.grep_log("^io.netty.handler.ssl.NotSslRecordException.*")) > 0, \
-            "Missing SSL handshake exception while connecting with non-SSL enabled client"
+        Self.assertGreater(len(node1.grep_log("^io.netty.handler.ssl.NotSslRecordException.*")), 0, "Missing SSL handshake exception while connecting with non-SSL enabled client")
```
ptnapoleon: nit: `Self` is a typo, also it seems the spacing is wrong.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -1805,7 +1805,7 @@ def do_round_trip(trueval, falseval, invalid=False):
                                  .format(tempfile.name, trueval, falseval))
 
             imported_results = list(self.session.execute("SELECT * FROM testbooleans"))
-            assert len(imported_results) == 2
+            self.assertEqual(len(imported_results), 2)
```
ptnapoleon: For this, and the change above, could you print `imported_results` on failure?

configuration_test.py
```diff
@@ -128,13 +128,10 @@ def _check_chunk_length(self, session, value):
             if 'compression' in result:
                 params = result
 
-        assert params is not '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result)
-
+        self.assertEqual(params, '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result))
```
ptnapoleon: I think you inverted this condition.

paging_test.py
mambocab: I vote `self.fail('This should not happen.')`, maybe with a more descriptive message, like `invalid scenario configuration` or something.

ptnapoleon: Those are equally unhelpful to me. Maybe "Invalid configuration, this should never happen, please go into the code"

paxos_tests.py
```diff
@@ -189,4 +189,4 @@ def run(self):
             errors = errors + w.errors
             retries = retries + w.retries
 
-        assert (value == threads * iterations) and (errors == 0), "value=%d, errors=%d, retries=%d" % (value, errors, retries)
+        self.assertTrue( (value == threads * iterations) and (errors == 0), "value={}, errors={}, retries={}".format(value, errors, retries))
```
mambocab: unnecessary leading space after the `(`.

mambocab:
Might be nice, in a separate PR, to add a `msg` parameter to these assertion methods (unless there already is one and I've missed it.

bootstrap_test.py
```diff
@@ -188,9 +188,8 @@ def resumable_bootstrap_test(self):
                                        'replication(factor=2)', '-rate', 'threads=8'],
                                       capture_output=True)
 
-        if stdout and "FAILURE" in stdout:
-            debug(stdout)
-            assert False, "Cannot read inserted data after bootstrap"
+        if stdout:
```
mambocab: Don't think it's necessary to put this in an `if` block:

```sh
❯ python -c "print 'FAILURE' in ''"
False
```

configuration_test.py
mambocab: Yes, and please use `assertNotEqual` -- the old assertion should have used `!=`, not `is not`.

counter_tests.py
```diff
@@ -38,10 +38,10 @@ def simple_increment_test(self):
             query = SimpleStatement("SELECT key, c FROM cf WHERE key IN (%s)" % keys, consistency_level=ConsistencyLevel.QUORUM)
             res = list(session.execute(query))
 
-            assert len(res) == nb_counter
+            self.assertEqual(len(res), nb_counter)
```
mambocab: In this and other `assertEqual(len(x), ...)` calls, could you please pass in `x` as a message parameter? That'll give us better debug output.

Alternatively, maybe we can introduce an `assert_length_equals` assertion that prints what we want on failure.

paging_test.py
```diff
@@ -176,7 +176,7 @@ def assertEqualIgnoreOrder(self, actual, expected):
         return self.assertItemsEqual(actual, expected)
 
     def assertIsSubsetOf(self, subset, superset):
-        assert flatten_into_set(subset).issubset(flatten_into_set(superset))
+        self.assertTrue(flatten_into_set(subset).issubset(flatten_into_set(superset)))
```
mambocab: Might be a more invasive change than we want to make here, but this should be logically equivalent to:

```python
self.assertLessEqual(flatten_into_set(subset), flatten_into_set(superset))
```

This version will have better debugging output. I'm making this assertion on the basis of this behavior:

```python
>>> {1, 2, 3} <= {1, 2, 3, 4}
True
>>> {1, 2, 3, 4} <= {1, 2, 3, 4}
True
```

assertions.py
```diff
@@ -7,79 +7,168 @@
 
 from tools import rows_to_list
 
+"""
+Here are examples of the following methods found in this file:
```
ptnapoleon: So, rather than just copying examples of callsites, which are of medium help, I was hoping you could provide prescriptive documentation

mambocab: You might give a high-level explanation in this module docstring, then provide examples in method docstrings. Just a thought.

bootstrap_test.py
ptnapoleon: Depends on if `stdout`'s falsey condition is empty string, or if it's None

compaction_test.py
```diff
@@ -320,7 +320,7 @@ def large_compaction_warning_test(self):
             session.execute("UPDATE ks.large SET properties[%i] = '%s' WHERE userid = 'user'" % (i, get_random_word(strlen)))
 
         ret = list(session.execute("SELECT properties from ks.large where userid = 'user'"))
-        assert len(ret) == 1
+        self.assertEqual(len(ret),1, "Expected length of response is 1.  Got {}.".format(len(ret)))
```
ptnapoleon: Message should just be `ret`

compaction_test.py
```diff
@@ -332,7 +332,7 @@ def large_compaction_warning_test(self):
 
         ret = list(session.execute("SELECT properties from ks.large where userid = 'user'"))
 
-        assert len(ret) == 1
+        self.assertEqual(len(ret),1, "Expected length of response is 1.  Got {}.".format(len(ret)))
```
ptnapoleon: Message should just be `ret`

cqlsh_tests/cqlsh_copy_tests.py
mambocab: We're on the same wavelength:

https://github.com/riptano/cassandra-dtest/pull/1012/files/c46d8833ae6548656975102ba2f8e62fe378e50f#r65405422

What do you think of an `assert_length_equal` method?

configuration_test.py
```diff
@@ -128,13 +128,10 @@ def _check_chunk_length(self, session, value):
             if 'compression' in result:
                 params = result
 
-        assert params is not '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result)
-
+        self.assertNotEqual(params, '', "Looking for the string 'sstable_compression', but could not find it in {str}".format(str=result))
```
ptnapoleon: @mambocab Just checking my python, this is functionally equivalent to what it used to be, yes?

configuration_test.py
```diff
@@ -128,13 +128,10 @@ def _check_chunk_length(self, session, value):
...
         chunk_string = "chunk_length_kb" if self.cluster.version() < '3.0' else "chunk_length_in_kb"
         chunk_length = int(re.search("{chunk}.*?:.*?'(\d*?)'".format(chunk=chunk_string), result).groups()[0])
-
-        assert chunk_length == value, "Expected chunk_length: %s.  We got: %s" % (value, chunk_length)
-
+        self.assertEqual(chunk_length, value, "Expected chunk_length: {}.  We got: {}".format(value, chunk_length))
```
ptnapoleon: You can remove the error message, it's the same as the default

paging_test.py
mambocab: What failure doesn't imply `go into the code`?

Might be helpful to print the value of `scenario` as well.

consistency_test.py
```diff
@@ -537,7 +536,8 @@ def run():
             err_type, err, err_traceback = exceptions_queue.get()
             debug("Failed with exception {}: {}".format(err_type, err.message))
             traceback.print_exception(err_type, err, err_traceback)
-            assert False, err.message
+            debug(err.message)
+            self.fail()
```
ptnapoleon: This can just be `self.fail(err.message)`

bootstrap_test.py
mambocab: Ah -- can this condition be `if stdout is not None`, then, to communicate that?

consistency_test.py
```diff
@@ -850,7 +850,7 @@ def short_read_quorum_delete_test(self):
         # we read the first row in the partition (so with a LIMIT 1) and A and B answer first.
         node3.flush()
         node3.stop(wait_other_notice=True)
-        assert_none(session, "SELECT * FROM t WHERE id = 0 LIMIT 1", cl=ConsistencyLevel.QUORUM)
+        self.assert_none(session, "SELECT * FROM t WHERE id = 0 LIMIT 1", cl=ConsistencyLevel.QUORUM)
```
ptnapoleon: I don't think this `self.` should be here

consistency_test.py
```diff
@@ -921,10 +921,10 @@ def short_read_reversed_test(self):
             query = SimpleStatement('SELECT c, v FROM cf WHERE key=\'k0\' ORDER BY c DESC LIMIT 3', consistency_level=ConsistencyLevel.QUORUM)
             rows = list(session.execute(query))
             res = rows
-            assert len(res) == 3, 'Expecting 3 values, got %d (%s)' % (len(res), str(res))
+            self.assertEqual(len(res), 3, 'Expecting 3 values, got {} ({})'.format(len(res), str(res)))
             # value 6, 7 and 8 have been deleted
             for i in xrange(0, 3):
-                assert res[i][1] == 'value%d' % (5 - i), 'Expecting value%d, got %s (%s)' % (5 - i, res[i][1], str(res))
+                self.assertEqual(res[i][1],'value%d' % (5 - i),  'Expecting value {}, got {} ({})'.format(5 - i, res[i][1], str(res)))
```
ptnapoleon: Make this string interpolation into a `.format` too, please

batch_test.py
```diff
@@ -190,8 +185,7 @@ def logged_batch_doesnt_throw_uae_test(self):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner')
             APPLY BATCH
         """, consistency_level=ConsistencyLevel.ANY)
-        session.execute(query)
-        assert True
+        assert_all(session, "SELECT * FROM users", [[0, u'Jack', u'Sparrow'], [1, u'Will', u'Turner']], ignore_order=True)
```
ptnapoleon: Can you remove the `ignore_order=True` on all of these, and just assert them in the order they come in? It should never change.

configuration_test.py
mambocab: No, but it is correct. `'' is ''` is (usually? always?) true in CPython, but I don't believe it's guaranteed by Python's semantics.

counter_tests.py
ptnapoleon: I am forced to agree with Jim here, that's a very good idea to have `assert_length_equals`, rather than constantly correcting PRs

delete_insert_test.py
```diff
@@ -68,7 +68,7 @@ def run(self):
                 session = self.connection
                 query = SimpleStatement("SELECT * FROM delete_insert_search_test.test WHERE group = 'group2'", consistency_level=ConsistencyLevel.LOCAL_QUORUM)
                 rows = session.execute(query)
-                assert len(list(rows)) == len(deleted)
+                self.assertEqual(len(list(rows)), len(deleted))
```
ptnapoleon: This should probably print out more helpful info on `rows` and `deleted`

largecolumn_test.py
```diff
@@ -25,11 +25,11 @@ def is_number(s):
         output = node.nodetool("gcstats", capture_output=True)
         debug(output)
         output = output[0].split("\n")
-        assert output[0].strip().startswith("Interval"), "Expected output from nodetool gcstats starts with a header line with first column Interval"
+        self.assertTrue(output[0].strip().startswith("Interval"), "Expected output from nodetool gcstats starts with a header line with first column Interval")
```
ptnapoleon: Can you change the message to print out `output[0]`? And @mambocab, is it wrong of me to want to use `self.assertRegexpMatches` instead of `self.assertTrue($STRING_CONDITION)`?

largecolumn_test.py
```diff
@@ -25,11 +25,11 @@ def is_number(s):
...
         fields = output[1].split()
-        assert len(fields) >= 6, "Expected output from nodetool gcstats has at least six fields"
+        self.assertGreaterEqual(len(fields), 6, "Expected output from nodetool gcstats has at least six fields")
         for field in fields:
-            assert is_number(field.strip()) or field == 'NaN', "Expected numeric from fields from nodetool gcstats"
+            self.assertTrue(is_number(field.strip()) or field == 'NaN', "Expected numeric from fields from nodetool gcstats")
```
ptnapoleon: I would like to know more in this error message, and the one above.

paging_test.py
ptnapoleon: > A set is less than another set if and only if the first set is a proper subset of the second set (is a subset, but is not equal)

From https://docs.python.org/2/library/sets.html  So you are correct

paxos_tests.py
ptnapoleon: autopep8 will handle it

range_ghost_test.py
```diff
@@ -22,7 +22,7 @@ def ghosts_test(self):
             session.execute("UPDATE cf SET c = 'value' WHERE key = 'k%i'" % i)
 
         res = list(session.execute("SELECT * FROM cf LIMIT 10000"))
-        assert len(res) == rows, res
+        self.assertEqual(len(res), rows)
```
ptnapoleon: I'm okay not printing out `res` here, because it is 1000 rows long

read_repair_test.py
```diff
@@ -62,35 +63,32 @@ def alter_rf_and_run_read_repair_test(self):
         read_repair_session = self.patient_exclusive_cql_connection(non_replicas[0])
         res = read_repair_session.execute(cl_all_stmt)
         # result of the CL ALL query contains only the selected column
-        assert rows_to_list(res) == [[1]], res
+        self.assertEqual(rows_to_list(res), [[1]], res)
```
ptnapoleon: Use assert_one here

read_repair_test.py
```diff
@@ -62,35 +63,32 @@ def alter_rf_and_run_read_repair_test(self):
...
 
         # Check the results of the read repair by querying each replica again at CL ONE
         debug("Re-running SELECTs at CL ONE to verify read repair")
         for n in self.cluster.nodelist():
             debug("Checking " + n.name)
             session = self.patient_exclusive_cql_connection(n)
             res = rows_to_list(session.execute(cl_one_stmt))
-            assert len(res) == 1
+            self.assertEqual(len(res), 1)
```
ptnapoleon: Print `res` please

read_repair_test.py
```diff
@@ -62,35 +63,32 @@ def alter_rf_and_run_read_repair_test(self):
...
             # Column a must be at 1 everywhere
-            assert res[0][1] == 1, res
+            self.assertEqual(res[0][1], 1, res)
             # Column b must be either 1 or None everywhere
-            assert res[0][2] == 1 or res[0][2] is None, res
+            self.assertTrue(res[0][2] == 1 or res[0][2] is None, res)
```
ptnapoleon: I don't feel like these comments match the assertions, please fix one of them

read_repair_test.py
```diff
@@ -62,35 +63,32 @@ def alter_rf_and_run_read_repair_test(self):
...
 
         # Now query at ALL but selecting all columns
-        cl_all_stmt = SimpleStatement("SELECT * FROM alter_rf_test.t1 WHERE k=1",
-                                      consistency_level=ConsistencyLevel.ALL)
+        query = "SELECT * FROM alter_rf_test.t1 WHERE k=1"
         debug("Executing SELECT on non-initial replica to trigger read repair " + non_replicas[0].name)
         read_repair_session = self.patient_exclusive_cql_connection(non_replicas[0])
-        res = read_repair_session.execute(cl_all_stmt)
-        assert rows_to_list(res) == [[1, 1, 1]], res
+        assert_all(session, query, [[1, 1, 1]], cl=ConsistencyLevel.ALL)
```
ptnapoleon: This can be assert_one

largecolumn_test.py
mambocab: Whatever's more readable and gives good error messages. In this case, I think the sweet spot is probably

```python
self.assertRegexpMatches(output[0].strip(), '^Interval')
```

which removes the need for giving an explicit `msg` argument.

(side note: oh man, [`assertRegexpMatches` is deprecated](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L1323)? so maybe `assertRegex`, if that's been backported to 2.7.)

(side side note: oh hell yes, [`assertRegex` has a `msg` param](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L1279))

range_ghost_test.py
mambocab: `maxDiff` should take care of that if implement an `assert_length_equals` method on top of a `unittest` assert.

wide_rows_test.py
```diff
@@ -102,4 +102,4 @@ def test_column_index_stress(self):
             rows = list(session.execute(select_column_query.format(name1="val" + values2fetch[0],
                                                                    name2="val" + values2fetch[1],
                                                                    name3="val" + values2fetch[2])))
-            assert len(rows) == expected_rows
+            self.assertEqual(len(rows), expected_rows)
```
ptnapoleon: Print rows please

user_types_test.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, *args, **kwargs):
     def assertUnauthorized(self, session, query, message):
         with self.assertRaises(Unauthorized) as cm:
             session.execute(query)
-        assert re.search(message, cm.exception.message), "Expected: %s" % message
+        self.assertTrue(re.search(message, cm.exception.message))
```
ptnapoleon: Why not use `self.assertRegexpMatches`?

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -660,7 +660,7 @@ def _check_counters(self):
                     # counter wasn't found
                     actual_value = None
 
-                assert actual_value == expected_value, "Counter not at expected value. Got %s, expected %s" % (actual_value, expected_value)
+                self.assertEqual(actual_value, expected_value, "Counter not at expected value. Got {}, expected {}".format(actual_value, expected_value))
```
ptnapoleon: Remove the error message, it's redundant

upgrade_tests/cql_tests.py
```diff
@@ -3756,11 +3727,11 @@ def nan_infinity_test(self):
 
             # selected should be [[nan], [inf], [-inf]],
             # but assert element-wise because NaN != NaN
-            assert len(selected) == 3
-            assert len(selected[0]) == 1
-            assert math.isnan(selected[0][0])
-            assert selected[1] == [float("inf")]
-            assert selected[2] == [float("-inf")]
+            self.assertEqual(len(selected), 3)
+            self.assertEqual(len(selected[0]), 1)
```
ptnapoleon: Print `selected` here and above

upgrade_tests/cql_tests.py
```diff
@@ -3756,11 +3727,11 @@ def nan_infinity_test(self):
...
+            self.assertTrue(math.isnan(selected[0][0]))
+            self.assertEqual(selected[1], [float("inf")])
+            self.assertEqual(selected[2], [float("-inf")])
```
ptnapoleon: And in these too, please

upgrade_tests/cql_tests.py
```diff
@@ -2612,21 +2589,16 @@ def timeuuid_test(self):
...
-            assert len(res) == 0, res
+            assert_row_count(cursor, test, 0, where="k = 0 AND t < {}".format(dates[0]))
 
-            res = list(cursor.execute("SELECT * FROM test WHERE k = 0 AND t > %s AND t <= %s" % (dates[0], dates[2])))
-            assert len(res) == 2, res
+            assert_row_count(cursor, test, 2, where="k = 0 AND t > {} AND t <= {}".format(dates[0], dates[2]))
 
-            res = list(cursor.execute("SELECT * FROM test WHERE k = 0 AND t = %s" % dates[0]))
-            assert len(res) == 1, res
+            assert_row_count(cursor, test, 1, where="k = 0 AND t = {}".format(dates[0]))
```
ptnapoleon: @mambocab Bikeshed with me. Should we extend `assert_row_count` the way Sean has suggested, or should we keep the queries as is, but use the new `assert_length_equal`? I prefer the latter

ptnapoleon: It also seems like all of these, and the ones above should just assert the rows, rather than the length, no?

upgrade_tests/cql_tests.py
```diff
@@ -1527,7 +1522,7 @@ def range_query_test(self):
             cursor.execute("INSERT INTO test (a, b, c, d, e, f) VALUES (1, 1, 1, 1, 5, '5');")
 
             res = cursor.execute("SELECT a, b, c, d, e, f FROM test WHERE a = 1 AND b = 1 AND c = 1 AND d = 1 AND e >= 2;")
-            assert rows_to_list(res) == [[1, 1, 1, 1, 2, u'2'], [1, 1, 1, 1, 3, u'3'], [1, 1, 1, 1, 5, u'5']], res
+            self.assertEqual(rows_to_list(res), [[1, 1, 1, 1, 2, u'2'], [1, 1, 1, 1, 3, u'3'], [1, 1, 1, 1, 5, u'5']])
```
ptnapoleon: Use assert_all

upgrade_tests/cql_tests.py
```diff
@@ -1416,23 +1411,23 @@ def map_test(self):
...
             # Check we correctly overwrite
             cursor.execute(q % "m = { 'm' : 4 , 'n' : 1, 'o' : 2 }")
             res = cursor.execute("SELECT m FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[{'m': 4, 'n': 1, 'o': 2}]], res
+            self.assertEqual(rows_to_list(res), [[{'m': 4, 'n': 1, 'o': 2}]])
 
             cursor.execute(q % "m = {}")
             res = cursor.execute("SELECT m FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [], res
+            self.assertEqual(rows_to_list(res), [])
```
ptnapoleon: Use assert_none

upgrade_tests/cql_tests.py
```diff
@@ -1416,23 +1411,23 @@ def map_test(self):
...
-            assert rows_to_list(res) == [[{'a': 4, 'b': 2, 'c': 3}]], res
+            self.assertEqual(rows_to_list(res), [[{'a': 4, 'b': 2, 'c': 3}]])
 
             time.sleep(.01)
 
             # Check we correctly overwrite
             cursor.execute(q % "m = { 'm' : 4 , 'n' : 1, 'o' : 2 }")
             res = cursor.execute("SELECT m FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[{'m': 4, 'n': 1, 'o': 2}]], res
+            self.assertEqual(rows_to_list(res), [[{'m': 4, 'n': 1, 'o': 2}]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1416,23 +1411,23 @@ def map_test(self):
...
 
             res = cursor.execute("SELECT m FROM user")
-            assert rows_to_list(res) == [[{'woot': 5, 'bar': 6}]], res
+            self.assertEqual(rows_to_list(res), [[{'woot': 5, 'bar': 6}]])
 
             q = "UPDATE user SET %s WHERE fn='Bilbo' AND ln='Baggins'"
             cursor.execute(q % "m = { 'a' : 4 , 'c' : 3, 'b' : 2 }")
             res = cursor.execute("SELECT m FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[{'a': 4, 'b': 2, 'c': 3}]], res
+            self.assertEqual(rows_to_list(res), [[{'a': 4, 'b': 2, 'c': 3}]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1416,23 +1411,23 @@ def map_test(self):
             cursor.execute("DELETE m['foo'] FROM user WHERE fn='Tom' AND ln='Bombadil'")
 
             res = cursor.execute("SELECT m FROM user")
-            assert rows_to_list(res) == [[{'woot': 5, 'bar': 6}]], res
+            self.assertEqual(rows_to_list(res), [[{'woot': 5, 'bar': 6}]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1371,26 +1366,26 @@ def set_test(self):
...
 
             cursor.execute("DELETE tags['m'] FROM user WHERE fn='Bilbo' AND ln='Baggins'")
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[set(['n'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['n'])]])
 
             cursor.execute("DELETE tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [], res
+            self.assertEqual(rows_to_list(res), [])
```
ptnapoleon: assert_none

upgrade_tests/cql_tests.py
```diff
@@ -1371,26 +1366,26 @@ def set_test(self):
...
 
             cursor.execute(q % "tags = { 'm', 'n' }")
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[set(['m', 'n'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['m', 'n'])]])
 
             cursor.execute("DELETE tags['m'] FROM user WHERE fn='Bilbo' AND ln='Baggins'")
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[set(['n'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['n'])]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1371,26 +1366,26 @@ def set_test(self):
...
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[set(['a', 'b', 'c'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['a', 'b', 'c'])]])
 
             time.sleep(.01)
 
             cursor.execute(q % "tags = { 'm', 'n' }")
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[set(['m', 'n'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['m', 'n'])]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1371,26 +1366,26 @@ def set_test(self):
...
-            assert rows_to_list(res) == [[set(['foo', 'foobar'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['foo', 'foobar'])]])
 
-            q = "UPDATE user SET %s WHERE fn='Bilbo' AND ln='Baggins'"
-            cursor.execute(q % "tags = { 'a', 'c', 'b' }")
+            q = "UPDATE user SET {} WHERE fn='Bilbo' AND ln='Baggins'"
+            cursor.execute(q.format("tags = { 'a', 'c', 'b' }"))
             res = cursor.execute("SELECT tags FROM user WHERE fn='Bilbo' AND ln='Baggins'")
-            assert rows_to_list(res) == [[set(['a', 'b', 'c'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['a', 'b', 'c'])]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1371,26 +1366,26 @@ def set_test(self):
             cursor.execute(q % "tags = tags - { 'bar' }")
 
             res = cursor.execute("SELECT tags FROM user")
-            assert rows_to_list(res) == [[set(['foo', 'foobar'])]], res
+            self.assertEqual(rows_to_list(res), [[set(['foo', 'foobar'])]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1345,7 +1340,7 @@ def range_query_2ndary_test(self):
 
             assert_invalid(cursor, "SELECT * FROM indextest WHERE setid = 0 AND row < 1;")
             res = cursor.execute("SELECT * FROM indextest WHERE setid = 0 AND row < 1 ALLOW FILTERING;")
-            assert rows_to_list(res) == [[0, 0, 0]], res
+            self.assertEqual(rows_to_list(res), [[0, 0, 0]])
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -1295,7 +1290,7 @@ def range_tombstones_compaction_test(self):
             self.cluster.compact()
 
             res = cursor.execute("SELECT v1 FROM test1 WHERE k = 0")
-            assert rows_to_list(res) == [['%i%i' % (c1, c2)] for c1 in xrange(0, 4) for c2 in xrange(0, 2) if c1 != 1], res
+            self.assertEqual(rows_to_list(res), [['{}{}'.format(c1, c2)] for c1 in xrange(0, 4) for c2 in xrange(0, 2) if c1 != 1])
```
ptnapoleon: assert_all

upgrade_tests/cql_tests.py
```diff
@@ -1245,25 +1240,25 @@ def range_tombstones_test(self):
...
-                assert rows_to_list(res) == [[x, x] for x in xrange(i * cpr + col1, (i + 1) * cpr)], res
+                self.assertEqual(rows_to_list(res), [[x, x] for x in xrange(i * cpr + col1, (i + 1) * cpr)])
 
             self.cluster.flush()
             time.sleep(0.2)
 
             for i in xrange(0, rows):
                 res = cursor.execute("SELECT v1, v2 FROM test1 WHERE k = %d" % i)
-                assert rows_to_list(res) == [[x, x] for x in xrange(i * cpr + col1, (i + 1) * cpr)], res
+                self.assertEqual(rows_to_list(res), [[x, x] for x in xrange(i * cpr + col1, (i + 1) * cpr)])
```
ptnapoleon: assert_all

upgrade_tests/cql_tests.py
```diff
@@ -1245,25 +1240,25 @@ def range_tombstones_test(self):
...
-                assert rows_to_list(res) == [[x, x] for x in xrange(i * cpr, (i + 1) * cpr)], res
+                self.assertEqual(rows_to_list(res), [[x, x] for x in xrange(i * cpr, (i + 1) * cpr)])
 
             for i in xrange(0, rows):
                 cursor.execute("DELETE FROM test1 WHERE k = %d AND c1 = 0" % i)
 
             for i in xrange(0, rows):
                 res = cursor.execute("SELECT v1, v2 FROM test1 WHERE k = %d" % i)
-                assert rows_to_list(res) == [[x, x] for x in xrange(i * cpr + col1, (i + 1) * cpr)], res
+                self.assertEqual(rows_to_list(res), [[x, x] for x in xrange(i * cpr + col1, (i + 1) * cpr)])
```
ptnapoleon: assert_all

upgrade_tests/cql_tests.py
```diff
@@ -1245,25 +1240,25 @@ def range_tombstones_test(self):
                 for j in xrange(0, col1):
                     for k in xrange(0, col2):
                         n = (i * cpr) + (j * col2) + k
-                        cursor.execute("INSERT INTO test1 (k, c1, c2, v1, v2) VALUES (%d, %d, %d, %d, %d)" % (i, j, k, n, n))
+                        cursor.execute("INSERT INTO test1 (k, c1, c2, v1, v2) VALUES ({}, {}, {}, {}, {})".format(i, j, k, n, n))
 
             for i in xrange(0, rows):
                 res = cursor.execute("SELECT v1, v2 FROM test1 where k = %d" % i)
-                assert rows_to_list(res) == [[x, x] for x in xrange(i * cpr, (i + 1) * cpr)], res
+                self.assertEqual(rows_to_list(res), [[x, x] for x in xrange(i * cpr, (i + 1) * cpr)])
```
ptnapoleon: assert_all

upgrade_tests/cql_tests.py
```diff
@@ -53,16 +54,14 @@ def static_cf_test(self):
 
             # Queries
             res = cursor.execute("SELECT firstname, lastname FROM users WHERE userid = 550e8400-e29b-41d4-a716-446655440000")
-            assert rows_to_list(res) == [['Frodo', 'Baggins']], res
+            self.assertEqual(rows_to_list(res), [['Frodo', 'Baggins']], res)
```
ptnapoleon: assert_one

upgrade_tests/cql_tests.py
```diff
@@ -53,16 +54,14 @@ def static_cf_test(self):
...
 
             res = cursor.execute("SELECT * FROM users WHERE userid = 550e8400-e29b-41d4-a716-446655440000")
-            assert rows_to_list(res) == [[UUID('550e8400-e29b-41d4-a716-446655440000'), 32, 'Frodo', 'Baggins']], res
+            self.assertEqual(rows_to_list(res), [[UUID('550e8400-e29b-41d4-a716-446655440000'), 32, 'Frodo', 'Baggins']], res)
```
ptnapoleon: assert_one

ptnapoleon:
Rather the commenting 100x more times, can you just use assert_one and assert_all everywhere they should be used in upgrade_test/cql_tests

ptnapoleon:
First pass of review probably done

ptnapoleon:
Also you need to rebase, and placate Travis

upgrade_tests/cql_tests.py
mambocab: > should just assert the rows

Do you mean we should use an equality comparison? If so, +1, good idea

paging_test.py
ptnapoleon: Some failures, but sure, prune that and print out the value of scenario

cqlsh_tests/cqlsh_copy_tests.py
ptnapoleon: It's being written as we speak

bootstrap_test.py
ptnapoleon: Yes

ptnapoleon:
Sean, I found a bug in the assert_X functions. They should not default to Cl.ONE, because we risk overriding a default consistency level set on that session

upgrade_tests/cql_tests.py
mambocab: > Should we extend `assert_row_count` the way Sean has suggested

The change to `assert_row_count` seems useful and doesn't break its API, so I see no reason not to add the feature.

> should we keep the queries as is, but use the new `assert_length_equal`?

Surprisingly enough, no strong preference from me. As is, though, `assert_length_equal` is slightly more clearly-named than `assert_row_count`, whose name obscures that it `SELECT *`s over a table. That makes me prefer the `assert_length_equal` formulation a little.

Though maybe calling it as `assert_row_count(cursor, table_name=test, where=...)` would make the behavior a little clearer? I dunno. I'm spitballing at this point.

range_ghost_test.py
mambocab: Then I guessssss `msg=str(rows)[:1000]` or something?

ptnapoleon: My thought is, it probably really doesn't matter which rows were dropped at this scale, but it's just disk space, so I'm fine either way

mambocab: I agree, up to y'all.

consistency_test.py
```diff
@@ -537,7 +536,7 @@ def run():
             err_type, err, err_traceback = exceptions_queue.get()
             debug("Failed with exception {}: {}".format(err_type, err.message))
             traceback.print_exception(err_type, err, err_traceback)
-            assert False, err.message
+            self.fail(err.message)
```
mambocab: I think this whole block can be replaced with

```python
if not exceptions_queue.is_empty():
    raise exceptions_queue.get()
```

Which should have the same behavior. I'll file a ticket to convert this to use `dtest.MultiError` though.

ptnapoleon:
Running the upgrade changes here http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/128/

batch_test.py
```diff
@@ -60,9 +59,8 @@ def logged_batch_accepts_regular_mutations_test(self):
             INSERT INTO users (id, firstname, lastname) VALUES (1, 'Will', 'Turner')
             APPLY BATCH
         """)
-        rows = session.execute("SELECT * FROM users")
-        res = sorted(rows)
-        assert [list(res[0]), list(res[1])] == [[0, u'Jack', u'Sparrow'], [1, u'Will', u'Turner']], res
+        assert_all(session, "SELECT * FROM users", [[1, u'Will', u'Turner'], [0, u'Jack', u'Sparrow']])
+        #assert_all(session, "SELECT * FROM users", [[0, u'Jack', u'Sparrow'], [1, u'Will', u'Turner']], ignore_order=True)
```
ptnapoleon: Remove this commented out line, and all the others like it in this file

compaction_test.py
```diff
@@ -320,7 +320,7 @@ def large_compaction_warning_test(self):
             session.execute("UPDATE ks.large SET properties[%i] = '%s' WHERE userid = 'user'" % (i, get_random_word(strlen)))
 
         ret = list(session.execute("SELECT properties from ks.large where userid = 'user'"))
-        assert len(ret) == 1
+        self.assertEqual(len(ret), 1, "Expected length of response is 1.  Got {}.".format(len(ret)))
```
ptnapoleon: Do assert_length_equal here

compaction_test.py
```diff
@@ -332,7 +332,7 @@ def large_compaction_warning_test(self):
 
         ret = list(session.execute("SELECT properties from ks.large where userid = 'user'"))
 
-        assert len(ret) == 1
+        self.assertEqual(len(ret), 1, "Expected length of response is 1.  Got {}.".format(len(ret)))
```
ptnapoleon: Do assert_length_equal here

consistency_test.py
```diff
@@ -769,10 +768,10 @@ def short_read_test(self):
             query = SimpleStatement('SELECT c, v FROM cf WHERE key=\'k0\' LIMIT 3', consistency_level=ConsistencyLevel.QUORUM)
             rows = list(session.execute(query))
             res = rows
-            assert len(res) == 3, 'Expecting 3 values, got %d (%s)' % (len(res), str(res))
+            self.assertEqual(len(res), 3, 'Expecting 3 values, got {} ({})'.format(len(res), str(res)))
```
ptnapoleon: Do assert_length_equal here

consistency_test.py
```diff
@@ -809,7 +808,7 @@ def short_read_delete_test(self):
 
         query = SimpleStatement('SELECT c, v FROM cf WHERE key=\'k0\' LIMIT 1', consistency_level=ConsistencyLevel.QUORUM)
         res = list(session.execute(query))
-        assert len(res) == 0, res
+        self.assertEqual(len(res), 0, res)
```
ptnapoleon: Do assert_length_equal here

consistency_test.py
```diff
@@ -921,10 +920,10 @@ def short_read_reversed_test(self):
             query = SimpleStatement('SELECT c, v FROM cf WHERE key=\'k0\' ORDER BY c DESC LIMIT 3', consistency_level=ConsistencyLevel.QUORUM)
             rows = list(session.execute(query))
             res = rows
-            assert len(res) == 3, 'Expecting 3 values, got %d (%s)' % (len(res), str(res))
+            self.assertEqual(len(res), 3, 'Expecting 3 values, got {} ({})'.format(len(res), str(res)))
```
ptnapoleon: Do assert_length_equal here

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -20,6 +20,7 @@
 from cassandra.util import SortedSet
 from ccmlib.common import is_win
 
+from assertionsn import assert_length_equal
```
ptnapoleon: Typo here

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -1763,7 +1765,7 @@ def _test(prepared_statements):
 
             imported_results = list(self.session.execute("SELECT * FROM testdatatype"))
 
-            assert len(imported_results) == 1
+            self.assertEqual(len(imported_results), 1)
```
ptnapoleon: Do assert_length_equal here

tools.py
```diff
@@ -80,11 +81,11 @@ def query_c1c2(session, key, consistency=ConsistencyLevel.QUORUM, tolerate_missi
     query = SimpleStatement('SELECT c1, c2 FROM cf WHERE key=\'k%d\'' % key, consistency_level=consistency)
     rows = list(session.execute(query))
     if not tolerate_missing:
-        assert len(rows) == 1
+        assert_equal(len(rows), 1)
```
ptnapoleon: We should print out rows, and maybe use assert_length_equal here

read_repair_test.py
smccarthy788: I think the comment reflects table values in each node, which the assertions are correctly checking.

tools.py
```diff
@@ -11,11 +11,13 @@
 from distutils.version import LooseVersion
 from threading import Thread
 
+from assertions import assert_length_equal
```
ptnapoleon: @mambocab I'm having a problem, where when a file imports assert_length_equal and then imports tools.py, an error is thrown

mambocab: AFK at the moment, but what happens on `python -c 'from assertions import assert_length_equal ; import tools'`

ptnapoleon:
```
python -c 'from assertions import assert_length_equal ; import tools'
  File "<string>", line 1, in <module>
  File "assertions.py", line 8, in <module>
    from tools import rows_to_list
  File "tools.py", line 14, in <module>
    from assertions import assert_length_equal
ImportError: cannot import name assert_length_equal
```

ptnapoleon:
So @smccarthy788, we'll need to run this again on cassci, to make sure my huge rebase didn't break anything, and @mambocab if you could re-review it with me, that would be great. I did turn 66 interleaved commits into 3 sequential ones.

smccarthy788:
http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/13/

ptnapoleon:
Dtest run reveals issues, @c-kodman can you fix?

meta_tests/assertion_test.py
```diff
@@ -0,0 +1,98 @@
+from cassandra import AlreadyExists, ConsistencyLevel
+from dtest import Tester
+from assertions import (assert_one, assert_none, assert_exception, assert_unavailable,
+                        assert_invalid, assert_unauthorized, assert_all, assert_almost_equal,
+                        assert_length_equal, assert_row_count)
+from cassandra.query import SimpleStatement
+
+
+class TestAssertionMethods(Tester):
```
mambocab: I'm so excited about these meta tests. I'm not certain that they need to spin up a cluster at all though. I can see that it's nice to have some confidence that they work with C\* itself. However, as is, we have to structure them in a non-standard way to avoid tearing down and restarting clusters over and over. That's a cost, and the extra runtime is another cost. I don't know those costs are worth it; I think we get almost as much confidence from mocking out the session out. I'm thinking something like

```python
>>> from mock import Mock
>>> from assertions import assert_unavailable
>>> from cassandra import Unavailable
>>>
>>> mock_session = Mock(**{'execute.side_effect': Unavailable('dummy unavailable message')})
>>> assert_unavailable(mock_session.execute, 'dummy string')
>>>
```

If we migrate all these tests to this style, this class will no longer need to inherit from `Tester`, which I think is a small win.

ptnapoleon: +1

meta_tests/assertion_test.py
```diff
@@ -1,98 +1,54 @@
...
-
-        assert_almost_equal_test(session)
-        self.clean_cluster(session)
-
-        assert_row_count_test(session)
-        self.clean_cluster(session)
-
-        assert_length_equal_test(session)
+        #assert_exception_test
+        mock_session = Mock(**{'execute.side_effect': AlreadyExists("Dummy exception message.")})
```
mambocab: Sweet, looking good. Now that these are normal "setup->assert" tests, you can move each to its own test method.

It'd also be good to spend some time on making sure changing input values change behavior and doing more thorough unit-style tests. For instance, `assert_exception(mock_session, "DUMMY QUERY", expected=AlreadyExists)` should fail if `mock_session.execute` raises a `KeyError`, and should also fail if `mock_session.execute` doesn't raise any error.

meta_tests/assertion_test.py
```diff
@@ -1,98 +1,54 @@
-from cassandra import AlreadyExists, ConsistencyLevel
+from cassandra import AlreadyExists, ConsistencyLevel, Unavailable, InvalidRequest, Unauthorized
 from dtest import Tester
+from mock import Mock
```
mambocab: Making this change will require adding `mock` to `requirements.txt`, and also to our install scripts, because our CI setup does not yet use our `requirements.txt`.

consistency_test.py
```diff
@@ -537,10 +535,7 @@ def run():
                 break
 
         if not exceptions_queue.empty():
-            err_type, err, err_traceback = exceptions_queue.get()
-            debug("Failed with exception {}: {}".format(err_type, err.message))
-            traceback.print_exception(err_type, err, err_traceback)
-            assert False, err.message
+            raise exceptions_queue.get()
```
smccarthy788: This is being called in our cassci builds. The error messages are unhelpful. See: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/15/testReport/

Looking more into this.

smccarthy788: Debuggers are great:
```
ipdb> x, y, z = exceptions_queue.get()
ipdb> x
<class 'cassandra.Unavailable'>
ipdb> y
Unavailable('Error from server: code=1000 [Unavailable exception] message="Cannot achieve consistency level ALL" info={\'required_replicas\': 3, \'alive_replicas\': 2, \'consistency\': \'ALL\'}',)
ipdb> z
<traceback object at 0x113fda9e0>
```

`exceptions_queue.get()` is just raising an instance object (or some kind of tupe? Its weird.)

I'm going to test this with `raise exceptions_queue.get()[1]` and see if the error messages become more helpful.

consistency_test.py
```diff
@@ -535,7 +535,7 @@ def run():
                 break
 
         if not exceptions_queue.empty():
-            raise exceptions_queue.get()
+            raise exceptions_queue.get()[1]
```
ptnapoleon: Why is this [1]? What does `exception_queue.get()` return? You should have a comment here

ptnapoleon:
Do all the tests pass now? Can you rebase, and squash the extra commits

consistency_test.py
smccarthy788:
```
ipdb> x, y, z = exceptions_queue.get()
ipdb> x
<class 'cassandra.Unavailable'>
ipdb> y
Unavailable('Error from server: code=1000 [Unavailable exception] message="Cannot achieve consistency level ALL" info={\'required_replicas\': 3, \'alive_replicas\': 2, \'consistency\': \'ALL\'}',)
ipdb> z
<traceback object at 0x113fda9e0>
```

So [1] grabs the actual exception object. Maybe we also want to use the traceback object?

We could go back to how it was, except use `self.fail(err.message)` at the end.

I.e.
```
if not exceptions_queue.empty():
            err_type, err, err_traceback = exceptions_queue.get()
            debug("Failed with exception {}: {}".format(err_type, err.message))
            traceback.print_exception(err_type, err, err_traceback)
            self.fail(err.message)
```

smccarthy788:
Tests have been hanging because meta_tests/assertions.py doesn't have it's mock dependency available. I'm running it again with my bucketizer here: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/17/

consistency_test.py
mambocab: Sean's proposed thing is good, because it gives the exception a named variable and doesn't require a comment.

I've spent some shower time thinking about `MultiError`, and I think it'd be good to use here:

```python
if not exceptions_queue.empty():
    _, exceptions, tracebacks = zip(*exceptions_queue.queue)
    raise MultiError(exceptions=exceptions, tracebacks=tracebacks)
```

^ haven't run the above, but I thiiiiink it's right?

mambocab: Here are the basic mechanisms at work: https://gist.github.com/mambocab/8f019b57d096a4aa56ef1c2e3ebc1089

mambocab: Put a little time into reading up on `Queue`, because I didn't understand why it was necessary to grab `exceptions_queue.queue`. It doesn't look like it's necessary for `exceptions_queue` to be a queue at all; it should probably just be a list. That's out of scope for the current changes, but it'd be nice to make that change at some point.

ptnapoleon:
Can you re-run against the main bucketizer? All dependency issues should be resolved. Also travis is unhappy

c-kodman:
Latest build:
http://cassci.datastax.com/view/Dev/view/c-kodman/job/c-kodman-trunk-dtest/

c-kodman:
http://cassci.datastax.com/view/Dev/view/c-kodman/job/c-kodman-trunk-dtest/lastCompletedBuild/testReport/

ptnapoleon:
Looks like you have three errors in that build, and travis is still sad. Still waiting on the changes to the recent PR from Alex P., as well, then we should be good to merge

c-kodman:
http://cassci.datastax.com/view/Dev/view/c-kodman/job/c-kodman-trunk-dtest/

c-kodman:
http://cassci.datastax.com/view/Dev/view/c-kodman/job/c-kodman-trunk-dtest/

ptnapoleon:
Craig is running a build, but this is ready for final review now, @mambocab want to both tackle, or just me?

mambocab:
> @mambocab want to both tackle, or just me?

I want in plz

c-kodman:
http://cassci.datastax.com/view/Dev/view/c-kodman/job/c-kodman-trunk-dtest/

cql_tests.py
```diff
@@ -139,20 +139,16 @@ def table_test(self):
             session.execute("INSERT INTO test1 (k, v1, v2) VALUES ({i}, {i}, {i})".format(i=i))
             session.execute("INSERT INTO test2 (k, c1, v1) VALUES ({i}, {i}, {i})".format(i=i))
 
-        res = sorted(session.execute("SELECT * FROM test1"))
-        self.assertEqual(rows_to_list(res), [[i, i, i] for i in range(0, 10)])
+        assert_all(session, "SELECT * FROM test1", [[i, i, i] for i in range(0, 10)], ignore_order=True)
 
-        res = sorted(session.execute("SELECT * FROM test2"))
-        self.assertEqual(rows_to_list(res), [[i, i, i] for i in range(0, 10)])
+        assert_all(session, "SELECT * FROM test2", [[i, i, i] for i in range(0, 10)], ignore_order=True)
```
ptnapoleon: Why do these need `ignore_order`

assertions.py
```diff
@@ -1,85 +1,199 @@
...
+    Examples:
+    assert_none(self.session1, "SELECT * FROM test where key=2;")
+    assert_none(cursor, "SELECT * FROM test WHERE k=2", cl=ConsistencyLevel.SERIAL)
+    """
     simple_query = SimpleStatement(query, consistency_level=cl)
     res = session.execute(simple_query)
-    list_res = rows_to_list(res)
-    assert list_res == [], "Expected nothing from %s, but got %s" % (query, list_res)
+    list_res = tools.rows_to_list(res)
+    assert list_res == [], "Expected nothing from {}, but got {}".format(query, list_res)
```
mambocab: Don't block merge on this, but for a follow-up PR: this should be implemented on top of `nose.tools.assertListEqual`, which has better error reporting.

assertions.py
```diff
@@ -102,3 +226,15 @@ def assert_crc_check_chance_equal(session, table, expected, ks="ks", view=False)
...
+def assert_length_equal(object_with_length, expected_length):
+    """
+    Assert an object has a specific length.
+    @param object_with_length The object whose length will be checked
+    @param expected_length The expected length of the object
+
+    Examples:
+    assert_length_equal(res, nb_counter)
+    """
+    assert_equal(len(object_with_length), expected_length, object_with_length)
```
mambocab: Should this have a more verbose message?

dtest.py
```diff
@@ -674,7 +674,7 @@ def create_ks(self, session, name, rf):
             # we assume simpleStrategy
             session.execute(query % (name, "'class':'SimpleStrategy', 'replication_factor':%d" % rf))
         else:
-            assert len(rf) != 0, "At least one datacenter/rf pair is needed"
+            self.assertNotEqual(len(rf), 0, "At least one datacenter/rf pair is needed")
```
ptnapoleon: This should be greaterEqual, we never want negative DCs

batch_test.py
```diff
@@ -191,7 +186,7 @@ def logged_batch_doesnt_throw_uae_test(self):
             APPLY BATCH
         """, consistency_level=ConsistencyLevel.ANY)
         session.execute(query)
-        assert True
+        assert_all(session, "SElECT * FROM users", [[1, u'Will', u'Turner'], [0, u'Jack', u'Sparrow']])
```
mambocab: Could you change this from `SElECT` to `SELECT`?

consistency_test.py
```diff
@@ -141,29 +141,29 @@ def create_users_table(self, session):
...
+        text = "UPDATE users SET age = {} WHERE userid = {}".format(age, userid)
         if serial_consistency and prev_age:
-            text = text + " IF age = %d" % (prev_age)
+            text = text + " IF age = {}".format(prev_age)
         statement = SimpleStatement(text, consistency_level=consistency, serial_consistency_level=serial_consistency)
         session.execute(statement)
 
     def delete_user(self, session, userid, consistency):
-        statement = SimpleStatement("DELETE FROM users where userid = %d" % (userid,), consistency_level=consistency)
+        statement = SimpleStatement("DELETE FROM users where userid = {}".format(userid,), consistency_level=consistency)
```
mambocab: Could you remove the comma in this `format` call?

consistency_test.py
```diff
@@ -141,29 +141,29 @@ def create_users_table(self, session):
...
         session.execute(statement)
 
     def query_user(self, session, userid, age, consistency, check_ret=True):
-        statement = SimpleStatement("SELECT userid, age FROM users where userid = %d" % (userid,), consistency_level=consistency)
+        statement = SimpleStatement("SELECT userid, age FROM users where userid = {}".format(userid,), consistency_level=consistency)
```
mambocab: same as above

scrub_test.py
```diff
@@ -207,7 +207,7 @@ def query_users(self, session):
         ret.extend(list(session.execute("SELECT * FROM users WHERE state='TX'")))
         ret.extend(list(session.execute("SELECT * FROM users WHERE gender='f'")))
         ret.extend(list(session.execute("SELECT * FROM users WHERE birth_year=1978")))
-        assert len(ret) == 8
+        self.assertEqual(len(ret), 8, ret)
```
ptnapoleon: Should be `assert_length_equal`

scrub_test.py
```diff
@@ -353,7 +353,7 @@ def update_users(self, session):
 
     def query_users(self, session):
         ret = list(session.execute("SELECT * FROM users"))
-        assert len(ret) == 5
+        self.assertEqual(len(ret), 5, ret)
```
ptnapoleon: Should be `assert_length_equal`

consistency_test.py
```diff
@@ -175,28 +175,29 @@ def create_counters_table(self, session):
         """)
 
     def update_counter(self, session, id, consistency, serial_consistency=None):
-        text = "UPDATE counters SET c = c + 1 WHERE id = %d" % (id,)
+        text = "UPDATE counters SET c = c + 1 WHERE id = {}".format(id,)
```
mambocab: same as above.

consistency_test.py
```diff
@@ -175,28 +175,29 @@ def create_counters_table(self, session):
...
         statement = SimpleStatement(text, consistency_level=consistency, serial_consistency_level=serial_consistency)
         session.execute(statement)
         return statement
 
     def query_counter(self, session, id, val, consistency, check_ret=True):
-        statement = SimpleStatement("SELECT * from counters WHERE id = %d" % (id,), consistency_level=consistency)
+        statement = SimpleStatement("SELECT * from counters WHERE id = {}".format(id,), consistency_level=consistency)
```
mambocab: same as above.

tools.py
```diff
@@ -176,14 +178,14 @@ def _put_with_overwrite(cluster, session, nb_keys, cl=ConsistencyLevel.QUORUM):
 
 
 def _validate_row(cluster, res):
-    assert len(res) == 100, len(res)
+    assert_equal(len(res), 100)
```
ptnapoleon: This should be assert_length_equal

ptnapoleon:
I left some nits

consistency_test.py
mambocab: Sorry for the repetitive comments. I think the thing to do is `g d master -U0 | grep 'format.*,)'` on your branch. See if you can fix them up with `sed` or something, then make a followup PR.

consistency_test.py
```diff
@@ -537,10 +536,8 @@ def run():
                 break
 
         if not exceptions_queue.empty():
-            err_type, err, err_traceback = exceptions_queue.get()
-            debug("Failed with exception {}: {}".format(err_type, err.message))
-            traceback.print_exception(err_type, err, err_traceback)
-            assert False, err.message
+            _, exceptions, tracebacks = zip(*exceptions_queue.queue)
+            raise MultiError(exceptions=exceptions, tracebacks=tracebacks)
```
mambocab: Did you ever see this in action?

cql_tests.py
```diff
@@ -573,11 +559,9 @@ def many_columns_test(self):
                         ",".join(map(lambda i: str(i), range(width))) +
                         ")")
 
-        res = session.execute("SELECT " +
-                              ",".join(map(lambda i: "c_{}".format(i), range(width))) +
-                              " FROM very_wide_table")
-
-        self.assertEqual(rows_to_list(res), [[i for i in range(width)]])
+        assert_all(session, "SELECT " +
+                   ",".join(map(lambda i: "c_{}".format(i), range(width))) +
```
mambocab: For a followup ticket (and a low-priority one): this would be more idomatic as

```python
','.join(['c_{}'.format(i) for i in range(width)])
```

Similar criticisms for the various `','.join(map(lambda ...` instances around here.

cql_tests.py
```diff
@@ -823,64 +807,45 @@ def lwt_with_static_columns_test(self):
...
 
-        res = session.execute("BEGIN BATCH\n" +
-                              "INSERT INTO lwt_with_static (a, b, d) values (3, 3, 'a');\n" +
-                              "UPDATE lwt_with_static SET s = 3 WHERE a = 3 IF s = null;\n" +
-                              "APPLY BATCH;")
-        self.assertEqual(rows_to_list(res), [[True]])
+        assert_one(session, "BEGIN BATCH\n" +
+                   "INSERT INTO lwt_with_static (a, b, d) values (3, 3, 'a');\n" +
+                   "UPDATE lwt_with_static SET s = 3 WHERE a = 3 IF s = null;\n" +
+                   "APPLY BATCH;", [True])
```
mambocab: Formatting nit for a followup PR: I'd prefer to see this as

```python
 assert_one(
    session,
    ("BEGIN BATCH\n"
     "INSERT INTO lwt_with_static (a, b, d) values (3, 3, 'a');\n"
     "UPDATE lwt_with_static SET s = 3 WHERE a = 3 IF s = null;\n"
     "APPLY BATCH;"),
    [True]
)
```

Same goes for the batch statement a few lines down from here.

mambocab: Ignore my earlier comment; that formatting stuff isn't important right now

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -1767,7 +1768,7 @@ def _test(prepared_statements):
 
             imported_results = list(self.session.execute("SELECT * FROM testdatatype"))
 
-            assert len(imported_results) == 1
+            assert_length_equal(imported_results, 1)
```
mambocab: Since there's an `assertEqual` below, do we even need an `assert_length_equal` here?

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -1812,7 +1813,7 @@ def do_round_trip(trueval, falseval, invalid=False):
                                  .format(tempfile.name, trueval, falseval))
 
             imported_results = list(self.session.execute("SELECT * FROM testbooleans"))
-            assert len(imported_results) == 2
+            assert_length_equal(imported_results, 2)
```
mambocab: Since there's an `assertEqual` below, do we even need an `assert_length_equal` here?

meta_tests/assertion_test.py
```diff
@@ -0,0 +1,52 @@
+from cassandra import AlreadyExists, Unavailable, InvalidRequest, Unauthorized
+from mock import Mock
+from assertions import (assert_one, assert_none, assert_exception, assert_unavailable,
+                        assert_invalid, assert_unauthorized, assert_all, assert_almost_equal,
+                        assert_length_equal, assert_row_count)
+
+
+class TestAssertionMethods():
+
+    def assertions_test(self):
```
mambocab: For a followup PR: I see no reason for all these assertions to be in the same test; they should be moved into separate test methods IMO.

paging_test.py
```diff
@@ -516,7 +516,8 @@ def handle_scenario(scenario):
                 )
             else:
                 # this should not happen
-                assert False
+                self.fail("Invalid scenario configuration. Scenario is: {}".format(scenario))
+                self.assertTrue(False, "This should not happen.")
```
mambocab: You can remove this assertion; the `fail` above is enough

paging_test.py
```diff
@@ -2207,8 +2129,8 @@ def test_ttl_deletions(self):
         for row in data:
             s = ("insert into paging_test (id, mytext, col1, col2, col3) "
                  "values ({}, '{}', {}, {}, {}) using ttl 3;").format(
-                     row['id'], row['mytext'], row['col1'],
-                     row['col2'], row['col3'])
+                row['id'], row['mytext'], row['col1'],
```
mambocab: Is `pep8` down with this?

paxos_tests.py
```diff
@@ -189,4 +189,4 @@ def run(self):
             errors = errors + w.errors
             retries = retries + w.retries
 
-        assert (value == threads * iterations) and (errors == 0), "value=%d, errors=%d, retries=%d" % (value, errors, retries)
+        self.assertTrue((value == threads * iterations) and (errors == 0), "value={}, errors={}, retries={}".format(value, errors, retries))
```
mambocab: Seems like this would have better output on failure with

```python
self.assertEqual(value, threads * iterations)
self.assertEqual(0, errors)
```

I haven't read the test, but why is `retries` printed in the `msg` for an assertion that doesn't use it? @ptnapoleon should we be asserting something about `retries` too?

paging_test.py
ptnapoleon: Evidently, since travis is

paxos_tests.py
ptnapoleon: +1 to two assertions. I don't know that we need another, but I also don't know that we don't.

read_repair_test.py
mambocab: I think all 3 of these assertions could be condensed to

```python
self.assertIn(res, [[[1, 1]],
                    [[1, None]]])

File: cql_tests.py (hunk 70 above)
Comment (ptnapoleon) [inline cql_tests.py]: I see why now, I guess it's fine to leave it like this

File: assertions.py (hunk 71 above)
Comment (ptnapoleon) [inline assertions.py]: Isn't it assert_list_equal?

Comment (ptnapoleon) [inline assertions.py]: Also what's the benefit on comparing it here, with none? Wouldn't it be better served in assert_one and assert_all

File: assertions.py (hunk 72 above)
Comment (ptnapoleon) [inline assertions.py]: Yes, and I thought it already did, but I guess not

File: consistency_test.py (hunk 76 above)
Comment (ptnapoleon) [inline consistency_test.py]: Damn, I fixed that and never pushed, my bad

File: consistency_test.py (hunk 82 above)
Comment (ptnapoleon) [inline consistency_test.py]: No

File: meta_tests/assertion_test.py (hunk 87 above)
Comment (ptnapoleon) [inline meta_tests/assertion_test.py]: I think that's left over from when we weren't mocking

File: assertions.py (hunk 71 above)
Comment (mambocab) [inline assertions.py]: > Isn't it assert_list_equal?

Yes.

> Also what's the benefit on comparing it here, with none?

Just better error reporting in edge cases we didn't think about.

> Wouldn't it be better served in assert_one and assert_all

oh, yeah. sorry, incomplete thought on my part. I should have said that in general, we should try to remove bare `assert`s from `assertions.py`.

File: consistency_test.py (hunk 76 above)
Comment (mambocab) [inline consistency_test.py]: As I mentioned elsewhere, should be doable with `sed`. I'd rather get this thing merged and deal with making and testing this change in a followup commit.

File: read_repair_test.py [hunk 91]
```diff
@@ -162,9 +158,9 @@ def range_slice_query_with_tombstones_test(self):
         for trace_event in trace.events:
             # Step 1, find coordinator node:
             activity = trace_event.description
-            assert "Appending to commitlog" not in activity
-            assert "Adding to cf memtable" not in activity
-            assert "Acquiring switchLock read lock" not in activity
+            self.assertNotIn("Appending to commitlog", activity)
```
mambocab: For a followup PR: this implicitly depends on trace events that log `"Appending to commitlog"`, `"Adding to cf memtable"`, and `"Acquiring switchLock read lock"`, so we should make sure there are tests for that behavior.

scrub_test.py
```diff
@@ -101,7 +101,7 @@ def launch_nodetool_cmd(self, cmd):
         node1 = self.cluster.nodelist()[0]
         response = node1.nodetool(cmd, capture_output=True)[0]
         if not common.is_win():  # nodetool always prints out on windows
-            assert len(response) == 0, response  # nodetool does not print anything unless there is an error
+            self.assertEqual(len(response), 0, response)  # nodetool does not print anything unless there is an error
```
mambocab: `assert_length_equal`?

scrub_test.py
mambocab: `assert_length_equal`?

tools.py
```diff
@@ -113,9 +115,9 @@ def insert_columns(tester, session, key, columns_count, consistency=ConsistencyL
 def query_columns(tester, session, key, columns_count, consistency=ConsistencyLevel.QUORUM, offset=0):
     query = SimpleStatement('SELECT c, v FROM cf WHERE key=\'k%s\' AND c >= \'c%06d\' AND c <= \'c%06d\'' % (key, offset, columns_count + offset - 1), consistency_level=consistency)
     res = list(session.execute(query))
-    assert len(res) == columns_count, "%s != %s (%s-%s)" % (len(res), columns_count, offset, columns_count + offset - 1)
+    assert_equal(len(res), columns_count, "{} != {} ({}-{})".format(len(res), columns_count, offset, columns_count + offset - 1))
```
mambocab: `assert_length_equal`?

tools.py
```diff
@@ -196,7 +198,7 @@ def range_putget(cluster, session, cl=ConsistencyLevel.QUORUM):
     paged_results = session.execute('SELECT * FROM cf LIMIT 10000000')
     rows = [result for result in paged_results]
 
-    assert len(rows) == keys * 100, len(rows)
+    assert_equal(len(rows), keys * 100)
```
mambocab: `assert_length_equal`?

upgrade_tests/paging_test.py
```diff
@@ -7,23 +7,18 @@
...
 from datahelp import create_rows, flatten_into_set, parse_data_into_dicts
 from dtest import debug, run_scenarios
 from tools import known_failure, rows_to_list, since
 from upgrade_base import UPGRADE_TEST_RUN, UpgradeTester
 from upgrade_manifest import build_upgrade_pairs
 
+from nose.tools import assert_not_in
+
 
 def assert_read_timeout_or_failure(session, query):
```
mambocab: For a followup PR: any reason not to move this to `assertions`?

mambocab:
Left some comments, mostly nits. Great work, folks, some awesome changes in there.

Could you collate my followup requests into an internal JIRA ticket?

paxos_tests.py
mambocab: Cool. Could someone please link this discussion in a followup thread?

meta_tests/assertion_test.py
mambocab: It is.

counter_tests.py
```diff
@@ -38,10 +38,10 @@ def simple_increment_test(self):
             query = SimpleStatement("SELECT key, c FROM cf WHERE key IN (%s)" % keys, consistency_level=ConsistencyLevel.QUORUM)
             res = list(session.execute(query))
 
-            assert len(res) == nb_counter
+            assert_length_equal(res, nb_counter)
             for c in xrange(0, nb_counter):
```
mambocab: For a followup PR: iterating over indexes is bad form; we should iterate over `res` itself.

consistency_test.py
c-kodman: Jim, what do you want us to do with this?

ptnapoleon:
Merging!

PR #1014, opened 2016-06-01
Readability nits

mambocab:
Had a few style and readability nits to pick with [a recent PR](https://github.com/riptano/cassandra-dtest/pull/998) that I didn't see until after it was merged.

Locally, my changes pass on `trunk` and `cassandra-3.0`, and fail as expected (with a bunch of `org.apache.cassandra.exceptions.WriteTimeoutException: Operation timed out - received only 0 responses.` messages) on `3.0.5`, which is consistent with the bug describtion in [CASSANDRA-11891](https://issues.apache.org/jira/browse/CASSANDRA-11891).

### DIFF
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -205,8 +205,7 @@ class TestCommitLog(Tester):
         node1.mark_log_for_errors()
 
         debug("Verify commitlog was written before abrupt stop")
-        commitlog_dir = os.path.join(node1.get_path(), 'commitlogs')
-        commitlog_files = os.listdir(commitlog_dir)
+        commitlog_files = os.listdir(os.path.join(node1.get_path(), 'commitlogs'))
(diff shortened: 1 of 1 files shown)


mambocab:
The while loop logic diff is hard to read. Have a look at the last 2 commits and their messages individually; that should be more understandable.

ptnapoleon:
+1, though I should note we use that while loop logic in several places, and seems like reasonable Python to me?

mambocab:
It's reasonable, but difficult to read. `This loop runs until replay is done. The test fails if this loop takes 2 minutes or longer.` is more readable than `This loop runs for 2 minutes, unless replay finishes. If the loop exits (except via the replay-finishing exit path), the test fails`, IMO.

ptnapoleon:
Yes, you are correct.

mambocab:

PR #1015, opened 2016-06-01
Cstar 565

smccarthy788:
Fixed all the places where it broke the build.

@mambocab, @ptnapoleon please review.

Passed cassci tests: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/4/


ptnapoleon:
Looks good now, just needs squashed

PR #1018, opened 2016-06-03
annotated known failure

smccarthy788:
@ptnapoleon, please review.

### DIFF
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -355,6 +355,9 @@ class TestAvailability(TestHelper):
 
         self._test_network_topology_strategy(combinations)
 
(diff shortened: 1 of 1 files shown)


consistency_test.py
```diff
@@ -356,6 +356,9 @@ def test_network_topology_strategy(self):
         self._test_network_topology_strategy(combinations)
 
     @since("3.0")
+    @known_failure(failure_source='cassandra',
```
ptnapoleon: I think this has to go above the `since`?

mambocab: Correct. Do we want to add this to `contributing.md`?

PR #1022, opened 2016-06-06
Added a user defined compaction test, CASSANDRA-11765

stef1927:
A new test for CASSANDRA-11765.

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -2,11 +2,14 @@ import os
 import random
 import re
 import string
+import subprocess
 import tempfile
 import time
 
 from assertions import assert_none, assert_one
+from ccmlib import common
 from dtest import Tester, debug
+from nose.tools import assert_equal
 from tools import known_failure, since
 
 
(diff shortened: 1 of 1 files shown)


compaction_test.py
```diff
@@ -451,11 +453,59 @@ def disable_autocompaction_alter_and_nodetool_test(self):
...
 
+def get_sstable_data_files(node, ks, table):
+    """
+    Read sstable data files by using sstableutil, so we ignore temporary files
+    """
+    env = common.make_cassandra_env(node.get_install_cassandra_root(), node.get_node_cassandra_root())
+    args = [node.get_tool('sstableutil'), '--type', 'final', ks, table]
+
+    p = subprocess.Popen(args, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    (stdout, stderr) = p.communicate()
```
ptnapoleon: Very minor style nit, not even worth fixing, but you don't need the parens around `stdout, stderr` here

compaction_test.py
```diff
@@ -451,11 +453,59 @@ def disable_autocompaction_alter_and_nodetool_test(self):
...
+
+    if p.returncode != 0:
```
ptnapoleon: Could we make this block
```
assert_equal(p.returncode, 0, "Error invoking sstableutil; returned {code}; {err}".format(code=p.returncode, err=stderr))
```
? Where we get assert_equal with `from nose.tools import assert_equal`

compaction_test.py
```diff
@@ -451,11 +453,59 @@ def disable_autocompaction_alter_and_nodetool_test(self):
...
+        sstable_files = get_sstable_data_files(node1, 'keyspace1', 'standard1')
+        self.assertEquals(len(node1.data_directories()), len(sstable_files),
+                          'Expected one sstable data file per node directory but got {}'.format(sstable_files))
+
     def skip_if_no_major_compaction(self):
         if self.cluster.version() < '2.2' and self.strategy == 'LeveledCompactionStrategy':
             self.skipTest('major compaction not implemented for LCS in this version of Cassandra')
 
 
+def get_sstable_data_files(node, ks, table):
```
ptnapoleon: Do we not already have this in ccm? Maybe we should add it there, not in this PR, but we'll ask either @smccarthy788 or @c-kodman to do it later

mambocab: Not digging very deep right now, but:

https://github.com/pcmanus/ccm/blob/4ab1bb9f01c2f8b902175c5bfe48a14deda39868/ccmlib/node.py#L1159

This uses globs rather than `sstableutil`, so I can see the value in having both.

mambocab: But, if the linked method is correct for this purpose, we should use it. If it's not, then committing this method to dtest is fine, and we can push this behavior down into ccm later.

stef1927:
I've fixed the two nits and squashed, thanks. 

You definitely need to add `get_sstable_data_files` to ccm, it is not the same as checking the files on disk because in this case we need to retrieve only the final files and not the temporary files. You also want to take a look at `_invoke_sstableutil` and `_get_sstable_files` in `sstableutil_test.py`, we should be able to replace them with ccm methods.

ptnapoleon:
+1

stef1927:
Thank you, merging.

PR #1023, opened 2016-06-07
8844 tests pr1

mambocab:
Depends on #1008. Another gross history; please don't merge until that's addressed. Alternately, we could squash both PRs together into one big thing. Up to reviewers -- @ptnapoleon and/or @knifewine, and @josh-mckenzie, please.

@josh-mckenzie I'm still not seeing the behavior I expect after flush -- I still don't find any commitlogs in `cdc_raw`. I assume I'm doing something wrong here; what am I missing?

### DIFF
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -204,6 +204,29 @@ class TestCDC(Tester):
     provides a view of the commitlog on tables for which it is enabled.
     """
 
+    def _create_temp_dir(self, dir_name, verbose=True):
+        """
+        Create a directory that will be deleted when this test class is torn
+        down.
+        """
+        if verbose:
+            debug('creating ' + dir_name)
+        try:
+            os.mkdir(dir_name)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                debug(dir_name + ' already exists. removing and recreating.')
+                shutil.rmtree(dir_name)
+                os.mkdir(dir_name)
+            else:
+                raise e
+
+        def debug_and_rmtree():
+            shutil.rmtree(dir_name)
+            debug(dir_name + ' removed')
+
+        self.addCleanup(debug_and_rmtree)
+
     def prepare(self, ks_name,
                 table_name=None, cdc_enabled_table=None,
                 gc_grace_seconds=None,
@@ -420,19 +443,92 @@ class TestCDC(Tester):
         session.cluster.shutdown()
         self.assertEqual(pre_non_cdc_write_cdc_raw_segments, _get_cdc_raw_files(node.get_path()))
 
-    # TODO: add tests to determine that CDC data is correctly flushed to
-    # cdc_raw and not discarded. Our basic goal is to write dataset A to CDC
-    # tables and dataset B to non-CDC tables, flush, then check that a superset
-    # of data A is available to a client in cdc_raw. We can do this by:
-    #   - writing a mixed CDC/non-CDC dataset, tracking all data that should be
-    #     exposed via CDC,
-    #   - flushing,
-    #   - saving off the contents of cdc_raw,
-    #   - shutting down the cluster,
-    #   - starting a new cluster,
-    #   - initializing schema as necessary,
-    #   - shutting down the new cluster,
-    #   - moving the saved cdc_raw contents to commitlog directories,
-    #   - starting the new cluster to commitlog replay, then
-    #   - asserting all data that should have been exposed via CDC is in the
-    #     table(s) that have been written to tables in the new cluster.
+    def _init_new_loading_node(self, ks_name, create_stmt):
+        loading_node = Node(
+            name='node2',
+            cluster=self.cluster,
+            auto_bootstrap=False,
+            thrift_interface=('127.0.0.2', 9160),
+            storage_interface=('127.0.0.2', 7000),
+            jmx_port='7400',
+            remote_debug_port='0',
+            initial_token=None,
+            binary_interface=('127.0.0.2', 9042)
+        )
+        debug('adding node')
+        self.cluster.add(loading_node, is_seed=True)
+        debug('starting new node')
+        loading_node.start(wait_for_binary_proto=True)
+        debug('recreating ks and table')
+        loading_session = self.patient_exclusive_cql_connection(loading_node)
+        self.create_ks(loading_session, ks_name, rf=1)
+        debug('creating new table')
+        loading_session.execute(create_stmt)
+        debug('stopping new node')
+        loading_node.stop()
+        loading_session.cluster.shutdown()
+        return loading_node
+
+    def test_cdc_data_available_in_cdc_raw(self):
+        ks_name = 'ks'
+        # First, create a new node just for data generation.
+        generation_node, generation_session = self.prepare(ks_name=ks_name)
+
+        cdc_table_info = TableInfo(
+            ks_name=ks_name, table_name='cdc_tab',
+            column_spec=_16_uuid_column_spec,
+            insert_stmt=_get_16_uuid_insert_stmt(ks_name, 'cdc_tab'),
+            options={
+                'cdc': 'true',
+                # give table an explicit id so when we create it again it's the
+                # same table and we can replay into it
+                'id': uuid.uuid4()
+            }
+        )
+        generation_session.execute(cdc_table_info.create_stmt)
+
+        # insert 10000 rows
+        inserted_rows = _insert_rows(generation_session, cdc_table_info.name, cdc_table_info.insert_stmt, repeat((), 10000))
+
+        # drain the node to guarantee all cl segements will be recycled
+        debug('draining')
+        generation_node.drain()
+        debug('stopping')
+        # stop the node and clean up all sessions attached to it
+        generation_node.stop()
+        generation_session.cluster.shutdown()
+
+        # create a new node to use for cdc_raw cl segment replay
+        loading_node = self._init_new_loading_node(ks_name, cdc_table_info.create_stmt)
+
+        # move cdc_raw contents to commitlog directories, then start the
+        # node again to trigger commitlog replay, which should replay the
+        # cdc_raw files we moved to commitlogs into memtables.
+        debug('moving cdc_raw and restarting node')
+        _move_contents(
+            os.path.join(generation_node.get_path(), 'cdc_raw'),
+            os.path.join(loading_node.get_path(), 'commitlogs')
+        )
+        loading_node.start(wait_for_binary_proto=True)
+        debug('node successfully started; waiting on log replay')
+        loading_node.grep_log('Log replay complete')
+        debug('log replay complete')
+
+        # final assertions
+        validation_session = self.patient_exclusive_cql_connection(loading_node)
+        data_in_cdc_table_after_restart = rows_to_list(
+            validation_session.execute('SELECT * FROM ' + cdc_table_info.name)
+        )
+        debug('found {cdc} values in CDC table'.format(
+            cdc=len(data_in_cdc_table_after_restart)
+        ))
+        # Then we assert that the CDC data that we expect to be there is there.
+        # All data that was in CDC tables should have been copied to cdc_raw,
+        # then used in commitlog replay, so it should be back in the cluster.
+        self.assertEqual(
+            inserted_rows,
+            data_in_cdc_table_after_restart,
+            # The message on failure is too long, since cdc_data is thousands
+            # of items, so we print something else here
+            msg='not all expected data selected'
+        )
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -1,17 +1,155 @@
 from __future__ import division
 
+import errno
 import os
+import shutil
 import time
+import uuid
+from collections import namedtuple
 from itertools import izip as zip
+from itertools import repeat
 
 from cassandra import WriteFailure
 from cassandra.concurrent import (execute_concurrent,
                                   execute_concurrent_with_args)
+from ccmlib.node import Node
+from nose.tools import assert_equal, assert_less_equal
 
 from dtest import Tester, debug
-from tools import since
+from tools import rows_to_list, since
 from utils.fileutils import size_of_files_in_dir
 from utils.funcutils import get_rate_limited_function
+
+_16_uuid_column_spec = (
+    'a uuid PRIMARY KEY, b uuid, c uuid, d uuid, e uuid, f uuid, g uuid, '
+    'h uuid, i uuid, j uuid, k uuid, l uuid, m uuid, n uuid, o uuid, '
+    'p uuid'
+)
+
+
+def _insert_rows(session, table_name, insert_stmt, values):
+    prepared_insert = session.prepare(insert_stmt)
+    values = list(values)  # in case values is a generator
+    execute_concurrent(session, ((prepared_insert, x) for x in values),
+                       concurrency=500, raise_on_first_error=True)
+
+    data_loaded = rows_to_list(session.execute('SELECT * FROM ' + table_name))
+    debug('{n} rows inserted into {table_name}'.format(n=len(data_loaded), table_name=table_name))
+    # use assert_equal over assert_length_equal to avoid printing out
+    # potentially large lists
+    assert_equal(len(values), len(data_loaded))
+    return data_loaded
+
+
+def _move_contents(source_dir, dest_dir, verbose=True):
+    for source_filename in os.listdir(source_dir):
+        source_path, dest_path = (os.path.join(source_dir, source_filename),
+                                  os.path.join(dest_dir, source_filename))
+        if verbose:
+            debug('moving {} to {}'.format(source_path, dest_path))
+        shutil.move(source_path, dest_path)
+
+
+def _get_16_uuid_insert_stmt(ks_name, table_name):
+    return (
+        'INSERT INTO {ks_name}.{table_name} '
+        '(a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p) '
+        'VALUES (uuid(), uuid(), uuid(), uuid(), uuid(), '
+        'uuid(), uuid(), uuid(), uuid(), uuid(), uuid(), '
+        'uuid(), uuid(), uuid(), uuid(), uuid())'
+    ).format(ks_name=ks_name, table_name=table_name)
+
+
+def _get_create_table_statement(ks_name, table_name, column_spec, options=None):
+    if options:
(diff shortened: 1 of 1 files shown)


ptnapoleon:
At a general skim, I like it. Why so much string concat instead of interpolating? Will wait for you to hear back from Josh before doing a thorough review

mambocab:
> Why so much string concat instead of interpolating?

When generating CQL, I know the inputs will be strings, or at least I'm ok with the statement generation blowing up if I try to concat with a non-string. I thought it was clearer in this case; do you agree?

mambocab:
This needs some refactoring for readability, but I would like some feedback from @josh-mckenzie, please. The jury-rigged "write data, flush to `cdc_raw`, save off `cdc_raw` contents, truncate and recreate the keyspace/tables, move old `cdc_raw` contents back to the commitlog dir, start the node to replay them, ?????, profit" maneuver isn't actually loading any data that I can `SELECT`. Can you tell what I'm doing wrong here?

mambocab:
After private discussion with Josh, I used `TRUNCATE` instead of `DROP` to remove data from the initial versions of the table. I'm still not successfully loading data into the tables on replay. Here's my output:

https://gist.github.com/mambocab/973cbb6b08eb0dcd935b5b99cf7746c9

So, `cdc_raw` is successfully populated, I copy the files out, `TRUNCATE`, stop the database, move the old `cdc_raw` contents back, and see in the logs that replay is complete, but when I `SELECT *` from the tables in the restarted db, I get nothing.

mambocab:
Ah, `TRUNCATE` might not do it either: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/commitlog/CommitLogReplayer.java#L98

mambocab:
Ok, I removed all references to CDC and tried it on 3.0.6, and it doesn't work. I'm guessing that C* doesn't replay the commitlogs because of the `TRUNCATE` record.

mambocab:
Setting `gc_grace_period=0` on the tables to clear, then `DELETE`ing every row, is working; I can now successfully replay old commitlogs back into tables and the test passes as of 8279c3e.

@josh-mckenzie @ptnapoleon @knifewine There's a lot of repetition I want to get rid of in here. I can wait for review, squash, merge, and file a followup, or I can spend another, say, half-day or so refactoring before merge. Opinions?

cdc_test.py
```diff
@@ -335,19 +354,129 @@ def test_insertion_and_commitlog_behavior_after_reaching_cdc_total_space(self):
...
+        self._create_temp_dir(saved_cdc_raw_contents_dir_name)
+
+        # Save cdc_raw files off to temporary directory
+        node.flush()
+        raw_dir = os.path.join(node.get_path(), 'cdc_raw')
+        cdc_raw_files = os.listdir(raw_dir)
+        debug('saving {n} file(s) from cdc_raw'.format(n=len(cdc_raw_files)))
+        for original_cdc_raw_filename in cdc_raw_files:
+            # (it'd be nice to just ln this, but we do a copy for Windows compat)
+            shutil.copy2(
```
josh-mckenzie: Why copy instead of move? There should be no handles open to the files in cdc_raw once C* is done with them so no problems on Windows.

cdc_test.py
```diff
@@ -335,19 +354,129 @@ def test_insertion_and_commitlog_behavior_after_reaching_cdc_total_space(self):
...
+                os.path.join(raw_dir, original_cdc_raw_filename),
+                os.path.join(saved_cdc_raw_contents_dir_name, original_cdc_raw_filename)
+            )
+
+        # Start clean so we can "import" commitlog files
```
josh-mckenzie: I think we can actually go the drop / recreate route, just need to use:
CREATE TABLE t1 (a date, b text, primary key(a)) WITH id=<cfId>; syntax.

Should be able to query out the cfid with:
cqlsh:system> select table_name, id from system_schema.tables;

Just tested, appears to work:
cqlsh:ks1> create table t1 (a date, b text, primary key(a)) with id=3a720cf0-3962-11e6-9800-3992d25a9de4;

cdc_test.py
mambocab: Hadn't thought about it. My gut reaction:

- In this case, the runtime overhead of copy vs. move should be pretty small. I don't think we care about disk overhead.
- Using move will make this test fail if the file handle behavior you describe changes, which is not the behavior we care about in this test.

I'll time the copy locally to see if the first is true.

mambocab: In the second copy, though, there's no reason not to move.

cdc_test.py
mambocab: To persist it publicly: Josh brought up that this would probably be faster than deleting every row. I'm timing it out now; my naive loop-and-delete code takes 3-5s per table, which could definitely be much better.

mambocab:
To address https://github.com/riptano/cassandra-dtest/pull/1023#discussion_r68271003, I pushed eaba8e1198b7704ee68777d6f1b9e68dc22508d5. It takes less than half the time, but the test fails now:

[8844-failure-ouptut.txt](https://github.com/riptano/cassandra-dtest/files/330563/8844-failure-ouptut.txt)

The relevant output is

```
log replay complete
found 4074 values in CDC table and 1987 values in non-CDC table
```

So, I'm not sure how this is getting different behavior than deleting the entire contents of the tables with gc_grace set to 0. Looking in the debug logs, I'm seeing

```
INFO  [main] 2016-06-23 14:09:11,972 CommitLog.java:167 - Log replay complete, 6061 replayed mutations
```

Which matches what we'd expect given the number of rows in the two tables.

mambocab:
> Are you missing 'AND cdc=true'?

In the table re-creation step: totally. Didn't seem to affect behavior though; I added it back and saw the same behavior.

mambocab:
Properly debugging this is getting pretty difficult; I'm going to put in some time on that refactor to make it easier. Do let me know if you have thoughts, though.

mambocab:
Most of the major code changes are in place, so a review from @ptnapoleon would be helpful here. I'm still debugging the failing test; it won't be ready to merge until that's figured out.

cdc_test.py
```diff
@@ -335,19 +426,93 @@ def test_insertion_and_commitlog_behavior_after_reaching_cdc_total_space(self):
...
+        self._create_temp_dir(saved_cdc_raw_contents_dir_name)
+
+        # Save cdc_raw files off to temporary directory
+        node.flush()
+        raw_dir = os.path.join(node.get_path(), 'cdc_raw')
+        _move_contents(raw_dir, saved_cdc_raw_contents_dir_name)
+
+        # Start clean so we can "import" commitlog files
+        debug('dropping and recreating tables')
+        session.execute('DROP TABLE ' + ks_name + '.' + cdc_table_info.table_name)
```
mambocab: I want to pull this "drop table, recreate table, assert table is emtpy" logic out into a method.

mambocab:
Ok, here's what I've found so far:

- If I cache all commitlogs, then replay them into the recreated tables, the test passes, so everything got replayed. This indicates that commitlogs are being written correctly. (see https://github.com/mambocab/cassandra-dtest/commit/89be29650f7151764cb3e331bdc7335415be5e2f)
- If I create both tables with `cdc=true`, not all values that should be there are replayed into the recreated tables (`dtest: DEBUG: found 4072 values in CDC table and 8050 values in non-CDC table`). The only notable difference is that more values are replayed into the other table -- the behavior on HEAD for this PR is `dtest: DEBUG: found 4076 values in CDC table and 1985 values in non-CDC table`. (for the debugging commit, see https://github.com/mambocab/cassandra-dtest/commit/15faeeec439d4d37140126cb661b8024e23be9dd)

@josh-mckenzie Do you have any thoughts on this?

cdc_test.py
```diff
@@ -419,24 +440,35 @@ def test_cdc_data_available_in_cdc_raw(self):
...
+        delete_2_start_time = time.time()
+        session.execute('DROP TABLE ' + ks_name + '.' + non_cdc_table_name)
+        session.execute(
+            'CREATE TABLE ' + ks_name + '.' + non_cdc_table_name + ' '
+            '(a uuid PRIMARY KEY, b uuid, c uuid, d uuid, e uuid, '
+            'f uuid, g uuid, h uuid, i uuid, j uuid, k uuid, l uuid, '
+            'm uuid, n uuid, o uuid, p uuid) WITH gc_grace_seconds=0 AND id={non_cdc_table_id}'.format(non_cdc_table_id=non_cdc_table_id)
+        )
+        debug('delete 2 took {0:.2f}s'.format(time.time() - delete_2_start_time))
+        self.assertEqual(0, len(list(session.execute('SELECT * FROM ' + ks_name + '.' + non_cdc_table_name))))
```
ptnapoleon: Why not assert_length_equal, other than because it doesn't exist yet nvm

ptnapoleon:
Once cstar-67 is in, use `assert_length_equal` all around, please

cdc_test.py
```diff
@@ -1,19 +1,140 @@
...
 
 
+_16_uuid_column_spec = (
+    'a uuid PRIMARY KEY, b uuid, c uuid, d uuid, e uuid, f uuid, g uuid, '
+    'h uuid, i uuid, j uuid, k uuid, l uuid, m uuid, n uuid, o uuid, '
+    'p uuid'
+)
+
+
+def _move_contents(source_dir, dest_dir, verbose=True):
```
ptnapoleon: Does this really not exist elsewhere, even in the stdlib? Why so much wrap around shutil.move

mambocab:
> use `assert_length_equal`

ah, yep. can do, thank you.

cdc_test.py
```diff
@@ -1,19 +1,140 @@
...
+        # If we haven't logged from here in the last 5s, do so.
+        rate_limited_debug(
+            '  data load step has lasted {s:.2f}s, '
+            'loaded {r} rows'.format(s=(time.time() - start), r=rows_loaded))
+
+        batch_results = list(execute_concurrent(
+            session,
+            ((prepared, ()) for _ in range(1000)),
+            concurrency=500,
+            # Don't propogate errors to the main thread. We expect at least
```
ptnapoleon: nit: `propagate`

cdc_test.py
mambocab: I want to move the contents only, not the dir itself -- [`shutil.move` moves the entire directory](https://docs.python.org/2/library/shutil.html#shutil.move), which I don't want to do here.

Reimplementing it here has the added benefit of letting me insert debug statements.

I didn't check to see if there was another implementation already in `dtest` somewhere; you want me to dig around for that?

cdc_test.py
```diff
@@ -66,16 +187,41 @@ class TestCDC(Tester):
...
+        """
+        Create a directory that will be deleted when this test class is torn
+        down.
+        """
+        if verbose:
+            debug('creating ' + dir_name)
+        try:
+            os.mkdir(dir_name)
+        except OSError as e:
+            if 'File exists' in e:
```
ptnapoleon: Is this check correct? It's not `e.msg` or something similar?

ptnapoleon:
```
In [3]: try:
    raise OSError('error')
except OSError as e:
    print e
   ...:
error

In [4]: try:
    raise OSError('error')
except OSError as e:
    if 'err' in e:
   ...:         print e
   ...:

In [5]:
```
From `ipython`. Maybe concerning?

mambocab: Good thinking, let me dig for a few.

ptnapoleon:
My two comments are all I've got here

cdc_test.py
josh-mckenzie: > If I create both tables with cdc=true, not all values that should be there are replayed into the recreated tables (dtest: DEBUG: found 4072 values in CDC table and 8050 values in non-CDC table).
What count do you expect in CDC enabled and what count in non in this context? The below listed counts (4076 / 1985)?

> The only notable difference is that more values are replayed into the other table
What is "other"?

If the code works correctly w/the initial approach rather than the "replay using cfId" approach, that to me implies that the C* code is working correctly and we're probably missing something w/our assumptions about re-creating a table re-using a cfId. What's our absolute delta on time here for the run? Are we talking 10 min instead of 20, or 30 seconds instead of 60? If the latter, I'd say just get this thing committed w/the initial implementation since 30 seconds of runtime isn't worth a) the headache and b) being exposed to potential future changes in cfid re-use on table recreation breaking your test.

cdc_test.py
mambocab: What we want is to import `errno` and then make this condition `if e.errno == errno.EEXIST`.

ptnapoleon: Oooh, that's a great fix

mambocab:
Alriiiiiight, got the test working with help from Josh. Thanks!

I will squash before merge, and I'll fix Travis's complaints, but this is otherwise ready for review. All tests in `cdc_test` pass on trunk. You can watch the new test fail by commenting out [the commitlog-moving step](https://github.com/mambocab/cassandra-dtest/blob/373e0f1999e566aab21da0c14577f3f10a014b60/cdc_test.py#L502-L505), to confirm there isn't anything wacky going on between the two nodes.

mambocab:
@ptnapoleon I've marked this as awaiting dev feedback since I won't merge without @josh-mckenzie's review, but I'm waiting on a final review from you as well.

ptnapoleon:
I'm +1

ptnapoleon:
This fails on trunk for me with
```
org.apache.cassandra.exceptions.ConfigurationException: Cannot change the number of tokens from 512 to 256
```

mambocab:
Ran all the tests on trunk locally, they all passed. Can you repro on CassCI?

mambocab:
Or generally give me more details?

mambocab:
Ping @josh-mckenzie, could you review?

@ptnapoleon did you ever figure out how to repro that failure you saw locally?

josh-mckenzie:
Will take a look tomorrow - sorry for the delay. Fell off my radar.

cdc_test.py
```diff
@@ -1,18 +1,156 @@
...
+    'h uuid, i uuid, j uuid, k uuid, l uuid, m uuid, n uuid, o uuid, '
+    'p uuid'
+)
+
+
+def _insert_rows(session, table_name, insert_stmt, values):
+    prepared_insert = session.prepare(insert_stmt)
+    values = list(values)  # in case values is a generator
+    execute_concurrent(session, ((prepared_insert, x) for x in values),
+                       concurrency=500, raise_on_first_error=True)
```
josh-mckenzie: Why 500 on concurrency value? Not a huge deal, just idly curious. Given any host running this test is going to have far < 500 cores.

cdc_test.py
```diff
@@ -1,18 +1,156 @@
...
+    return data_loaded
+
+
+def _move_contents(source_dir, dest_dir, verbose=True):
+    for source_filename in os.listdir(source_dir):
+        source_path, dest_path = (os.path.join(source_dir, source_filename),
+                                  os.path.join(dest_dir, source_filename))
+        if verbose:
+            debug('moving {} to {}'.format(source_path, dest_path))
+        shutil.move(source_path, dest_path)
```
josh-mckenzie: Have you tried running this dtest on Windows? If not, want me to give it a whirl?

cdc_test.py
```diff
@@ -1,18 +1,156 @@
...
+    else:
+        options_string = ''
+
+    return (
+        'CREATE TABLE ' + ks_name + '.' + table_name + ' '
+        '(' + column_spec + ') ' + options_string
+    )
+
+
+def _write_to_cdc_WriteFailure(session, insert_stmt):
```
josh-mckenzie: nit: consider rename to '_force_cdc_WriteFailureException' or '_test_cdc_WriteFailureException'?

josh-mckenzie:
Other than the very minor nits/questions above, +1.

ptnapoleon:
```
CASSANDRA_VERSION=git:trunk nose -xsv cdc_test.py:TestCDC.test_cdc_data_available_in_cdc_raw
```
is what fails for me, and continues to not fail for you. Let's see what CI says

mambocab:
Running this on CI now: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/164/

ptnapoleon:
Passes on CI. Must be in my env :shrug:

cdc_test.py
mambocab: True. Honestly, I picked a big number and it was fast enough, so I went with it. Will try tuning it down to a more reasonable number, thanks.

cdc_test.py
mambocab: I haven't, and I'd appreciate that @josh-mckenzie, thanks.

cdc_test.py
mambocab: In my Freakishly Scientific Test of loading speeds (I messed around with the `concurrency` value running locally), I didn't see anything working noticeably faster than 500. Seems fine? Takes about 3s on my machine, doesn't seem worth further effort.

mambocab:
Ok, I'm happy with this. I switched from using flushing to draining in 4f42f73 to guarantee segments get recycled.

Unless there are objections, I'll squash and merge post-weekend and post-Windows-run.

ptnapoleon:
No objections from me

cdc_test.py
ptnapoleon: I think he means, "Why not drop that to ~100, does 500 really provide a benefit"?

josh-mckenzie: Yep. Seems like 500 is massive overkill on # threads to populate. If it takes 3 seconds it's no big deal, more just wondering for future tests, tribal knowledge, etc.

Probably should have some kind of centralized util for stress population that handles concurrency value for us so we don't have to make that decision on a per-test basis.

mambocab:
Hoo, I let this sit, my mistake. @josh-mckenzie, did you get in a couple Windows runs?

josh-mckenzie:
I was a bad panda and did not.

Just ran it, and it's not passing on Win10:

> FAIL: test_cdc_data_available_in_cdc_raw (cdc_test.TestCDC)
> AssertionError: not all expected data selected
> dtest: DEBUG: found 4091 values in CDC table and 1970 values in non-CDC table
I'd be fine just committing it as working on linux and leave it in the pile of "here's things that need TLC on Windows", especially since https://issues.apache.org/jira/browse/CASSANDRA-12148 is basically going to necessitate a rewrite (since we hard-link commit log segments instead of waiting for drain/flush, so no more need to copy / move files, etc).

mambocab:
squashing from 5f17f3c.

mambocab:
Ok, sqashed into 2 logically separate commits. Merging. Thanks to all for shepherding and reviewing.

PR #1026, opened 2016-06-08
Speed up sstableofflinerelevel_test

thobbs:
This takes the run time from ~15m to ~80s.

This needs to be merged at the same time as a related ccm changed.  I'll post the PR link in a sec.

PR #1030, opened 2016-06-10
The API for node run_sstableemetadata has changed

ptnapoleon:
Make the corresponding changes to the dtest code. Previously
we were getting back a list of tuples, now we just have
one tuple

@knifewine to review


ptnapoleon:
This fixes a number of dtest failures across all versions

ptnapoleon:
Dupe of #1026

PR #1034, opened 2016-06-13
save gc & compaction logs on tear down

smccarthy788:
@ptnapoleon: Please review.

Connected with https://github.com/pcmanus/ccm/pull/513

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -488,16 +488,20 @@ class Tester(TestCase):
             name = os.path.join(directory, name)
         if not os.path.exists(directory):
             os.mkdir(directory)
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Can you give me a dtest run with these changes, so I know we don't break existing log capture? It looks like we don't, and the code looks right, but safe > sorry

smccarthy788:
Running: http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/11/

ptnapoleon:
+1

PR #1035, opened 2016-06-13
Check log length is not zero

smccarthy788:
@ptnapoleon: 

Passed cassci build.
http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-cassandra-3.0-dtest/12/


ptnapoleon:
+1

PR #1037, opened 2016-06-14
add proper thread cleanup for active log watching

knifewine:
depends on ccm changes here https://github.com/pcmanus/ccm/pull/515

this makes sure that in cleanup the log watching thread will be joined and allowed to stop.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -3,6 +3,7 @@ from __future__ import with_statement
 import ConfigParser
 import copy
 import errno
+import glob
 import logging
 import os
 import pprint
@@ -17,7 +18,6 @@ import threading
 import time
 import traceback
 import types
-import glob
 from collections import OrderedDict
 from unittest import TestCase
 
@@ -316,6 +316,10 @@ class Tester(TestCase):
         self.var_debug(cluster)
         self.var_trace(cluster)
 
+    def _cleanup_last_test_dir(self):
+        if os.path.exists(LAST_TEST_DIR):
+            os.remove(LAST_TEST_DIR)
+
     def _cleanup_cluster(self):
         if SILENCE_DRIVER_ON_SHUTDOWN:
             # driver logging is very verbose when nodes start going down -- bump up the level
@@ -331,22 +335,24 @@ class Tester(TestCase):
                 self.cluster.stop(gently=True)
 
             # Cleanup everything:
-            debug("removing ccm cluster " + self.cluster.name + " at: " + self.test_path)
-            self.cluster.remove()
+            try:
+                self.stop_active_log_watch()
+            finally:
+                debug("removing ccm cluster {name} at: {path}".format(name=self.cluster.name, path=self.test_path))
+                self.cluster.remove()
 
-            debug("clearing ssl stores from [{0}] directory".format(self.test_path))
-            for filename in ('keystore.jks', 'truststore.jks', 'ccm_node.cer'):
-                try:
-                    os.remove(os.path.join(self.test_path, filename))
-                except OSError as e:
-                    # once we port to py3, which has better reporting for exceptions raised while
-                    # handling other excpetions, we should just assert e.errno == errno.ENOENT
-                    if e.errno != errno.ENOENT:  # ENOENT = no such file or directory
-                        raise
+                debug("clearing ssl stores from [{0}] directory".format(self.test_path))
+                for filename in ('keystore.jks', 'truststore.jks', 'ccm_node.cer'):
+                    try:
+                        os.remove(os.path.join(self.test_path, filename))
+                    except OSError as e:
+                        # once we port to py3, which has better reporting for exceptions raised while
+                        # handling other excpetions, we should just assert e.errno == errno.ENOENT
+                        if e.errno != errno.ENOENT:  # ENOENT = no such file or directory
+                            raise
 
-            os.rmdir(self.test_path)
-        if os.path.exists(LAST_TEST_DIR):
-            os.remove(LAST_TEST_DIR)
+                os.rmdir(self.test_path)
+                self._cleanup_last_test_dir()
 
     def set_node_to_current_version(self, node):
         version = os.environ.get('CASSANDRA_VERSION')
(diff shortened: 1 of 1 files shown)


knifewine:
upgrade test trial here: http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/28/
dtest trial here: http://cassci.datastax.com/view/Dev/view/knifewine/job/trunk_dtest_with_log_watching_cs502/1/

knifewine:
links above updated for newer jobs (upgrade failed mysteriously on first try, dtest had a little bug I fixed).

dtest.py
```diff
@@ -420,12 +421,29 @@ def begin_active_log_watch(self):
         # log watching happens in another thread, but we want it to halt the main
         # thread's execution, which we have to do by registering a signal handler
         signal.signal(signal.SIGINT, self._catch_interrupt)
-        self.cluster.actively_watch_logs_for_error(self._log_error_handler, interval=0.25)
+        # actively_watch_logs_for_error returns an event that needs to be .set() on exit to allow the thread to stop
```
mambocab: `.set` isn't called anywhere -- maybe `.join`?

dtest.py
```diff
@@ -420,12 +421,29 @@ def begin_active_log_watch(self):
...
+        Should not be called if begin_active_log_watch was never called to start log watching.
+        Can be called multiple times without error.
+        If not called, threads will remain running until the parent process exits.
+        """
+        try:
+            self._log_watch_thread.join()
+        except AttributeError:
+            # no thread apparently, so nothing to stop
+            # may happen when allow_log_errors is True, since in that case we don't init a logging thread
+            pass
```
mambocab: It'd be nice if this signaled whether a log watch thread was actually stopped or not. Maybe return `True` if the `try` block completes, and `False` in the `except` block? I'm not sure what the best API is, but I'm assuming raising a new error isn't what we want to do, if that's a common case.

dtest.py
```diff
@@ -305,6 +305,7 @@ def _cleanup_cluster(self):
                 self.cluster.stop(gently=True)
 
             # Cleanup everything:
+            self.stop_active_log_watch()
```
mambocab: How certain are we that `.join` won't raise errors that prevent `self.cluster.remove` from being called below?

mambocab:
Looks reasonable, though I want to talk out my comments before I give a +1, as I'm not very familiar with how this behavior works. I'd be more comfortable if @ptnapoleon had a look as well.

dtest.py
knifewine: oh right, no longer a relevant comment. that happens internally to the ccm thread code now. will remove and update related docstring.

dtest.py
knifewine: Yeah, I don't think another error is good in that case. I expect it to be a rare one, but didn't want to halt a test if the thread doesn't exist and throw a wrench into other activities. True/False seems reasonable, will push up a fix.

knifewine: pushed. I also separated the join, because if that could somehow raise an attribute error (unsure if it woulcd) but wouldn't want that to be caught unintentionally. and added a reasonable timeout for join to take place, since we don't want to hang things up indefinitely.

dtest.py
knifewine: hmm, this is a tricky one. I want the watcher to wrap right after stop, in case there's any final shutdown logs written with errors, so we get a final scan of the written log. I also want things to basically blow up so we fix the problem quickly, but don't want to leave clusters laying around.

knifewine: maybe I try/except, briefly defer if an exception arises and raise after cluster is properly removed?

mambocab: Will `finally:` do the trick?

dtest.py
mambocab: Looks great.

dtest.py
knifewine: ayup, much better idea than what I was going for. patched and pushed.

knifewine:
depends on ccm pr https://github.com/pcmanus/ccm/pull/515

knifewine:
now that new errors can be ignored via https://github.com/pcmanus/ccm/pull/518 and https://github.com/riptano/cassandra-dtest/pull/1056 i've kicked off a new upgrade trial job to test out active log watching against this dtest code and the ccm changes it depends on, over here:

http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/31/

knifewine:
actually, make that http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/32/
since the job flaked once again on openstack

knifewine:
@ptnapoleon ping

ptnapoleon:
Still afraid of the failing job

dtest.py
```diff
@@ -316,13 +316,18 @@ def modify_log(self, cluster):
         self.var_debug(cluster)
         self.var_trace(cluster)
 
+    def _cleanup_last_test_dir(self):
```
ptnapoleon: Jim would say we should make this a function and not a method.

dtest.py
```diff
@@ -331,22 +336,24 @@ def _cleanup_cluster(self):
                 self.cluster.stop(gently=True)
 
             # Cleanup everything:
-            debug("removing ccm cluster " + self.cluster.name + " at: " + self.test_path)
-            self.cluster.remove()
+            try:
+                self.stop_active_log_watch()
+            finally:
+                debug("removing ccm cluster " + self.cluster.name + " at: " + self.test_path)
```
ptnapoleon: String interpolation? For stylistic consistency, if naught else

dtest.py
```diff
@@ -316,13 +316,18 @@ def modify_log(self, cluster):
...
+            os.remove(LAST_TEST_DIR)
+
     def _cleanup_cluster(self):
         if SILENCE_DRIVER_ON_SHUTDOWN:
             # driver logging is very verbose when nodes start going down -- bump up the level
             logging.getLogger('cassandra').setLevel(logging.CRITICAL)
 
         if KEEP_TEST_DIR:
             self.cluster.stop(gently=RECORD_COVERAGE)
+            self._cleanup_last_test_dir()
```
ptnapoleon: Why do we do this here now? Not that it's bad, it just seems orthogonal to the PR and I am not following

dtest.py
```diff
@@ -442,16 +449,39 @@ def begin_active_log_watch(self):
...
-        self.cluster.actively_watch_logs_for_error(self._log_error_handler, interval=0.25)
+        self._log_watch_thread = self.cluster.actively_watch_logs_for_error(self._log_error_handler, interval=0.25)
+
+    def stop_active_log_watch(self):
+        """
+        Joins the log watching thread, which will then exit.
+        Should be called after each test, ideally after nodes are stopped but before cluster files are removed.
+
+        Can be called multiple times without error.
+        If not called, log watchin thread will remain running until the parent process exits.
```
ptnapoleon: nit: `watching`

dtest.py
knifewine: good catch, I don't think that should be there at all. removed and pushed.

dtest.py
knifewine: yeah, though it doesn't rely on self, it still feels conceptually like it should belong with the other class methods to me. don't have a super strong opinion about it. @mambocab ?

dtest.py
knifewine: pushed.

dtest.py
knifewine: thanks, pushed.

knifewine:
all but one thing addressed, I believe.

knifewine:
(though still waiting on my side for a decent run of the upgrade tests against this).

dtest.py
mambocab: shrug emojus

mambocab: I have no idea what `LAST_TEST_DIR` actually does, so I can't speak to what it's like conceptually. Y'all do what you feel.

ptnapoleon: Leave as method then

knifewine:
got a clean run of the upgrade job trial here: http://cassci.datastax.com/view/Upgrades/job/upgrade_tests-all-custom_branch_runs/41/testReport/

I think we're ready to merge. @ptnapoleon @mambocab  ?

ptnapoleon:
I agree, +1

PR #1048, opened 2016-06-20
Use `wait_for_any_log` to wait/grep for logs to avoid test failures caused by buffering.

ifesdjeen:
Multiplexed test can be found [here](https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/129/)


ptnapoleon:
+1

PR #1028, opened 2016-06-09
added 'no-warmup' arg to stress calls

c-kodman:
Is this as expected?

PR #1051, opened 2016-06-21
Tests to reproduce CASSANDRA-11848

pauloricardomg:


bootstrap_test.py
```diff
@@ -132,6 +135,73 @@ def read_from_bootstrapped_node_test(self):
...
+        """
+        Test to guarantee bootstrap will not succeed when there are insufficient replicas
+        @jira_ticket CASSANDRA-11848
+        """
+        cluster = self.cluster
+        cluster.populate(3)
+        cluster.start()
+        node1, node2, node3 = cluster.nodelist()
+
+        node1.stress(['write', 'n=10K', '-rate', 'threads=8', '-schema', 'replication(factor={})'.format(rf)])
```
mambocab: We recently changed the way we call stress for our tests to make the data loading faster:

https://github.com/riptano/cassandra-dtest/pull/1028/files

Could you please add `no-warmup` to all stress calls in this PR?

bootstrap_test.py
```diff
@@ -132,6 +135,73 @@ def read_from_bootstrapped_node_test(self):
...
+
+        # on 2.2+ change system_auth rf to 2 so it doesn't emit unsufficient sources for streaming warning
```
mambocab: do you mean `sufficient` here?

bootstrap_test.py
```diff
@@ -132,6 +135,73 @@ def read_from_bootstrapped_node_test(self):
...
+                ALTER KEYSPACE system_auth
+                    WITH replication = {'class':'SimpleStrategy', 'replication_factor':2};
+            """)
+
+        # Stop remaining nodes
+        node2.stop(wait_other_notice=True)
+        if stop_two_replicas:
+            node3.stop(wait_other_notice=True)
+
+        should_bootstrap = not consistent_range_movement and not stop_two_replicas
```
mambocab: If this is an expectation, a better name might be `successful_bootstrap_expected`.

bootstrap_test.py
```diff
@@ -34,6 +34,11 @@ def __init__(self, *args, **kwargs):
         Tester.__init__(self, *args, **kwargs)
         self.allow_log_errors = True
 
+    def assert_bootstrap_state(self, node, state):
```
mambocab: I'm tempted to ask you to pull this out into a top-level function and make it take a `session` instead of a `node`.

EDIT: Oh, I see, we have to get an exclusive connection. Could you make a comment noting that the exclusivity is important so that I don't rewrite this to take a `session` in the future? :stuck_out_tongue:

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
         debug("Verifying querying works again.")
         finalData = list(session.execute(query))
         self.assertListEqual(initialData, finalData)
+
+    def replace_test_with_insufficient_replicas(self):
+        """
+        Test to reproduce CASSANDRA-11848
```
mambocab: Could you please add a `@jira_ticket` annotation to this docstring?

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+        cluster.populate(3).start()
+        node1, node2, node3 = cluster.nodelist()
+
+        if DISABLE_VNODES:
+            numNodes = 1
+        else:
+            # a little hacky but grep_log returns the whole line...
+            numNodes = int(node3.get_conf_option('num_tokens'))
+
+        debug(numNodes)
```
mambocab: Could you make this debug statement a little more verbose? Right now it's just going to print a number and that's not very useful for us.

mambocab: Also, could you change the name of this variable -- camelcase isn't idiomatic Python. (I know a lot of assertion methods use camelcase; they were written before these conventions existed.)

mambocab: Sorry for so many comments on this variable's name, but: the term `node` is already being used. I think `num_tokens` would be a better name for this variable.

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+        node3.stop(wait_other_notice=True)
+
+        # stop other replica
+        debug("Stopping node2 (other replica)")
+        node2.stop(wait_other_notice=True)
+
+        # replace node 3 with node 4
+        debug("Starting node 4 to replace node 3")
+
+        node4 = Node('node4', cluster, True, ('127.0.0.4', 9160), ('127.0.0.4', 7000), '7400', '0', None, binary_interface=('127.0.0.4', 9042))
```
mambocab: This signature is hard to read. Could you please pass these arguments as keyword args? I think it would look like

```python
Node('node4', cluster=cluster, auto_bootstrap=True, thrift_interface=('127.0.0.4', 9160), storage_interface=('127.0.0.4', 7000), jmx_port='7400', remote_debug_port='0', initial_token=None, binary_interface=('127.0.0.4', 9042))
```

bootstrap_test.py
```diff
@@ -132,6 +135,73 @@ def read_from_bootstrapped_node_test(self):
         new_rows = list(session.execute("SELECT * FROM %s" % (stress_table,)))
         self.assertEquals(original_rows, new_rows)
 
+    def check_not_running(self, node):
```
mambocab: This doesn't need to be a method; it can be a top-level function. Just remove the `self` parameter and replace `self.assertFalse` with `nose.tools.assert_equal`.

Since this asserts, it should be called `assert_not_running`.

@ptnapoleon @knifewine Once this gets merged we might make the timeout length a parameter and move this to the `assertions` module, yeah?

bootstrap_test.py
```diff
@@ -132,6 +135,73 @@ def read_from_bootstrapped_node_test(self):
...
+    def consistent_range_movement_true_with_one_replica_down_should_fail_test(self):
+        self._bootstrap_test_with_replica_down(True)
+
+    def consistent_range_movement_false_with_one_replica_down_should_succeed_test(self):
+        self._bootstrap_test_with_replica_down(False)
+
+    def consistent_range_movement_false_with_two_replicas_down_should_fail_test(self):
+        self._bootstrap_test_with_replica_down(False, stop_two_replicas=True)
+
+    def consistent_range_movement_true_with_ks_rf1_shoud_fail_test(self):
```
mambocab: `s/shoud/should/`

bootstrap_test.py
```diff
@@ -132,6 +135,73 @@ def read_from_bootstrapped_node_test(self):
...
+        self._bootstrap_test_with_replica_down(True, rf=1)
+
+    def consistent_range_movement_true_with_ks_rf1_shoud_succeed_test(self):
```
mambocab: `s/shoud/should/`

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
         debug("Verifying querying works again.")
         finalData = list(session.execute(query))
         self.assertListEqual(initialData, finalData)
+
+    def replace_test_with_insufficient_replicas(self):
```
mambocab: The name's a little awkward. Could you please change it to `replace_with_insufficient_replicas_test`?

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+
+        node4 = Node('node4', cluster, True, ('127.0.0.4', 9160), ('127.0.0.4', 7000), '7400', '0', None, binary_interface=('127.0.0.4', 9042))
+        cluster.add(node4, False)
+        node4.start(replace_address='127.0.0.3', wait_for_binary_proto=False, wait_other_notice=False)
+
+        # replace should fail due to insufficient replicas
+        node4.watch_log_for("Unable to find sufficient sources for streaming range")
+        self.check_not_running(node4)
+
+    def multi_dc_replace_with_rf1(self):
```
mambocab: This method won't run because `test` isn't in the name. Could you append `_test` to it?

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+        # stop node to replace
+        debug("Stopping node to replace.")
+        node2.stop(wait_other_notice=True)
+
+        node3 = new_node(cluster, data_center='dc2')
+        node3.start(replace_address='127.0.0.2', wait_for_binary_proto=True)
+
+        session = self.exclusive_cql_connection(node3)
+        rows = list(session.execute("SELECT bootstrapped FROM system.local WHERE key='local'"))
+        assert len(rows) == 1
```
mambocab: @c-kodman @smccarthy788 Could you make a note to rewrite this with the new length assertion once both this and your assertions PR are merged?

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+        assert rows[0][0] == 'COMPLETED', rows[0][0]
```
mambocab: Could you rewrite this to use your `assert_bootstrap_state` method?

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+
+        # Check that keyspace was replicated from dc1 to dc2
+        self.assertFalse(node3.grep_log("Unable to find sufficient sources for streaming range"))
+
+        # query should work again with node1 stopped
+        node1.stop(wait_other_notice=True)
+        debug("Verifying querying works again.")
+        query = SimpleStatement('select * from %s LIMIT 1' % stress_table, consistency_level=ConsistencyLevel.LOCAL_ONE)
+        finalData = list(session.execute(query))
+        self.assertListEqual(initialData, finalData)
```
mambocab: I think these 3 lines can be condensed into one with [`assertions.assert_all`](https://github.com/riptano/cassandra-dtest/blob/master/assertions.py#L61):

```python
assert_all(session, 'SELECT * from {} LIMIT 1'.format(stress_table), expected=initialData, cl=ConsistencyLevel.LOCAL_ONE)
```

replace_address_test.py
```diff
@@ -456,3 +457,117 @@ def replace_with_reset_resume_state_test(self):
...
+          partitions: fixed(1)
+          batchtype: UNLOGGED
+        queries:
+          read:
+            cql: select * from users where username = ?
+            fields: samerow
+        """
+        stress_config = tempfile.NamedTemporaryFile(mode='w+', delete=False)
+        stress_config.write(yaml_config)
+        stress_config.close()
```
mambocab: I'm not positive, but I think you can remove the `delete=False` and move this `.close` call after it's used in the `stress` call. Better still:

```python
with tempfile.NamedTemporaryFile(mode='w+') as stress_config:
    stress_config.write(yaml_config)
    node1.stress(...)
```

and the file should be closed and deleted when the `with` block closes. Am I right @knifewine / @ptnapoleon?

mambocab:
I'm still working on running all these locally, but I can tell you for sure that `bootstrap_test.py:TestBootstrap.consistent_range_movement_false_with_two_replicas_down_should_fail_test` fails on `trunk` locally for me.

pauloricardomg:
Addressed comments and pushed updated patch.

> I'm still working on running all these locally, but I can tell you for sure that `bootstrap_test.py:TestBootstrap.consistent_range_movement_false_with_two_replicas_down_should_fail_test` fails on `trunk` locally for me.

Can you double check it's not working? What is the error? It's working locally for me.  In any case, I submitted a new CI run: http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/8/ (still queued)

ptnapoleon:
Bumping to make sure this is on @mambocab's queue

pauloricardomg:
I'm going on vacation next week so it would be nice to have this reviewed until the end of this week so it doesn't atrophy until I'm back. (I understand you guys are pretty busy this week, so this is just a friendly heads up, not a huge deal if this slips a few more weeks)

mambocab:
Thanks for the pings, both of you. Reviewing now!

replace_address_test.py
```diff
@@ -452,11 +478,130 @@ def replace_with_reset_resume_state_test(self):
...
+        if DISABLE_VNODES:
+            num_tokens = 1
+        else:
+            # a little hacky but grep_log returns the whole line...
+            num_tokens = int(node3.get_conf_option('num_tokens'))
+
+        debug("testing with num_tokens: {}".format(num_tokens))
+
+        debug("Inserting Data...")
+        node1.stress(['write', 'n=10K', '-schema', 'replication(factor=2)'])
```
mambocab: Should this have `no-warmup`? If it should, could you add it, and otherwise, could you leave a comment?

replace_address_test.py
```diff
@@ -452,11 +478,130 @@ def replace_with_reset_resume_state_test(self):
...
+          batchtype: UNLOGGED
+        queries:
+          read:
+            cql: select * from users where username = ?
+            fields: samerow
+        """
+        with tempfile.NamedTemporaryFile(mode='w+') as stress_config:
+            stress_config.write(yaml_config)
+            stress_config.flush()
+            node1.stress(['user', 'profile=' + stress_config.name, 'n=10k',
```
mambocab: Same as above -- could you either add `no-warmup` here or make a comment as to why we don't?

replace_address_test.py
```diff
@@ -452,11 +478,130 @@ def replace_with_reset_resume_state_test(self):
...
+                          'ops(insert=1)', '-rate', 'threads=50'])
+
+        session = self.patient_cql_connection(node1)
+
+        # on 2.2+ change system_auth rf to 2 so it doesn't emit unsufficient sources for streaming warning
```
mambocab: This still confuses me. Do you mean `so it emits enough sources for streaming warnings`?

assertions.py
```diff
@@ -102,3 +106,17 @@ def assert_crc_check_chance_equal(session, table, expected, ks="ks", view=False)
                    "SELECT crc_check_chance from system_schema.tables WHERE keyspace_name = 'ks' AND "
                    "table_name = '{table}';".format(table=table),
                    [expected])
+
+
+def assert_bootstrap_state(exclusive_session, state):
```
mambocab: (continuation of https://github.com/riptano/cassandra-dtest/pull/1051#discussion_r68060887)

Sorry, I wasn't clear in my last review -- I meant that I'd like to see this left as a method, with a comment that this requires an exclusive connection. My first thought about it was incorrect.

What happens when we use a non-exclusive session? I'm worried that we'll use it wrong and silently get bad results. Is there a way to assert that the session argument is actually exclusive? If not, I think the old implementation is safer. Sorry for being confusing.

mambocab:
Thanks for your patience, @pauloricardomg. There are a couple spots where it looks like you missed adding `no-warmup` to stress calls, there's a comment I'm confused about, and I was unclear about what I wanted done with the assertion method. Unless we can enforce the constraint of using an exclusive connection, I think it should be moved back to where it was.

Go ahead and address those, then feel free to merge!

pauloricardomg:
Addressed review comments, and moved `assert_bootstrap_state` to `bootstrap_test.py`, but keeping it as a static method receiving a `dtest.Tester` object where it can fetch the exclusive connection from, so it can be reused by `replace_address_test.py`.

I submitted a [dtest run](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/20/testReport/) but `replace_nonexistent_node_test` failed with a missing import, so I added the missing import and resubmitted the [dtest run](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/22/).

assertions.py
```diff
@@ -240,3 +241,16 @@ def assert_length_equal(object_with_length, expected_length):
     assert_equal(len(object_with_length), expected_length,
                  "Expected {} to have length {}, but instead is of length {}".format(object_with_length,
                                                                                      expected_length, len(object_with_length)))
+
+
+def assert_not_running(node):
```
ptnapoleon: Beautiful

bootstrap_test.py
```diff
@@ -315,24 +380,22 @@ def local_quorum_bootstrap_test(self):
...
         """
-        stress_config = tempfile.NamedTemporaryFile(mode='w+', delete=False)
-        stress_config.write(yaml_config)
-        stress_config.close()
-        node1.stress(['user', 'profile=' + stress_config.name, 'n=2000000',
-                      'ops(insert=1)', '-rate', 'threads=50'])
+        with tempfile.NamedTemporaryFile(mode='w+') as stress_config:
+            stress_config.write(yaml_config)
+            stress_config.flush()
+            node1.stress(['user', 'profile=' + stress_config.name, 'n=2000000', 'no-warmup',
```
ptnapoleon: Can you make this `n=2M`, please?

ptnapoleon:
+1 after nit, this looks great.

mambocab:
Yep, I'm +1 too.

yukim:
+1

mambocab:
Great -- if you still think it's ready, feel free to merge @pauloricardomg, thanks!

pauloricardomg:
Addressed nit and submitted dtest run: http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/23/ (will commit if it looks good)

pauloricardomg:
dtests [look good](http://cassci.datastax.com/job/pauloricardomg-trunkcopy-dtest/23/testReport/), merging PR.

PR #1052, opened 2016-06-22
restricted regex to 2.2+

smccarthy788:
@thobbs, please review. Looking at CASSANDRA-11598, it seems we version gated the wrong searches.


thobbs:
+1

PR #1053, opened 2016-06-22
Only run test from CASSANDRA-11038 on 2.2+

beobal:
Since 2.1 is critical fixes only, the patch to ensure correct behaviour for 11038 was only added from 2.2 onwards, so it's probably wise to only run it's test on those versions.

(see also: CASSANDRA-12056)


smccarthy788:
+1

PR #1058, opened 2016-06-23
Improve startup time by prelocating libjemalloc

thobbs:
This cuts about 1s off of Cassandra's startup time, which should save us ~15m on the dtests.

This requires https://github.com/pcmanus/ccm/pull/519 in order to work
correctly.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -115,6 +115,26 @@ else:
     CASSANDRA_VERSION_FROM_BUILD = get_version_from_build(CASSANDRA_DIR)
 
 
+# Determine the location of the libjemalloc jar so that we can specify it
+# through environment variables when start Cassandra.  This reduces startup
+# time, making the dtests run faster.
+def find_libjemalloc():
+    if is_win():
+        # let the normal bat script handle finding libjemalloc
+        return ""
+
+    this_dir = os.path.dirname(os.path.realpath(__file__))
+    script = os.path.join(this_dir, "findlibjemalloc.sh")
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

dtest.py
```diff
@@ -115,6 +115,26 @@
...
+# Determine the location of the libjemalloc jar so that we can specify it
+# through environment variables when start Cassandra.  This reduces startup
+# time, making the dtests run faster.
+def find_libjemalloc():
+    if is_win():
+        # let the normal bat script handle finding libjemalloc
+        return ""
+
+    this_dir = os.path.dirname(os.path.realpath(__file__))
+    script = os.path.join(this_dir, "findlibjemalloc.sh")
```
ptnapoleon: Did you need to commit this script?

dtest.py
```diff
@@ -115,6 +115,26 @@
...
+    p = subprocess.Popen([script], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout, stderr = p.communicate()
+    if stderr or not stdout:
+        return "-"  # tells C* not to look for libjemalloc
```
ptnapoleon: The error handling here is less than perfect, as if this fails for certain reasons, it 100% explodes everywhere

thobbs:
Damnit, forgot to include a new script file in the commit.  I'll open a new pull request.

dtest.py
thobbs: I'll make that a bit more robust.

PR #1060, opened 2016-06-23
Improve startup time by prelocating libjemalloc

thobbs:
This requires https://github.com/pcmanus/ccm/pull/519 in order to work
correctly.

This should fix the problems with PR #1058


ptnapoleon:
Doesn't break everything, +1

PR #1061, opened 2016-06-24
Use long message to provide more information about failed query.

ifesdjeen:
The patch is related to [CASSANDRA-12025](https://issues.apache.org/jira/browse/CASSANDRA-12025).

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -1088,31 +1088,47 @@ class TestPagingData(BasePagingTester, PageAssertionMixin):
             for j in xrange(10):
                 session.execute("UPDATE test SET cnt = cnt + {} WHERE a={} AND b={} AND c={}".format(j + 2, i, j, j + 1))
 
+        self.longMessage = True
         for page_size in (2, 3, 4, 5, 7, 10, 20):
             session.default_fetch_size = page_size
+            page_size_error_msg = "Query failed with page size {}".format(page_size)
 
             # single partition
-            assert_all(session, "SELECT * FROM test WHERE a = 4 AND b > 3 AND c > 3 AND cnt > 8 ALLOW FILTERING", [[4, 7, 8, 9],
-                                                                                                                   [4, 8, 9, 10],
-                                                                                                                   [4, 9, 10, 11]])
+            res = rows_to_list(session.execute("SELECT * FROM test WHERE a = 4 AND b > 3 AND c > 3 AND cnt > 8 ALLOW FILTERING"))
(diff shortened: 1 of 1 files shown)


mambocab:
Could you show us the difference between these failures with and without the `self.longMessage = True`? This is the first I've seen it, and [it seems to be `True` by default](https://github.com/python/cpython/blob/2410a06c294d9b091f39f09f638c1d83d0ff440e/Lib/unittest/case.py#L371). From what I read in the `unittest` docs, it looks like `longMessage` just makes failred assertions print the `msg` parameter, which they do by default.

ptnapoleon:
Bump, why set longMessage to True?

ifesdjeen:
The output without `longMessage: True` is: 

```
AssertionError: Query failed with page size 2
```

With:

```
AssertionError: Lists differ: [[4, 7, 8, 9], [4, 8, 9, 10], ... != [[4, 7, 8, 91], [4, 8, 9, 10],...

First differing element 0:
[4, 7, 8, 9]
[4, 7, 8, 91]

- [[4, 7, 8, 9], [4, 8, 9, 10], [4, 9, 10, 11]]
+ [[4, 7, 8, 91], [4, 8, 9, 10], [4, 9, 10, 11]]
?             +
 : Query failed with page size 2
```

I'm not sure why it's not set to True for me by default though...

mambocab:
Whoa, I have no idea where that gets set to `False`, but it sure does:

```python
~/scratch  
(dtest) ❯ cat test_longMessage.py 
import unittest


class TestLongMessage(unittest.TestCase):
    def test_longMessage(self):
        assert self.longMessage
                                                                                                                                                                                                                                        
~/scratch  
(dtest) ❯ nosetests test_longMessage.py 
F
FAIL: test_longMessage (test_longMessage.TestLongMessage)
  File "/home/mambocab/tmp/scratch-1467209909/test_longMessage.py", line 6, in test_longMessage
    assert self.longMessage
AssertionError

Ran 1 test in 0.000s

FAILED (failures=1)
```

But that looks good. You can feel free to set it to `True` for this test, as that message is more useful. You don't have to set it back to `False`; there's a new `Tester` object created for every test, so it won't affect the next test run.

ifesdjeen:
Great, thank you!

mambocab:
@ptnapoleon, any objections? LGTM.

mambocab:
Eh, I'm sure Philip won't mind it, the change is fine. Merging.

mambocab:
Thanks @ifesdjeen

PR #1062, opened 2016-06-27
corrected known failure problem

c-kodman:


ptnapoleon:
Travis still isn't happy, which is confusing, and I don't see where you removed the bad annotation?

PR #902, opened 2016-03-30
Add bootstrap test with -Dcassandra.write_survey and nodetool join

pauloricardomg:
Test for CASSANDRA-11379

PR #1064, opened 2016-06-28
Add bootstrap test with -Dcassandra.write_survey and nodetool join

ptnapoleon:
Replacing #902 with this

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -14,7 +14,7 @@ from ccmlib.node import NodeError
 from assertions import assert_almost_equal, assert_one
 from dtest import Tester, debug
 from tools import (InterruptBootstrap, KillOnBootstrap, known_failure,
-                   new_node, query_c1c2, since)
+                   new_node, no_vnodes, query_c1c2, since)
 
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Rebased, now addressing nits

ptnapoleon:
@pauloricardomg and @mambocab can you review

pauloricardomg:
+1, I think this can be commited since it was already reviewed before and nits addressed.

but I really don't know why we don't enable these bootstrap tests for vnodes, so I created this [branch](https://github.com/riptano/cassandra-dtest/tree/vnodes_bootstrap) enabling these tests for vnodes and will run them through the multiplexer to see if some flakiness will arise due to the data size comparisons.

pauloricardomg:
submitted here: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/152/

pauloricardomg:
hmm, multiplexer results didn't look very good, so there's a reason why this is enable only for `@no_vnodes`, so let's keep it the way it is for now (will maybe try to refactor later)
+1

PR #1065, opened 2016-06-28
Create ReusableClusterTester, use on thrift_tests.py

thobbs:
The ``ReusableClusterTester`` reuses the same cluster across multiple test
methods.  I picked ``thrift_tests.py`` to start with because it has a large
number of small, simple tests.  These changes reduce the run time for
thrift_tests from 28m to 1m.

These changes should **NOT** be merged in until cassci buckets tests by class (instead of by method).  I'm opening the pull request now for early review.    If we're in favor of doing this, I can start working on converting other similar test modules to be reusable.


ptnapoleon:
Have you tested that this properly tears down if only one test in the suite is run? In which case, why not merge prior to the cassci portion?

thobbs:
Hmm, good point.  It does work just fine if only a single test is run, so I suppose we could merge before the cassci bucketing changes are in.  It just wouldn't be any faster :)

thobbs:
Looking at some of the other long-running modules, I think ``auth_test.py``, ``paging_test.py``, and ``cqlsh_tests/cqlsh_tests.py`` are good candidates to convert next.

mambocab:
Reading through these changes now.

knifewine:
looks nice. a couple thoughts. given the plumbing changes, I'd like to see this code run on full dtest and upgrade jobs, so we can compare to existing and make sure no new issues arise.

seems like we could possibly get some good use out of python context managers in some of the tests, to have a nice syntax for "hey I'm using this table, but clean it up when this block exits". we could possibly create a context manager that uses ccm's data wipe feature (forget what it's called atm).

dtest.py
```diff
@@ -907,6 +756,263 @@ def dump_jfr_recording(self, nodes):
...
+            try:
+                # save the logs for inspection
+                if failed or KEEP_LOGS:
+                    self.copy_logs(self.cluster)
+            except Exception as e:
+                print "Error saving log:", str(e)
+            finally:
+                if failed:
+                    cleanup_cluster(self.cluster, self.test_path)
+                    reset_environment_vars()
```
mambocab: Shouldn't `reset_environment_vars` be called on success as well?

dtest.py
```diff
@@ -907,6 +756,263 @@ def dump_jfr_recording(self, nodes):
...
+                    self.__class__.setUpClass()
```
mambocab: I'm trying to figure out whether this should be moved or not. The alternative I'm thinking of is something like setting `self.__class__.class_setup_required = True`, and then in `setUp`, run

```python
if self.__class__.class_setup_required:
    self.__class__.setUpClass()  # or, pull out the setup logic and use it both here and in `setUpClass`
    self.__class__.class_setup_required = False
```

On one hand, I don't really like the idea of communicating between methods by setting random attributes on the class -- we deal with enough of that crap -- but on the other, I also don't like the idea of `tearDown` calling a setup method sometimes.

thrift_tests.py
```diff
@@ -40,66 +40,67 @@ def pid():
...
+        cls.client = get_thrift_client()
+        cls.client.transport.open()
+        cls.define_schema()
+
+    def setUp(self):
+        ReusableClusterTester.setUp(self)
 
-class ThriftTester(BaseTester):
-    client = thrift_client
+        # this is ugly, but the whole test module is written against a global client
```
mambocab: Is it? Attribute lookup should allow methods that are passed `self` to just access `ThriftTester.client` as `self.client`.

thrift_tests.py
```diff
@@ -1251,6 +1231,8 @@ def test_range_with_remove_cf(self):
 
     def test_range_collation(self):
         _set_keyspace('Keyspace1')
+        client.truncate('Standard1')
```
mambocab: Any reason not to add all these `truncate`s to this class's `tearDown`?

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -254,7 +254,7 @@ def setUp(self):
         debug("Versions to test (%s): %s" % (type(self), str([v.version for v in self.test_version_metas])))
 
     def init_config(self):
-        self.init_default_config()
+        Tester.init_config(self)
```
mambocab: Should this be using `super`?

thrift_tests.py
```diff
@@ -1954,211 +1972,201 @@ def req():
 
     def test_incr_decr_standard_add(self):
         _set_keyspace('Keyspace1')
+        key = 'test_incr_decr_standard_add'
```
mambocab: Fun `unittest.TestCase` fact: you can just use `key = self._testMethodName` here, if that's all you're going for.

thrift_tests.py
```diff
@@ -1954,211 +1972,201 @@ def req():
...
-        client.add('key1', ColumnParent(column_family='Counter1'), CounterColumn('c1', d1), ConsistencyLevel.ONE)
-        time.sleep(5)
-        rv1 = client.get('key1', ColumnPath(column_family='Counter1', column='c1'), ConsistencyLevel.ONE)
+        client.add(key1, ColumnParent(column_family='Counter1'), CounterColumn('c1', d1), ConsistencyLevel.ONE)
+        rv1 = client.get(key1, ColumnPath(column_family='Counter1', column='c1'), ConsistencyLevel.ONE)
         assert rv1.counter_column.value == d1
 
         # remove the previous column and check that it is gone
-        client.remove_counter('key1', ColumnPath(column_family='Counter1', column='c1'), ConsistencyLevel.ONE)
-        time.sleep(5)
```
mambocab: What's the scoop on removing all these `sleep`s?

mambocab:
Done with my review. Definitely some great changes in there. The core of it, though -- making the cluster-reusing `Tester` subclass -- is hard to follow, because the original code is hard to follow. Like Russ, I'd like to see some test runs before +1ing this.

dtest.py
thobbs: My thought was that it shouldn't matter if the cluster isn't restarted, but you're right, it's better to be safe about this.  Something like cqlsh might get set up with different env vars.

dtest.py
thobbs: I could make it a bit more explicit by having a third classmethod like, say, ``setup_environment()``.  This would contain most of what's in ``setUpClass()``, so both ``setUpClass()`` and ``tearDown()`` (after failure) would call it.  Docs on that method could make it pretty clear what it's used for.

thrift_tests.py
thobbs: Yes, but there are a bunch of helper methods that rely on ``client`` being global.  Plus, changing this would change almost every line in the module, so I figured it wasn't worth the messy diff.  I could definitely fix it in a follow up commit, if it's too ugly to tolerate :)

thrift_tests.py
thobbs: Just that truncating a dozen column families is a little slow.

upgrade_tests/upgrade_through_versions_test.py
thobbs: Where possible, I avoid using ``super`` for the reasons listed here: https://fuhm.net/super-harmful/

thrift_tests.py
thobbs: Yeah, I should probably switch to that.

Also a fun fact: my most upvoted StackOverflow answer is (bizarrely) about this very topic: http://stackoverflow.com/questions/4504622/a-way-to-output-pyunit-test-name-in-setup/4506296#4506296

thrift_tests.py
thobbs: I honestly cannot think of a reason why they should have been there in the first place.  Maybe they were a misguided effort to fix a broken test?

thobbs:
> The core of it, though -- making the cluster-reusing Tester subclass -- is hard to follow, because the original code is hard to follow

Yes, the code was a bit crazy.  I do think that splitting most of the logic in ``setUp`` and ``tearDown`` into distinct methods makes it clearer, so hopefully the changes are neutral overall.  I considered going with the ``_preserve_cluster`` attribute style of making this work, but I figured that having a bunch of ``if`` statements that completely changed the behavior would be extremely messy.  Hopefully with all of the special behavior encapsulated in a separate class, it should be easier to pick up.

> Like Russ, I'd like to see some test runs before +1ing this.

Definitely, I'll start some.

thobbs:
I've pushed a commit to address the review comments.

> seems like we could possibly get some good use out of python context managers in some of the tests, to have a nice syntax for "hey I'm using this table, but clean it up when this block exits". we could possibly create a context manager that uses ccm's data wipe feature (forget what it's called atm)

I have a bit of a preference for "defensive" cleanup, where each test ensures that it has the clean state that it needs to run correctly.  If you rely on other tests cleaning up properly after themselves, it can lead to more confusing test failures.  So, if we went with a context manager, we would probably want it to clean things up in the ``__enter__`` phase rather than the ``__exit__`` phase.  I think this behavior would be a little surprising (since context managers are generally used to clean up a resource when you're done with it), so for now I stuck with a more explicit truncate at the start of each test.  I did make a helper method to make truncating multiple tables cleaner.

If you had something else in mind, or see a way to further improve this, let me know.

I've also started some test runs:
http://cassci.datastax.com/view/Dev/view/thobbs/job/thobbs-cassandra-2.1-dtest/
http://cassci.datastax.com/view/Dev/view/thobbs/job/thobbs-cassandra-2.2-dtest/
http://cassci.datastax.com/view/Dev/view/thobbs/job/thobbs-cassandra-3.0-dtest/
http://cassci.datastax.com/view/Dev/view/thobbs/job/thobbs-trunk-dummy-dtest/

thrift_tests.py
mambocab: :+1:

thrift_tests.py
mambocab: Nope, this works -- didn't know about the dependency on the `global` client. These tests rarely see changes, so factoring that out probably isn't worth much effort. Thanks for the explanation.

dtest.py
mambocab: I like the change you made :+1:

thrift_tests.py
mambocab: Fair enough -- we'll see if anything gets flaky in test runs.

mambocab:
Seeing some failures in the dtests. @ptnapoleon can sift through through those faster than I can, and he should give this a review anyway, so he says he'll do that later today.

I'm +1 if the changes didn't break anything on those dtests runs.

ptnapoleon:
Can you rebase and then re-run the tests? There are a lot of failures that I think rebasing will help, but I've never seen some of them before, and I'm unsure if this PR causes them or not

thobbs:
I believe the ``snapshot_test`` failures are caused by this.  I pushed one commit to try to fix them, but it looks like there are still problems.  I'll let you know once I've sorted that out.

thrift_tests.py
mambocab: that is the funnest fact I've heard all day

thobbs:
The test runs on the rebased branch are complete.  The 2.x and 3.0 test runs look good to me.  The trunk run has a couple of failures that I'm less sure about.

The {{json_test}} that failed also [failed in a similar way recently](http://cassci.datastax.com/job/trunk_dtest/1295/testReport/json_test/FromJsonInsertTests/complex_data_types_test/) on the trunk dtests.

The internode_ssl test also appears to have [flaked recently](http://cassci.datastax.com/job/trunk_dtest/1297/testReport/internode_ssl_test/TestInternodeSSL/putget_with_internode_ssl_test/history/).

The cqlsh copy test failure looks like it might be caused by the ongoing netty issues.

So, from my perspective I think this is good to go.  @ptnapoleon can you give those test results a look and see if anything needs to be investigated?

ptnapoleon:
Good to go! +1

thobbs:
Great, merged.  Thanks for the reviews and feedback!

PR #1067, opened 2016-06-29
Fix broken trunk dtests due to version change to 3.10

pauloricardomg:
We'll probably need to fix the ccm/dtest versioning scheme to be more resilient and use `distutils.version.LooseVersion` exclusively, but this is a more immediate fix to fix trunk dtests more promptly.

Submitted dtest run: http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy2-dtest/4/

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -9,6 +9,7 @@ import time
 from assertions import assert_none, assert_one
 from ccmlib import common
 from dtest import Tester, debug
+from distutils.version import LooseVersion
 from nose.tools import assert_equal
 from tools import known_failure, since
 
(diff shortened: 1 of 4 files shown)


ptnapoleon:
+1

ptnapoleon:
It's in poor taste, but I'm going to merge this, and then mop up any problems CI shows after. I'm sure we need these changes ASAP, and I'm confident they're correct as-is.

mambocab:
I don't know about the general case, but looking at these changes, we should be able to change `Cluster.version()` to return a `LooseVersion` and revert this PR.

pauloricardomg:
@mambocab if this works I'm totally +1 with this cleaner alternative approach, as @ptnapoleon pointed out this was more of dirty workaround to return CI to a healthy state until we can think of something better.

ptnapoleon:
The cleaner approach is a disaster so far, in testing. But this CI looked great, so this PR is done

PR #1071, opened 2016-06-30
added another known failure

c-kodman:
my note isn't a standard one, but was couldn't think of anything better than what the error was.  I expect you'll tell me to change it?


mambocab:
Hmm. Not actually a bad idea to include a little information about the expected failure in the notes. My thought: add the version or the job that it fails on and merge it. @ptnapoleon, thoughts on that?

ptnapoleon:
I would in the general case, but this will take like two seconds to fix. I just wanted a ticket so we didn't forget about it. You and I broke it. +1, let's just merge now

PR #1073, opened 2016-07-01
Fix repair_tests.repair_test.TestRepair.repair_after_upgrade_test (CASSANDRA-12057)

pauloricardomg:
Main changes are:
* Upgrade from 2.2.5 instead of 2.1.9 to avoid windows incompatibility and other minor bugs
* Set dclocal_read_repair_chance when read_repair is specified on dtest.create_cf
* Move repair upgrade test to upgrade suite and only run when UPGRADE_TEST_RUN=true

Multiplexer 100x run: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/157/ (the only failure is due to CASSANDRA-11393)
Dtest trunk run: http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/19/

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -701,7 +701,7 @@ class Tester(TestCase):
             query += ' AND compression = {}'
 
         if read_repair is not None:
-            query = '%s AND read_repair_chance=%f' % (query, read_repair)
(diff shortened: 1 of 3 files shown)


ptnapoleon:
+1

PR #1075, opened 2016-07-01
Combine short_read_test and short_read_reversed_test

thobbs:
These two tests do basically the same thing, but one of them focuses on
reversed reads.  Since each test takes ~10m to run, I figured it's worth
it to combine them.


consistency_test.py
```diff
@@ -767,29 +765,53 @@ def short_read_test(self):
...
             res = rows
             assert_length_equal(res, 3)
 
             # value 0, 1 and 2 have been deleted
             for i in xrange(1, 4):
-                self.assertEqual(res[i - 1][1], 'value{}'.format(i + 2), 'Expecting value{}, got {} ({})'.format(i + 2, res[i - 1][1], str(res)))
+                expected = 'value{}'.format(i + 2)
+                actual = res[i - 1][1]
+                message = 'Expecting value{}, got {} ({})'.format(i + 2, res[i - 1][1], str(res))
+                self.assertEqual(expected, actual, message)
```
mambocab: I think that this whole block could be replaced with

```python
self.assertEqual('value{}'.format(i + 2), res[i - 1][1])
```

and `assertEqual` would take care of the message. Not sure why the original version calcluated its own message.

Same goes for the assertion block below.

consistency_test.py
```diff
@@ -767,29 +765,53 @@ def short_read_test(self):
...
-            insert_columns(self, session, 0, 9)
+            # insert 9 columns in two rows
+            insert_columns(self, session, normal_key, 9)
+            insert_columns(self, session, reversed_key, 9)
 
             # Deleting 3 first columns with a different node dead each time
-            self.stop_delete_and_restart(1, 0)
-            self.stop_delete_and_restart(2, 1)
-            self.stop_delete_and_restart(3, 2)
+            self.stop_delete_and_restart(1, normal_key=normal_key, normal_column=0, reversed_key=reversed_key, reversed_column=6)
```
mambocab: The new call signature is a little cumbersome. It'd look cleaner to just add a `key` parameter, rather than new 4 you added:

```python
for key, node_number, column in itertools.product((normal_key, reversed_key), ((1, 6), (2, 7), (3, 8))):
    stop_delete_and_restart(self, key=key, node_number=node_number, column=column)
```

I bounced this off Philip, who noted that you might have changed it to do fewer stops and restarts. If that's the case -- if you did this for performance -- could you either document that, or pull out the stop-and-restart logic into separate methods or a context manager or something?

thobbs:
Okay, I've tried to address your review comments in the above commit.  Thanks for the feedback.

> you might have changed it to do fewer stops and restarts

Yes, the 10 restarts take up the majority of the test time, so I'd like to avoid doing that more than is necessary.  I've refactored the code to be a little clearer.

ptnapoleon:
I'm +1, but I'll wait for Jim before merge

mambocab:
Merging, thanks!

mambocab:
New stop-delete-restart logic looks great.

PR #1078, opened 2016-07-05
Only check first two columns of result

ptnapoleon:
Also, use correct number of brackets. This is a fix from a mistake
in the big Assertion Refactoring

Literally anyone to review
@knifewine or @mambocab or @smccarthy788


mambocab:
+1

PR #1081, opened 2016-07-07
Write at CL.ALL for paging tests

stef1927:
Shouldn't we insert at CL.ALL if by default we read at CL.ONE? I managed to get a failure locally by forcing an exclusive connection to one node, which I was using for debugging some paging problems in CASSANDRA-11521.

I also fixed a typo.

### DIFF
--- a/paging_test.py
+++ b/paging_test.py
@@ -183,10 +183,11 @@ class BasePagingTester(Tester):
 
     def prepare(self):
         cluster = self.cluster
-        cluster.populate(3).start()
+        cluster.populate(3).start(wait_for_binary_proto=True)
         node1 = cluster.nodelist()[0]
         session = self.patient_cql_connection(node1)
         session.row_factory = dict_factory
+        session.default_consistency_level = CL.QUORUM
         return session
 
 
@@ -711,7 +712,7 @@ class TestPagingData(BasePagingTester, PageAssertionMixin):
         PAGE_SIZES = (2, 3, 4, 5, 15, 16, 17, 100)
 
         for page_size in PAGE_SIZES:
-            debug("Current page size is %s".format(page_size))
+            debug("Current page size is {}".format(page_size))
             session.default_fetch_size = page_size
             for selector in selectors:
                 results = list(session.execute("SELECT %s FROM test" % selector))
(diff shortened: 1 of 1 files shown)


mambocab:
Thanks for this. I'm reviewing the code change now, and your reasoning is sound, but I'm not an expert on the paging tests.

@blerer Quick question: is Stefania correct that we should change the way we insert and/or read in these tests?

@stef1927 It looks like every session other than [the one created for `test_node_unavailabe_during_paging`](https://github.com/stef1927/cassandra-dtest/blob/f37a1997eec837a165fe4bb896ea2f2f3789c596/paging_test.py#L1768) (sic) is created in [`BasePagingTester.prepare`](https://github.com/stef1927/cassandra-dtest/blob/f37a1997eec837a165fe4bb896ea2f2f3789c596/paging_test.py#L184). I think that means you could replace your CL changes by setting [`default_consistency_level`](https://datastax.github.io/python-driver/api/cassandra/cluster.html#cassandra.cluster.Session.default_consistency_level) to `ALL` or `QUORUM` in `prepare` and `test_node_unavailabe_during_paging`, which I think would be easier to maintain. Do you agree? Do you see any downsides to that implementation?

knifewine:
seems reasonable to me. +1.

stef1927:
+1 for setting default_consistency_level to QUORUM in BasePagingTester.prepare, code updated, thanks.

stef1927:
I've also added `wait_for_binary_proto=True` and fixed a typo that is causing a dtest to fail, see sample failure [here](http://cassci.datastax.com/view/trunk/job/trunk_dtest/lastCompletedBuild/testReport/paging_test/TestPagingData/test_paging_with_filtering_on_counter_columns/).

blerer:
Either we should always insert and read at CL.Quorum or we should insert at CL.ALL. Everything else might cause some unexpected problems.
To be on the safe side I would be in favor of inserting at CL.ALL.

mambocab:
Looks great, thanks to both of you! @stef1927 Feel free to squash and merge, or to use `ALL` if you agree with Benjamin; I don't have strong feelings either way.

knifewine:
I think quorum would probably be more stable, since trying to get 'all' seems to occasionally flake tests.

mambocab:
> quorum would probably be more stable

Ah, yeah, that's right -- so, I vote leaving the default value at `QUORUM`.

stef1927:
I'm also +1 for leaving default_consistency_level at QUORUM. I've rebased and squashed, merging soon.

mambocab:
Beautiful, thanks!

PR #1085, opened 2016-07-11
remove jmxmetrics_test.py

mambocab:
It looks like this test was added, then immediately skipped because it fails. According to Philip, it depends on a number of incorrect assumptions about metrics. We get no value from this, so let's delete it?


ptnapoleon:
+10

mambocab:
Thanks, Philip. I guess we don't need to wait on Travis?

jkni:
Belated +1. This has been around long enough without being fixed that the collective cost of people rediscovering it outweighs the progress it has made toward a comprehensive test for JMX metrics.

ptnapoleon:
It made only backwards progress, except possibly a string based list of all the metrics, which is trivially locatable

PR #1086, opened 2016-07-11
add compaction waits for CASSANDRA-11446

mambocab:
[@ptnapoleon suggests blocking before assertions in this test](https://issues.apache.org/jira/browse/CASSANDRA-11446?focusedCommentId=15296657&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15296657). So, this PR does that.

### DIFF
--- a/scrub_test.py
+++ b/scrub_test.py
@@ -369,6 +369,9 @@ class TestScrub(TestHelper):
         cluster = self.cluster
         cluster.populate(1).start()
         node1 = cluster.nodelist()[0]
+        # we don't want automatic minor compaction because we want to block on
+        # compactions
+        node1.nodetool('disableautocompaction')
 
         session = self.patient_cql_connection(node1)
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I don't think we need compaction waits before every assertion, just the ones related to sstables on disk? We can also turn on minor compaction, depending on the test's assumptions

mambocab:
Makes sense.

mambocab:
After offline discussion, I think Philip means that, instead of waiting for compactions, we could just disable automatic compaction, which I've changed this PR to do.

ptnapoleon:
I'm +1, but multiplex before you close the flaky test failure ticket?

mambocab:
I dig. Running 300 times here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/167/

mambocab:
Lookin' clean. Merging, closing 11446.

PR #1087, opened 2016-07-12
Fix cqlsh login tests to expect error format from CASSANDRA-12076

jkni:
[CASSANDRA-12076](https://issues.apache.org/jira/browse/CASSANDRA-12076) changed the format of error messages on failed login attempts. A PR at the time updated auth_test.py but not the cqlsh tests.

I've verified these pass after the fix on 3.9 and 3.10, so the version gating works as intended.

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -1901,6 +1901,13 @@ class CqlLoginTest(Tester):
         self.node1.watch_log_for('Created default superuser')
         self.session = self.patient_cql_connection(self.node1, user='cassandra', password='cassandra')
 
+    def assert_login_not_allowed(self, user, input):
+        message = ("Provided username {user} and/or password are incorrect".format(user=user)
+            if LooseVersion(self.cluster.version()) >= LooseVersion('3.10')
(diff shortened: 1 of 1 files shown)


cqlsh_tests/cqlsh_tests.py
```diff
@@ -1901,6 +1901,13 @@ def setUp(self):
         self.node1.watch_log_for('Created default superuser')
         self.session = self.patient_cql_connection(self.node1, user='cassandra', password='cassandra')
 
+    def assert_login_not_allowed(self, user, input):
+        message = "Provided username {user} and/or password are incorrect".format(user=user) \
```
mambocab: Style nit: Could you parenthesize this expression rather than using line continuations? e.g.

```python
message = ("Provided username {user} and/or password are incorrect".format(user=user)
           if LooseVersion(self.cluster.version()) >= LooseVersion('3.10')
           else "Username and/or password are incorrect")
```

mambocab:
Other than my style nit, looks great. Passes as expected on 3.9, 3.10, and 3.1. Thanks!

jkni:
Merged with style nit addressed - thanks for the review

PR #1090, opened 2016-07-12
Add test for CASSANDRA-10635

beobal:


ptnapoleon:
Can you please use the unittest assert functions, instead of naked python asserts, as listed in `CONTRIBUTING.md`?

auth_test.py
```diff
@@ -920,6 +921,49 @@ def restart_node_doesnt_lose_auth_data_test(self):
...
+
+            self.get_session(user='cassandra', password='cassandra')
+
+            success = jmx.read_attribute(
+                make_mbean('metrics', type='Client', name='AuthSuccess'), 'Count')
+            failure = jmx.read_attribute(
+                make_mbean('metrics', type='Client', name='AuthFailure'), 'Count')
+
+            self.assertTrue(success > 0)
+            self.assertTrue(failure > 0)
```
ptnapoleon: Could you make these `self.assertGreater(var, 0)`?

ptnapoleon:
+1 after nit

PR #1091, opened 2016-07-13
Fix 12158

ptnapoleon:
@jkni to review. It passes now, alone, and when run as the entire file.

### DIFF
--- a/thrift_tests.py
+++ b/thrift_tests.py
@@ -1497,10 +1497,12 @@ class TestMutations(ThriftTester):
             slice = client.get_slice('key1', ColumnParent('Super1', 'sc2'), p, ConsistencyLevel.ONE)
             assert len(slice) == n, "expected %s results; found %s" % (n, slice)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-12158',
-                   flaky=True)
     def test_describe_keyspace(self):
+        try:
+            client.system_drop_keyspace("ValidKsForUpdate")
+        except InvalidRequestException:
+            pass  # The keyspace doesn't exit, because this test was run in isolation.
+
         kspaces = client.describe_keyspaces()
         if self.cluster.version() >= '3.0':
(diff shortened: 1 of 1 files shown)


thrift_tests.py
```diff
@@ -1497,10 +1497,12 @@ def test_super_reinsert(self):
             slice = client.get_slice('key1', ColumnParent('Super1', 'sc2'), p, ConsistencyLevel.ONE)
             assert len(slice) == n, "expected %s results; found %s" % (n, slice)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-12158',
-                   flaky=True)
     def test_describe_keyspace(self):
+        try:
+            client.system_drop_keyspace("ValidKsForUpdate")
```
mambocab: I don't like that this is hardcoded, but looking at the rest of the tests, I don't see what we could reasonably pull out. If you're cleverer than I am, maybe you can figure out a way to make this use some data shared with other tests.

jkni: Yeah, it seems a little weird for a test to clean up after a different test. I'd prefer to do drop this keyspace in the test that creates it and maybe just drop a comment in test_describe_keyspace that suggests failures may be due to other tests failing to clean up created keyspaces.

ptnapoleon: Tyler has set the precedent, that I agree with, that it is better for a test to ensure it's own starting state, than rely on every single other test to have cleaned up after itself. Seeing as how thrift tests gets virtually zero new development, I think this is fine.

mambocab: As far as cleanup logic goes: leave it. In principle, I lean towards Joel's position -- we don't know from the test's behavior that the `ValidKsForUpdate` keyspace wasn't created, say, on cluster init. However, we do know that it wasn't created on cluster init because C* would never do that.

But still: I'd be happier if there were some bit of data somewhere that said "the tests might leave behind these keyspaces", or "the tests have left behind these keyspaces". I feel like a hardcoded list that could be shared between methods would be appropriate and reasonably self-documenting?

But ultimately, yeah, do what you feel; if you disagree, go ahead and merge.

PR #1094, opened 2016-07-13
CASSANDRA-11698: Flaky clustering_column_test

carlyeks:
Wait for schema agreement after creating table and view


ptnapoleon:
Should we really need to do this?

ptnapoleon:
If yes, +1, but feels like a thing that should work, from a user's perspective.

carlyeks:
We have this in other tests in this suite. Since we aren't making sure that the schema has propagated to all the other nodes before we start writing to them, a write that goes to a node that hasn't yet gotten the schema updates will fail.

PR #1095, opened 2016-07-14
default to env-relevant upgrade tests

knifewine:
This will enable the main dtest CI jobs to also run related upgrade tests.

The basic idea is to figure out what version and version "family" (e.g. 3.x) are being tested, and limit tests to those ending on that version.

In this way we'll be able to divide the upgrade tests among their respective jobs by their end versions.

This would possibly introduce the risk of the upgrade manifest being out of sync with versions being tested, so this also adds logic to make sure the final version upgraded matches the environment (basically overrides that last upgrade's version). This is so we can say "this dtest job ran against SHA xyz, and upgrades were also completed against that same exact SHA".

This also does away with the upgrade variant of 'next', because I think it's obsolete, now that we can run upgrade tests against specific a specific SHA, same as anything else. With that ability, there's not anything really unique about tentative tags. They'll be just another SHA we test upgrading-to.

Finally, I removed 'trunk' as a unique upgrade point in the manifest, and changed the indev 3.x to point to trunk in source. The overlap between the 3.x staging branches and trunk itself seems unlikely to warrant testing both as distinct, but my reasoning about this could be incorrect.


ptnapoleon:
Is this ready for review?

knifewine:
not just yet, but soon, maybe by EOD

knifewine:
@ptnapoleon and @mambocab ready for a proper review now.

upgrade_tests/upgrade_manifest.py
```diff
@@ -1,44 +1,77 @@
...
+    """
+    VersionMeta's are namedtuples that capture data about version family, protocols supported, and current version identifiers
+    they must have a 'variant' value of 'current' or 'indev', where 'current' means most recent released version,
+    'indev' means where changing code is found.
     """
     @property
     def java_version(self):
         return max(self.java_versions)
 
+    def matches_current_env_version_family(self):
```
mambocab: I believe that all the data this depends on is known at import time, so this could just be assigned in `__new__`. Or, since this isn't performance-sensitive, this could be a `@property`.

upgrade_tests/upgrade_manifest.py
```diff
@@ -1,44 +1,77 @@
...
+        """
+        Returns boolean indicating whether this meta matches the current version family of the environment.
+
+        e.g. Returns true if the current env version family is 3.x and the meta's family attribute is a match.
+        """
+        return self.family == get_version_family()
+
+    def clone_with_local_env_version(self):
```
mambocab: If I read this correctly, this logic reimplements `_replace` -- I think this method could just be

```python
return self._replace(version=CASSANDRA_GITREF or CASSANDRA_VERSION_FROM_BUILD)
```

upgrade_tests/upgrade_manifest.py
```diff
@@ -102,13 +128,13 @@ def _is_targeted_variant_combo(origin_meta, destination_meta):
...
       current -> in-dev (aka: released -> branch)
-      current -> next (aka: released -> proposed release point)
-      next -> in-dev (aka: proposed release point -> branch)
     """
-    return (bool(OVERRIDE_MANIFEST) or  # if we're overriding the test manifest, we don't want to filter anything out
-            (origin_meta.variant == 'current' and destination_meta.variant == 'indev') or
-            (origin_meta.variant == 'current' and destination_meta.variant == 'next') or
-            (origin_meta.variant == 'next' and destination_meta.variant == 'indev'))
+    # if we're overriding the test manifest, we don't want to filter anything out
+    if bool(OVERRIDE_MANIFEST):
```
mambocab: Nit: this can be simplified slightly to `if OVERRIDE_MANIFEST:`.

upgrade_tests/upgrade_manifest.py
knifewine: good point. I fixed to keep version family set once as global-ish at the module level and for the meta objects to use a property which then compares to that module level variable. using __new__ seemed like overkill to me.

upgrade_tests/upgrade_manifest.py
knifewine: hey, that's great, updated to use _replace.

upgrade_tests/upgrade_manifest.py
knifewine: updated.

upgrade_tests/repair_test.py
```diff
@@ -1,18 +1,15 @@
...
-from upgrade_base import UPGRADE_TEST_RUN
+from tools import since
 
 LEGACY_SSTABLES_JVM_ARGS = ["-Dcassandra.streamdes.initial_mem_buffer_size=1",
                             "-Dcassandra.streamdes.max_mem_buffer_size=5",
                             "-Dcassandra.streamdes.max_spill_file_size=16"]
 
 
 @since('3.0')
-@skipIf(not UPGRADE_TEST_RUN, 'set UPGRADE_TEST_RUN=true to run upgrade tests')
```
mambocab: It's nice to be able to get rid of these.

dtest.py
```diff
@@ -100,6 +102,14 @@
 # set python-driver log level to INFO by default for dtest
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
+
+def get_sha(repo_dir):
+    try:
+        return "git:{}".format(subprocess.check_output(['git', 'rev-parse', 'HEAD'], cwd=repo_dir).strip())
+    except CalledProcessError:
```
mambocab: I think this should check that the `check_output` call failed for the right reasons, and re-raise if not. Every now and then we run on systems without `git` installed, or on OpenStack, where god only knows what can happen, and it'd be good to explicitly fail in those environments.

mambocab:
Code looks good. Gonna give it another couple reads, but I like what I've seen.

To make sure I understand: the new `RUN_STATIC_UPGRADE_MATRIX` env var unskips all upgrade tests and runs them as specified in the manifest. Otherwise, upgrade tests are skipped unless the class upgrades through (or the upgrade path ends at?) a version in the current version family, where families are `2.1`, `2.2`, `3.0`, `3.X`. In that case, the 'current family' version is replaced by `dtest.CASSANDRA_GITREF`.

upgrade_tests/upgrade_through_versions_test.py
```diff
@@ -822,37 +819,38 @@ def create_upgrade_class(clsname, version_metas, protocol_version,
...
+    # if any version_metas are None, this means they are versions not to be tested currently
     if all(upgrade.version_metas):
-        create_upgrade_class(upgrade.name, [m for m in upgrade.version_metas], protocol_version=upgrade.protocol_version, extra_config=upgrade.extra_config)
+        metas = upgrade.version_metas
+
+        if not RUN_STATIC_UPGRADE_MATRIX:
+            # since we're not running the full test matrix, we're going to test
+            # only upgrades applicable to the current environment, so we need to
+            # change the last version in the upgrade path to match the current env's version exactly
+            metas[-1] = metas[-1].clone_with_local_env_version()
```
mambocab: I think that in this case I'd like a debug statement indicating what's being changed here. Maybe `clone_with_local_env_version` should return a `VersionMeta` with a different name as well, just for clarity?

upgrade_tests/repair_test.py
knifewine: yeah, this makes me happy!

dtest.py
knifewine: ah, good call. this was pretty blind the way I wrote it.

I reworked to catch common failure cases, a) git isn't present, or repo_dir doesn't exist b) git is present, but directory isn't a repo, and c) git fails for some unknown reason

small downside here is if the git output changes this could break in the future, but not a huge risk imo.

let me know what you think here.

knifewine:
yeah, your summary sounds accurate. RUN_STATIC_UPGRADE_MATRIX might be wholly unnecessary, but I think we keep it for now and remove if a few months go by without finding a need for it.

upgrade_tests/upgrade_through_versions_test.py
knifewine: added debug explicity showing version change.

knifewine: this might actually get kinda confusing, silence may be better. since we're generating all test cases we also sub the current env version into each one, so in many cases it will appear like the tests are doing the wrong thing, but those tests aren't actually going to run (because they won't be a match for the env version family).

knifewine: ok, I just need to gate this better and then debug output won't be terribly confusing. will push in a few.

knifewine: pushed, let me know what you think por favor.

dtest.py
```diff
@@ -100,6 +102,22 @@
...
+        return "git:{}".format(output)
+    except CalledProcessError, e:
+        if re.search('Not a git repository', e.message) is not None:
+            # we tried to get a sha, but repo_dir isn't a git repo. No big deal, must just be working from a non-git install.
+            return None
+        else:
+            # git call failed for some unknown reason
+            raise
+    except OSError:
+        raise RuntimeError("Git not installed, or directory {} doesn't exist".format(repo_dir))
```
mambocab: Why not just let the `OSError` get raised here? Whether that's what you want to do or not, it'd be nice for the information from that error propagate to error report.

dtest.py
mambocab: I approve of this for sure: https://github.com/riptano/cassandra-dtest/pull/1095/commits/b301da74cccedb7dd3cb89d8c86551dc6501d2c6#diff-4cbdcd9896ce5009acd37ed830ce8959R114

I'm not sure about catching `OSError`, and I've left a comment to that effect.

> small downside here is if the git output changes this could break in the future, but not a huge risk imo.

Yeah, I agree. Having to deal with different `git` versions should happen pretty infrequently, and it'll fail loudly, so I'm good wiht it.

upgrade_tests/upgrade_through_versions_test.py
mambocab: I dig!

dtest.py
knifewine: Mainly wanted to provide a bit more obvious error message, but OSError alone is probably fine for that. pushed.

upgrade_tests/upgrade_manifest.py
```diff
@@ -1,44 +1,79 @@
...
+    if current_version.vstring.startswith('2.0'):
+        version_family = '2.0.x'
+    elif current_version.vstring.startswith('2.1'):
+        version_family = '2.1.x'
+    elif current_version.vstring.startswith('2.2'):
+        version_family = '2.2.x'
+    elif current_version.vstring.startswith('3.0'):
+        version_family = '3.0.x'
+    elif current_version > '3.0':
+        version_family = '3.x'
```
mambocab: I think it might be worth blowing up if the version >= 4, just so there's something loud to remind us to handle it correctly.

knifewine: good idea, pushed.

mambocab:
Ok, reread, and I'm down with the code. Lemme know when you have CI links you want reviewed as well.

knifewine:
Each branch on cassci has a job that's been running on this dtest branch, so it would be good to review those recent runs and see that things look ok to you (keeping in mind the code has likely changed a bit depending on when the run took place).

once we merge, I'll update the templates to use dtest master

PR #1096, opened 2016-07-14
Move 5 node tests to large tests

ptnapoleon:
@mambocab to review


mambocab:
So, this could be why we saw the failure in [CASSANDRA-12176](https://issues.apache.org/jira/browse/CASSANDRA-12176)?

ptnapoleon:
Yes, those should be in large tests

conf/trunk_test-select.cfg
```diff
@@ -7,6 +7,9 @@ consistency_test.py:TestAccuracy.test_network_topology_strategy_each_quorum_user
   tools.py:TestAccuracy.test_network_topology_strategy_each_quorum_users
 consistency_test.py:TestAccuracy.test_network_topology_strategy_each_quorum_counters
   tools.py:TestAccuracy.test_network_topology_strategy_each_quorum_counters
+consistency_test.py:TestAccuracy.test_simple_strategy_each_quorum_users
+  tools.py:TestAccuracy.test_simple_strategy_each_quorum_users
+consistency_test.py:TestAccuracy.test_simple_strategy_users
```
mambocab: What about `tools.py:...`?

ptnapoleon: Don't need it here, since this test isn't annotated with `since`

mambocab:
+1.

PR #1097, opened 2016-07-14
error message showing actual value found if assertEqual fails

knifewine:
@mambocab to review. related to CASSANDRA-11687


mambocab:
+1, good idea.

PR #1102, opened 2016-07-15
12097: failure in view_tombstone_test

carlyeks:
Make sure to drain the batchlog after each commit

### DIFF
--- a/materialized_views_test.py
+++ b/materialized_views_test.py
@@ -755,7 +755,8 @@ class TestMaterializedViews(Tester):
         # Set initial values TS=0, verify
         session.execute(SimpleStatement("INSERT INTO t (id, v, v2, v3) VALUES (1, 1, 'a', 3.0) USING TIMESTAMP 0",
                                         consistency_level=ConsistencyLevel.ALL))
-
+        self._replay_batchlogs()
+        
         assert_one(
             session,
             "SELECT * FROM t_by_v WHERE v = 1",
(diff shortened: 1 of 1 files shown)


materialized_views_test.py
```diff
@@ -1435,7 +1439,7 @@ def handle_errors(row, exc):
                 name = type(exc).__name__
                 self.exception_type[name] += 1
             except Exception as e:
-                print traceback.format_exception_only(type(e), e)
+                print(traceback.format_exception_only(type(e), e))
```
ptnapoleon: Why are you making these changes to print? We are using python 2

mambocab: My guess is that Carl is using a `pyflakes` or `flake8` installed in a different environment than he's running dtest. +1, we should be using `print` as a statement here.

ptnapoleon:
No more complaints from me, I'll let you wait for Joel's review

materialized_views_test.py
```diff
@@ -784,6 +787,7 @@ def view_tombstone_test(self):
         )
 
         node2.start(wait_other_notice=True, wait_for_binary_proto=True)
+        self._replay_batchlogs()
```
jkni: I'm not sure why we need to replay batchlogs here - batchlog replay is already up-to-date on all nodes I think.

jkni:
Looks good to me - I left one question for @carlyeks inline. I'm multiplexing this on CassCI and will update here once that's done.

jkni:
200 run multiplex passed [here](https://cassci.datastax.com/view/Dev/view/jkni/job/jkni_parameterized_dtest_multiplexer/1/testReport/).

materialized_views_test.py
carlyeks: Yeah, this should be waiting for hints to replay before continuing since the up nodes will have already converted the batchlogs into hints for node2 with the last `_replay_batchlogs()`.

The MBean which would should provide the ability to playback a single node's hints isn't implemented after 6230. I'll get rid of this line, since we're already allowing read repair to fix things by using CL.ALL.

PR #1105, opened 2016-07-18
Up amount of data so relevel pushes levels up

carlyeks:


ptnapoleon:
Could you make sure you remove the known failure annotation for this ticket, either in this PR, or after commit? Otherwise, I'm +1 on bumping up data to get this to pass (I see Tyler is reviewer, so you can wait for him to +1 before merge if you like).

I should also note, and I can do this separately, we're not specifying thread count in our stress calls, causing them to be repeated, which is part of the test's slowness problem. Do you want me to handle that in a separate PR, or would you like to do it here?

PR #1106, opened 2016-07-18
make `nodetool rebuild` calls use same logic

mambocab:
Noticed this while looking into

https://issues.apache.org/jira/browse/CASSANDRA-11687

This code wasn't using the same logic between `rebuild` and the code at the top level in the test -- the to-level code would increment `self.rebuild_errors` on any `NodetoolError`, rather than just the one we expect.

I've also re-raised the error when we find an unexpected `NodetoolErrors` -- these should fail the test, yes?


ptnapoleon:
+1

mambocab:
Thanks, merging.

PR #1107, opened 2016-07-18
tag flaky test

mambocab:
@ptnapoleon @thobbs This is the same failure as https://issues.apache.org/jira/browse/CASSANDRA-12213 right?


ptnapoleon:
Yes, and can you remove it from `test_mutation_any`? It was placed on the wrong test, and I am 100% sure I caught that on review. I have no idea how the unfixed version was merged, probably my fault. +1

mambocab:
Done, thanks for the review. Merging.

PR #1113, opened 2016-07-20
Update consistent bootstrap tests to support non-vnodes case

pauloricardomg:
Basically rf=1 tests were flaky for non-vnodes because depending where the new node was bootstrapped, the consistent bootstrap could succeed or not (depending if the node would take or not the range of the down node). The fix is to use deterministic tokens for the non-vnode case.

Also took the chance to improve comments, test naming and simplify test to use 2 nodes instead of 3.

Non-vnode dtest run submitted here: http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunk-novnode-dtest/

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -12,7 +12,7 @@ from cassandra.concurrent import execute_concurrent_with_args
 from ccmlib.node import NodeError
 
 from assertions import assert_almost_equal, assert_one, assert_not_running
-from dtest import Tester, debug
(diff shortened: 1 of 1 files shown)


ptnapoleon:
If the test passes on CI, +1!

pauloricardomg:
Test results look good, merging.

pauloricardomg:
thanks!

PR #1114, opened 2016-07-20
correctly fail on errors from rebuild thread

mambocab:
This test didn't fail when the `rebuild` call in the thread raised an exception; this PR fixes that.

Clarity question: I've written a function that generically takes a function here. Since it's only used the once, would it be clearer to write:

```python
def rebuild_and_register_exc_info_on_exception():
    """
    Closes over self to catch any exceptions raised by func and
    register them at self.thread_exc_info
    Based on http://stackoverflow.com/a/1854263
    """
    try:
        rebuild()  # closed over
    except Exception:
        import sys
        self.thread_exc_info = sys.exc_info()

cmd1 = Thread(target=rebuild_and_register_exc_info_on_exception)
```

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -76,6 +76,8 @@ class TestRebuild(Tester):
         session.execute('USE ks')
 
         self.rebuild_errors = 0
+        # make it possible for thread to raise errors
+        self.thread_exc_info = None
 
         # rebuild dc2 from dc1
         def rebuild():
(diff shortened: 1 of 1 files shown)


knifewine:
+1 if it works.

ptnapoleon:
Ditto what Russ said

mambocab:
Cool, thanks, folks. Merging.

PR #1116, opened 2016-07-21
add known failure annotations

smccarthy788:
Waiting on Travis to merge.


user_functions_test.py
```diff
@@ -32,6 +32,10 @@ def prepare(self, create_keyspace=True, nodes=1, rf=1):
             self.create_ks(session, 'ks', rf)
         return session
 
+
+    @known_failure(failure_source='test',
```
ptnapoleon: Don't even both adding this, it's already fixed. I don't want to race and merge the fix, then the annotation, and then it sits there forever

jkni:
@smccarthy788 can you reduce this PR to just the mv known failure annotation? the repair_tests one was added in another one and the one Philip noted as unnecessary has been fixed.

PR #1120, opened 2016-07-22
Fixes for CASSANDRA-11850

stef1927:
These are the companion changes for [CASSANDRA-11850](https://issues.apache.org/jira/browse/CASSANDRA-11850), which upgrades the python driver version embedded with cqlsh. 

This PR must be merge at the same time as the c* patch.


ptnapoleon:
+1, iff you are sure the change to json_test.py is true across all versions.

stef1927:
Thanks for the review. The json tests execute only for 2.2+ so the change is relevant for all versions since they all share the same driver version.

Ticket has been committed, merging.

PR #1121, opened 2016-07-22
Add test for nodetool garbagecollect (CASSANDRA-7019).

blambov:


compaction_test.py
```diff
@@ -480,6 +480,58 @@ def user_defined_compaction_test(self):
         self.assertEquals(len(node1.data_directories()), len(sstable_files),
                           'Expected one sstable data file per node directory but got {}'.format(sstable_files))
 
+    @since('3.9')
+    def nodetool_garbagecollect_test(self):
+        """
+        Test that nodetool garbagecollect decreases size and has no effect on the contents of the table.
```
ptnapoleon: Can you add a line with
`@jira_ticket CASSANDRA-7019` to this docstring?

ptnapoleon:
Could you use prepared statements to speed up this test? http://datastax.github.io/python-driver/api/cassandra/query.html?highlight=prepared%20statements

Also, instead of doing string concatenation in the form `"insert " + str(x) + ", " + str(y)`, could you use interpolation? It makes it more readable:
`"INSERT INTO TABLE (key, column) VALUES ({}, {})".format(x, y)`

compaction_test.py
```diff
@@ -480,6 +480,58 @@ def user_defined_compaction_test(self):
...
+                session.execute('delete from cf where key = ' + str(x) + ' and subkey = ' + str(y))
+        # row delete
+        for x in range(10, 15):
+            for y in range(0, 5):
+                session.execute('delete from cf where key = ' + str(x) + ' and subkey = ' + str(y) + ' and column = ' + str(y))
+        # cell delete
+        for x in range(15, 20):
+            for y in range(0, 5):
+                session.execute('delete val from cf where key = ' + str(x) + ' and subkey = ' + str(y) + ' and column = ' + str(y))
+        rowcount = 21 * 6 - 5 * 6 - 5 * 5 - 5 * 5
```
mambocab: This might be easier to follow if you increment it in the loops above -- it's hard to see what's supposed to be happening here.

ptnapoleon:
@blambov, would you like one of us to take over the review changes on this PR?

PR #1122, opened 2016-07-22
Expect extra notifications during restart on 2.1

beobal:


beobal:
See CASSANDRA-11360

pushed_notifications_test.py
```diff
@@ -164,22 +164,25 @@ def restart_node_test(self):
         waiter = NotificationWaiter(self, node1, ["STATUS_CHANGE", "TOPOLOGY_CHANGE"])
 
         # need to block for up to 2 notifications (NEW_NODE and UP) so that these notifications
-        # don't confuse the state below
+        # don't confuse the state below.
         debug("Waiting for unwanted notifications...")
         waiter.wait_for_notifications(timeout=30, num_notifications=2)
         waiter.clear_notifications()
 
+        expected_notifications = 2 if self.cluster.cassandra_version() >= '2.2' else 3
```
mambocab: It'd be good to note in a comment that this is a known bug in 2.1 with a link to CASSANDRA-11360.

Also, I believe you'll want to change the condition here to

```python
LooseVersion(self.cluster.cassandra_version()) >= '2.2'
```

I'm not positive it matters for cassandra versions less than `3.10`, but it'd be cleaner to do this rather than a lexical comparison.

mambocab:
Other than my nit, I'm +1; feel free to merge after that's addressed.

beobal:
Thanks @mambocab

PR #1124, opened 2016-07-24
Use schema_wait_session to create UDFs

ptnapoleon:
@mambocab or @knifewine or @smccarthy788. This WILL fix the test, but I'm only 80% sure doesn't change the test in a bad way. Someone sanity check this is reasonable?

### DIFF
--- a/user_functions_test.py
+++ b/user_functions_test.py
@@ -65,17 +65,17 @@ class TestUserFunctions(Tester):
         node1_session.execute("INSERT INTO udf_kv (key, value) VALUES ({}, {})".format(2, 2))
         node1_session.execute("INSERT INTO udf_kv (key, value) VALUES ({}, {})".format(3, 3))
 
-        node1_session.execute("""
+        schema_wait_session.execute("""
             create or replace function x_sin ( input double ) called on null input
             returns double language java as 'if (input==null) return null;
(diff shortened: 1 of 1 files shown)


mambocab:
I don't have enough context for this. What's the error?

The implication I'm getting from this, though, is that you have to use a particular session for the `wait_for_schema_agreement` call below this code to successfully block on schema agreement. That sounds like a bug.

ptnapoleon:
I don't think its a bug, so much as a race condition/necessary evil

mambocab:
I still don't follow. The restrictions on blocking on schema agreement are on the session you use to block, not on the session you execute the schema changes from.

ptnapoleon:
Yep, we need someone to confirm if this is a bug or not. Would it be in the driver maybe? @jkni ?

ptnapoleon:
Oh god I've forgotten all the context on this, Jim, let's get back to it on Monday?

ptnapoleon:
From discussion with Joel, this just seems like a python driver limitation. Whether that gets improved or not, I'd like to merge this, please

mambocab:
Fair enough. File a PYTHON ticket and merge IMO.

ptnapoleon:
From Joel's comments, sounds like it is fixed in 3.6.0, so no need.

PR #1125, opened 2016-07-25
CASSANDRA-11465: increase timeout for tracing synchronization

stef1927:
These are the changes that should be committed after CASSANDRA-11465 is also committed, to improve reliability with tracing events.


ptnapoleon:
+1

stef1927:
C* ticket committed, merging.

PR #1130, opened 2016-07-26
CASSANDRA-12287: Fix sstablelevelreset_test

nastra:
This should fix the test by removing a duplicate **-rate ...** parameter, which was passed to **stress**.


carlyeks:
Thanks for catching that!

ptnapoleon:
+1, Sorry I missed this!

PR #1133, opened 2016-07-26
add ALTER KEYSPACE to make rebuild commands work

mambocab:
This is a fix for the immediate problem in [CASSANDRA-11687](https://issues.apache.org/jira/browse/CASSANDRA-11687). This will not close the ticket, as the tests fails.

I'm not sure this is a long-term solution, so I've filed [CASSANDRA-12296](https://issues.apache.org/jira/browse/CASSANDRA-12296).

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -73,6 +73,9 @@ class TestRebuild(Tester):
         # alter keyspace to replicate to dc2
         session = self.patient_exclusive_cql_connection(node2)
         session.execute("ALTER KEYSPACE ks WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
+        # alter system_auth -- rebuilding it no longer possible after
+        # CASSANDRA-11848 prevented local node from being considered a source
+        session.execute("ALTER KEYSPACE system_auth WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
         session.execute('USE ks')
 
         self.rebuild_errors = 0


rebuild_test.py
```diff
@@ -73,6 +73,9 @@ def simple_rebuild_test(self):
         # alter keyspace to replicate to dc2
         session = self.patient_exclusive_cql_connection(node2)
         session.execute("ALTER KEYSPACE ks WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
+        # alter system_auth -- rebuilding it no longer possible after
+        # CASSANDRA-11848 prevented local node from being considered a source
+        session.execute("ALTER KEYSPACE system_auth WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
```
ptnapoleon: I thought making it RF=1 just in dc1 would pass the test?

mambocab: What makes you think that? Here's what Paulo said:

> Fix is to either use `-Dcassandra.consistent.rangemovement=false` or update `system_auth` to use `NetworkTopologyStrategy` with 2 dcs.

Also, it's already `SimpleStrategy` with `RF=1` by default.

ptnapoleon: I just totally brain farted on how NTS works. Both settings should work, so leave it as is. Also I don't get your comment on SS. SS with RF=1 != NTS with RF dc1=1

+1

mambocab: > Also, it's already SimpleStrategy with RF=1 by default

Sorry, didn't make my referent clear. I was referring to the default replication for `system_auth`.

mambocab:

PR #1134, opened 2016-07-26
Clean up token_generator.py

ptnapoleon:
Rename to token_generator_test.py so nose picks it up
Do some better string formatting
Add jira ticket and since annotaitons
Use os.path.join

@mambocab and @c-kodman to review

### DIFF
--- a/token_generator_test.py
+++ b/token_generator_test.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 
+import os
 import re
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


token_generator_test.py
```diff
@@ -8,16 +9,20 @@
...
     """
     Basic tools/bin/token-generator test.
+    Token-generator was removed in CASSANDRA-5261
+    @jira_ticket CASSANDRA-5261
+    @jira_ticket CASSANDRA-9300
     """
 
     def call_token_generator(self, install_dir, randomPart, nodes):
-        executable = install_dir + "/tools/bin/token-generator"
+        executable = os.path.join(install_dir, 'tools', 'bin', 'token-generator')
```
mambocab: This is fine, but should probably be in CCM somewhere, right?

mambocab:
+1, looks good

PR #1135, opened 2016-07-26
Block until index is finished before inserting data

ptnapoleon:
@mambocab, I know you won't like this, so I'm asking you to review

### DIFF
--- a/secondary_indexes_test.py
+++ b/secondary_indexes_test.py
@@ -440,9 +440,6 @@ class TestSecondaryIndexes(Tester):
                            [("127.0.0.1", 1, 200), ("127.0.0.2", 1, 200), ("127.0.0.3", 1, 200)],
                            retry_on_failure)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-12288',
-                   flaky=True)
     @skipIf(DISABLE_VNODES, "Test should only run with vnodes")
     def test_query_indexes_with_vnodes(self):
         """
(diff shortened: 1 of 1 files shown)


secondary_indexes_test.py
```diff
@@ -459,6 +456,19 @@ def test_query_indexes_with_vnodes(self):
...
+        def index_is_built(table_name, idx_name):
+            index_query = (
+                """SELECT * FROM system_schema.indexes WHERE keyspace_name = 'ks' AND table_name = '{}' AND index_name = '{}'""".format(table_name, idx_name)
+                if self.cluster.version() > '3.0' else
+                """SELECT * FROM system."IndexInfo" WHERE table_name = 'ks' AND index_name = '{}.{}'""".format(table_name, idx_name)
+            )
+            return len(list(session.execute(index_query))) == 1
+
+        start = time.time()
+        while not index_is_built('regular_table', 'composites_index') and time.time() + 10 < start:
```
mambocab: this `time.time() + 10 < start` should be checked as an assertion in the loop, so we can't continue the test if the index doesn't build in time.

mambocab:
Do we really not have this code in CCM somewhere? Sigh. It'd be nice if you made this more generic and moved it up there, if you have the time. Code looks fine though, go ahead and get the test fixed if you don't have time to push this down to CCM.

secondary_indexes_test.py
ptnapoleon: The test will fail if the index doesn't build in time, with an error explaining the index didn't build in time, so feels redundant to me?

ptnapoleon:
I think if it's getting genericized (Which I swear I did and can't find), it belongs in a dtest util location somewhere, not in CCM. I'll merge this (with your +1), and open a ticket to genericize this? It is used in soooo many places across all of dtest

mambocab:
> it belongs in a dtest util location somewhere, not in CCM

Huh, why? Seems like something that uses a bunch of data from a `Node` and could easily be a method on one.

But, yeah +1 to the "merge and file a ticket" plan.

ptnapoleon:
For future readers, this function doesn't work on 3.0+ clusters

PR #1138, opened 2016-07-26
change assert to self.assertIn

c-kodman:


mambocab:
`in` and `re.search` aren't the same. Have a look in the `unittest.TestCase` documentation; there should be a useful regex assertion in there somewhere.

ptnapoleon:
They are if you use re.search to find a string literal

ptnapoleon:
We might need to change the escape chars, but there's no reason to use re.search if we aren't using a regex

mambocab:
Oh, yup. My mistake. `assertIn` is fine then, but some escape characters will have to be removed I believe.

ptnapoleon:
+1

PR #1139, opened 2016-07-27
CASSANDRA-11235: Ignore JNA/JavaLaunchHelper warning

nastra:
Couldn't really try it out because I don't have **windows**. @ptnapoleon can you take a look?

### DIFF
--- a/offline_tools_test.py
+++ b/offline_tools_test.py
@@ -16,10 +16,6 @@ class TestOfflineTools(Tester):
     # in the classpath
     ignore_log_patterns = ["Unable to initialize MemoryMeter"]
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11235',
-                   flaky=False,
-                   notes='windows')
     def sstablelevelreset_test(self):
         """
         Insert data and call sstablelevelreset on a series of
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Could you restrict the additional check to only happen `if common.is_win()`? Where `from ccmlib import common` is our source of common, i.e., only ignore this error on windows?

nastra:
ok will do

nastra:
@ptnapoleon updated the PR

offline_tools_test.py
```diff
@@ -405,9 +401,10 @@ def sstabledump_test(self):
         self.assertEqual(dumped_row, '1')
 
     def _check_stderr_error(self, error):
+        jna_link_failure = "JNA link failure" if common.is_win() else ''
         if len(error) > 0:
             for line in error.splitlines():
-                self.assertTrue("Max sstable size of" in line or "Consider adding more capacity" in line,
+                self.assertTrue("Max sstable size of" in line or "Consider adding more capacity" in line or jna_link_failure in line,
```
ptnapoleon: Won't this always evaluate to true now when not on linux?

```
'' in 'longer_string'
True
```

offline_tools_test.py
```diff
@@ -405,9 +401,12 @@ def sstabledump_test(self):
         self.assertEqual(dumped_row, '1')
 
     def _check_stderr_error(self, error):
+        acceptable = ["Max sstable size of", "Consider adding more capacity", "JNA link failure", "Class JavaLaunchHelper is implemented in both"]
```
nastra: @ptnapoleon I was actually able to hit this same **JNA link failure** warning message on OS X. So maybe we should ignore the warning not only for win?

I also added the **JavaLaunchHelper** warning, since it's constantly happening on OS X due to a bug in the JDK.

Let me know what you think.

offline_tools_test.py
```diff
@@ -405,9 +401,12 @@ def sstabledump_test(self):
...
+
         if len(error) > 0:
+            debug(error)
```
nastra: oh this should be removed

mambocab:
:/ https://github.com/riptano/cassandra-dtest/issues/1152 sorry @nastra!

mambocab:
Ah, good point about the `known_failure`. Looks good, sorry for missing all this earlier. +1, feel free to merge.

nastra:
@ptnapoleon are you fine with merging?

PR #1144, opened 2016-07-27
add known failure annotation

c-kodman:


compaction_test.py
```diff
@@ -250,6 +253,10 @@ def compaction_throughput_test(self):
                                         '''.format(units), re.X)
 
         avgthroughput = re.match(throughput_pattern, stringline).group(1).strip()
+        print "throughput_patter is"
```
ptnapoleon: These snuck in, they shouldn't be here

ptnapoleon:
+1

PR #1145, opened 2016-07-27
jira-10810 resumable_rebuild_test

kdmu:


knifewine:
Thanks for the test, @kdmu ! I'll be testing and reviewing your pr shortly.

rebuild_test.py
```diff
@@ -128,6 +132,105 @@ def call_and_register_exc_info_on_exception(func):
...
+        session.execute("ALTER KEYSPACE ks WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
+        session.execute("ALTER KEYSPACE system_auth WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
+
+        self.rebuild_errors = 0
+
+        # Path to byteman script which makes node2 throw an exception making rebuild fail
+        script = ['./rebuild_failure_inject.btm']
+
+        try:
+            print('Executing first rebuild -> '),
```
knifewine: please update print statements to use dtest.py:debug function.

rebuild_test.py
```diff
@@ -128,6 +132,105 @@ def call_and_register_exc_info_on_exception(func):
...
+
+        cluster = self.cluster
+        cluster.set_configuration_options(values={'endpoint_snitch': 'org.apache.cassandra.locator.PropertyFileSnitch'})
+
+        # Create 2 nodes on dc1
+        node1 = cluster.create_node('node1', False,
+                                   ('127.0.0.1', 9160),
+                                   ('127.0.0.1', 7000),
+                                   '7100', '2000', None,
+                                   binary_interface=('127.0.0.1', 9042))
```
knifewine: please indent this line one more space, and the same for the 3 lines above. This might be what caused the travis linter to complain.

rebuild_test.py
```diff
@@ -128,6 +132,105 @@ def call_and_register_exc_info_on_exception(func):
...
+
+        # Path to byteman script which makes node2 throw an exception making rebuild fail
+        script = ['./rebuild_failure_inject.btm']
+
+        try:
+            print('Executing first rebuild -> '),
+            node3.byteman_submit(script)
+            node3.nodetool('rebuild dc1')
+            print('Unexpected: OK')
+        except NodetoolError as e:
```
knifewine: can remove the 'as e' clause here, since e is not used.

rebuild_test.py
```diff
@@ -128,6 +132,105 @@ def call_and_register_exc_info_on_exception(func):
...
             query_c1c2(session, i, ConsistencyLevel.ALL)
 
+    @since('2.2')
+    def resumable_rebuild_test(self):
+        """
+        @jira_ticket CASSANDRA-10810
+
+        Test rebuild operation is resumable
+        """
+
```
knifewine: can remove this blank line

rebuild_test.py
```diff
@@ -128,6 +132,105 @@ def call_and_register_exc_info_on_exception(func):
...
+        # First rebuild must fail and data must be incomplete
+        self.assertEqual(self.rebuild_errors, 1)
+        session.execute('USE ks')
+        try:
+            print('Checking data is complete -> '),
+            for i in xrange(0, 20000):
+                query_c1c2(session, i, ConsistencyLevel.LOCAL_ONE)
+            print('Unexpected: OK')
+        except AssertionError:
+            print('Expected: INCOMPLETE')
```
knifewine: if I'm understanding this correctly, the except clause should probably re-raise to allow the test to fail immediately

knifewine:
I'm finding the print/debug statements like 'UNEXPECTED: OK' and 'EXPECTED: OK' to be a bit confusing. Could these be made a bit more descriptive?

knifewine:
byteman fault injection seems reasonable to me if it's the most directly way to test this. if there's a way that doesn't require byteman, that would be most ideal.

it looks like byteman isn't pulled in by 2.2's build.xml which means this test won't be able to run against 2.2. byteman looks to be available on 3.0+.

knifewine:
Tested on 3.0 and trunk and appears to work fine via ccm, looks like test should function as long as executing environment has JAVA_HOME set and is against a cassandra version which pulls in byteman as a dependency (2.2 does not appear to bring in byteman, unfortunately).

kdmu:
In any case the first data check-up may not be necessary because we are only testing if rebuild operation is resumable and if it does properly its work.

In a hypothetical scenario that stream session failed when all data is already streamed, those streamed data will be skipped by our implementation, so I guess we only need to check if we can retrieve from a failed rebuild and if after all we have all data.

yukim:
It is eaisier to add byteman support for 2.2 (I think this is done just by adding dependencies in 2.2 build.xml) than modify test only for 2.2.
If that sounds reasonable, I will create C* JIRA ticket.

knifewine:
Sounds reasonable to me.

rebuild_test.py
```diff
@@ -128,6 +132,104 @@ def call_and_register_exc_info_on_exception(func):
...
+            self.rebuild_errors += 1
+            debug('Expected: FAILED')
+
+        # First rebuild must fail and data must be incomplete
+        self.assertEqual(self.rebuild_errors, 1)
+        session.execute('USE ks')
+        try:
+            debug('Checking data is complete -> '),
+            for i in xrange(0, 20000):
+                query_c1c2(session, i, ConsistencyLevel.LOCAL_ONE)
```
knifewine: If I'm understanding correctly, we want to confirm that an AssertionError is always raised here?

If so we should use the assertRaises assertion (probaby as a context manager).

rebuild_test.py
```diff
@@ -128,6 +132,104 @@ def call_and_register_exc_info_on_exception(func):
...
+
+        # Path to byteman script which makes node2 throw an exception making rebuild fail
+        script = ['./rebuild_failure_inject.btm']
+
+        try:
+            debug('Executing first rebuild -> '),
+            node3.byteman_submit(script)
+            node3.nodetool('rebuild dc1')
+            debug('Unexpected: SUCCEED')
+        except NodetoolError:
```
knifewine: I think this might be another good place to use assertRaises, assuming that the test would be tainted if the exception doesn't occur where it's expected (see note below this one).

rebuild_test.py
kdmu: You're right. I'll update them to use assertRaises instead.

kdmu:
A part from what's reflected in the commit message, self.rebuild_errors have been removed since it's no longer used.

rebuild_test.py
knifewine: Thanks! Just want to be sure the test is as deterministic as possible.

rebuild_test.py
```diff
@@ -20,7 +20,11 @@ def __init__(self, *args, **kwargs):
             r'Can\'t send migration request: node.*is down',
             # ignore streaming error during bootstrap
             r'Exception encountered during startup',
-            r'Streaming error occurred'
+            r'Streaming error occurred',
+            # Following errors are related to resumable_rebuild_test
+            r'Error while rebuilding node',
+            r'Streaming error occurred on session with peer 127.0.0.3',
+            r'Remote peer 127.0.0.3 failed stream session'
```
knifewine: Since log patterns are ignored at the class level, is it possible that some of these ignore patterns may silence legitimate problems in unrelated tests in the same class?

If that's the case, it might make sense to move some of those new ignore patterns into the test method in such a way that it won't mutate the class-shared ignore list. Something like:

```
def resumable_rebuild_test(self):
  self.ignore_log_patterns = self.ignore_log_patterns[:] + [patterns_to_ignore_on_this_test]
```

knifewine:
just labeled as pending C* commit, so we remember to hold off merge until 2.2 can get byteman added.

rebuild_test.py
kdmu: Right now 127.0.0.3, i.e. node3 is only used by resumable_rebuild_test, so I don't think there would be any conflicts with related errors, but instead "Error while rebuilding node" might cause some problem to "rebuild_ranges_test", for safety and to avoid possible problem in the future I agree it's better to add them at method level. I'll update it in the next patch. Thanks!

yukim:
Cassandra JIRA here: [CASSANDRA-12377](https://issues.apache.org/jira/browse/CASSANDRA-12377)

knifewine:
Looks like a rebase is probably needed here.

yukim:
FYI CASSANDRA-12377 was committed and closed.

knifewine:
Great, thanks for the update @yukim . I think we're just about ready to merge here.

@kdmu Looks like the travis linter is probably unhappy about some trailing spaces in the code. Once those are fixed, can you rebase against master?

kdmu:
@knifewine Fixed those nits and rebased my branch, I guess it's ready to be merged. Thanks!

rebuild_test.py
```diff
@@ -132,6 +132,96 @@ def run(self):
...
+                                    ('127.0.0.3', 7000),
+                                    '7300', '2002', None,
+                                    binary_interface=('127.0.0.3', 9042),
+                                    byteman_port='8300')
+
+        cluster.add(node3, False, data_center='dc2')
+
+        node3.start(wait_other_notice=False, wait_for_binary_proto=True)
+
+        time.sleep(60)
```
knifewine: Is this sleep needed still, or is there possibly an alternative to using this?

kdmu: It's used to ensure snitch is refreshed before rebuild. I'm not sure if it's convenient remove it, we could reduce it to 10 which is the value I've been using while testing in my local machine.

yukim: Looks like after [CASSANDRA-10243](https://issues.apache.org/jira/browse/CASSANDRA-10243), PropertyFileSnitch's refresh rate is reduced from 60 sec to 5 sec, we can change this to 5.

pauloricardomg:
@kdmu it seems there's an import error after the rebase: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/243/testReport/junit/node_0_iter_002.nose.failure/Failure/runTest/

I think it is due to the fact that the exception was renamed from `NodetoolError` to `ToolError`. Can you rebase again and fix this? I will submit another multiplexer run after this is fixed. Thanks!

knifewine:
I'm running this test locally, and seeing failures on 3.0 and trunk: 

> ERROR [STREAM-OUT-/127.0.0.3] 2016-08-12 16:24:15,427 Ref.java:212 - BAD RELEASE: attempted to release a reference (org.apache.cassandra.utils.concurrent.Ref$State@29efe800) that has already been released

This should be easy to repro locally, but I can re-run attach a full stacktrace if that's helpful.

kdmu:
Tested again locally with C* 3.0 and trunk by tailing system and debug logs, but I couldn’t find any Ref error in rebuild section, in fact STREAM-OUT logs are only related with ConnectionHandler. Could you attach the stacktrace?

And for some reason which I don't why, this test does not write its logs to logs/ directory (at least in my case)

pauloricardomg:
I was able to reproduce this on some [multiplexer run|https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/244/]. It seems there is a race causing a reference to be release twice. I will investigate this later.

pauloricardomg:
This was a regression caused by [CASSANDRA-11345](https://issues.apache.org/jira/browse/CASSANDRA-11345), already submitted a fix and after it's committed we can re-run multiplexer results for this and merge after it looks good. Thanks!

pauloricardomg:
Merging this after regression for [CASSANDRA-11345](https://issues.apache.org/jira/browse/CASSANDRA-11345) was fixed, [multiplexer](https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/250/) and [dtest](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-2.2-10810-dtest/3/) results look good. Thank you all for the review.

PR #1146, opened 2016-07-27
Fix simple_rebuild_test

yukim:
Current test implementation didn't trigger rebuild concurrently even though we are using thread in my local laptop. I think the same thing is happening in CI since it is hitting assertion error that indicates it.

So I changed the code to run nodetool in background first, and then immediately trigger second one in foreground, and the latter should fail at that point.

Rebuild completion from the former is guarded by wathing log.

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -78,55 +78,21 @@ class TestRebuild(Tester):
         session.execute("ALTER KEYSPACE system_auth WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
         session.execute('USE ks')
 
-        self.rebuild_errors = 0
-        # make it possible for thread to raise errors
-        self.thread_exc_info = None
-
-        # rebuild dc2 from dc1
-        def rebuild():
-            try:
-                node2.nodetool('rebuild dc1')
-            except NodetoolError as e:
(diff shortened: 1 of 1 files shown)


rebuild_test.py
```diff
@@ -78,55 +77,21 @@ def simple_rebuild_test(self):
...
         # exactly 1 of the two nodetool calls should fail
-        # usually it will be the one in the main thread,
-        # but occasionally it wins the race with the one in the secondary thread,
-        # so we check that one succeeded and the other failed
-        self.assertEqual(self.rebuild_errors, 1,
-                         msg='rebuild errors should be 1, but found {}. Concurrent rebuild should not be allowed, but one rebuild command should have succeeded.'.format(self.rebuild_errors))
+        try:
+            node2.nodetool('rebuild dc1')
+        except NodetoolError as e:
+            self.assertTrue('Node is still rebuilding' in e.message)
```
mambocab: This should fail if the `nodetool` call doesn't raise an error.

yukim: You are right, pushed fix.

pauloricardomg:
+1

mambocab:
This is great, sorry to miss the final +1. We in TE should have seen this. Thank you both!

PR #1147, opened 2016-07-27
Cstar 97

smccarthy788:
@mambocab, for your review.

In conjunction with https://github.com/pcmanus/ccm/pull/510

CI run here: https://cassci.datastax.com/job/ptnapoleon-trunk-dtest/8/testReport/

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -2,7 +2,6 @@ import os
 import random
 import re
 import shutil
-import subprocess
 import tempfile
 import threading
 import time
@@ -283,9 +282,8 @@ class TestBootstrap(Tester):
 
         debug("Check data is present")
         # Let's check stream bootstrap completely transferred data
-        stdout, stderr = node3.stress(['read', 'n=100k', 'no-warmup', '-schema',
-                                       'replication(factor=2)', '-rate', 'threads=8'],
-                                      capture_output=True)
+
+        stdout, stderr, _ = node3.stress(['read', 'n=100k', 'no-warmup', '-schema', 'replication(factor=2)', '-rate', 'threads=8'])
 
         if stdout is not None:
             self.assertNotIn("FAILURE", stdout)
@@ -397,18 +395,14 @@ class TestBootstrap(Tester):
         node3.start(no_wait=True)
         time.sleep(3)
 
-        with tempfile.TemporaryFile(mode='w+') as tmpfile:
-            node1.stress(['user', 'profile=' + stress_config.name, 'ops(insert=1)',
-                          'n=500K', 'no-warmup', 'cl=LOCAL_QUORUM',
-                          '-rate', 'threads=5',
-                          '-errors', 'retries=2'],
-                         stdout=tmpfile, stderr=subprocess.STDOUT)
-            tmpfile.seek(0)
-            output = tmpfile.read()
+        out, err, _ = node1.stress(['user', 'profile=' + stress_config.name, 'ops(insert=1)',
+                                    'n=500K', 'no-warmup', 'cl=LOCAL_QUORUM',
+                                    '-rate', 'threads=5',
+                                    '-errors', 'retries=2'])
 
-        debug(output)
+        debug(out)
         regex = re.compile("Operation.+error inserting key.+Exception")
-        failure = regex.search(output)
+        failure = regex.search(out)
         self.assertIsNone(failure, "Error during stress while bootstrapping")
 
     @known_failure(failure_source='test',
--- a/commitlog_test.py
+++ b/commitlog_test.py
@@ -3,7 +3,6 @@ import glob
 import os
 import stat
 import struct
-import subprocess
 import time
 from distutils.version import LooseVersion
 
@@ -158,8 +157,8 @@ class TestCommitLog(Tester):
 
         self._change_commitlog_perms(0)
 
-        with open(os.devnull, 'w') as devnull:
-            self.node1.stress(['write', 'n=1M', 'no-warmup', '-col', 'size=FIXED(1000)', '-rate', 'threads=25'], stdout=devnull, stderr=subprocess.STDOUT)
+        # Use stress_process to skip internal error handling in ccm. Grep node logs for specific errors in test method.
+        self.node1.stress_process(['write', 'n=1M', 'no-warmup', '-col', 'size=FIXED(1000)', '-rate', 'threads=25']).communicate()
 
     @since('3.0.7')
     def test_mv_lock_contention_during_replay(self):
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -79,7 +79,7 @@ class TestCompaction(Tester):
         node1.flush()
 
         table_name = 'standard1'
-        output = node1.nodetool('cfstats', True)[0]
+        output = node1.nodetool('cfstats').stdout
         if output.find(table_name) != -1:
             output = output[output.find(table_name):]
             output = output[output.find("Space used (live)"):]
@@ -90,7 +90,7 @@ class TestCompaction(Tester):
 
         block_on_compaction_log(node1)
 
-        output = node1.nodetool('cfstats', True)[0]
+        output = node1.nodetool('cfstats').stdout
         if output.find(table_name) != -1:
             output = output[output.find(table_name):]
             output = output[output.find("Space used (live)"):]
@@ -133,7 +133,7 @@ class TestCompaction(Tester):
         node1.wait_for_compactions()
 
         table_name = 'standard1'
-        output = node1.nodetool('cfstats', True)[0]
+        output = node1.nodetool('cfstats').stdout
         output = output[output.find(table_name):]
         output = output[output.find("Bloom filter space used"):]
         bfSize = int(output[output.find(":") + 1:output.find("\n")].strip())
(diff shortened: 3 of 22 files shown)


bootstrap_test.py
```diff
@@ -283,9 +282,8 @@ def resumable_bootstrap_test(self):
 
         debug("Check data is present")
         # Let's check stream bootstrap completely transferred data
-        stdout, stderr = node3.stress(['read', 'n=100k', 'no-warmup', '-schema',
-                                       'replication(factor=2)', '-rate', 'threads=8'],
-                                      capture_output=True)
+
+        stdout = node3.stress(['read', 'n=100k', 'no-warmup', '-schema', 'replication(factor=2)', '-rate', 'threads=8'])
```
mambocab: The `stress` method will only ever return `None`:
https://github.com/smccarthy788/ccm/blob/1a3f940f503fd6559310703dffc83944afe815e0/ccmlib/node.py#L1240

mambocab: Assuming `stress` is changed to return the `(out, err, rc)` tuple, this assignment needs to be changed to account for that

bootstrap_test.py
```diff
@@ -397,18 +395,14 @@ def local_quorum_bootstrap_test(self):
...
-            tmpfile.seek(0)
-            output = tmpfile.read()
+        out, err, rc = node1.stress(['user', 'profile=' + stress_config.name, 'ops(insert=1)',
+                                     'n=500K', 'no-warmup', 'cl=LOCAL_QUORUM',
+                                     '-rate', 'threads=5',
+                                     '-errors', 'retries=2'])
 
-        debug(output)
+        debug(out)
         regex = re.compile("Operation.+error inserting key.+Exception")
```
mambocab: Maybe for a followup ticket: this assumes that errors are of the form `Operation .+error`, etc. Do we have a test for that?

commitlog_test.py
```diff
@@ -158,8 +157,8 @@ def _provoke_commitlog_failure(self):
 
         self._change_commitlog_perms(0)
 
-        with open(os.devnull, 'w') as devnull:
-            self.node1.stress(['write', 'n=1M', 'no-warmup', '-col', 'size=FIXED(1000)', '-rate', 'threads=25'], stdout=devnull, stderr=subprocess.STDOUT)
+        p = self.node1.stress_process(['write', 'n=1M', 'no-warmup', '-col', 'size=FIXED(1000)', '-rate', 'threads=25'])
```
mambocab: I don't understand why the previous behavior was the way it was, or why you've deliberately changed it here.

compaction_test.py
```diff
@@ -79,7 +79,7 @@ def data_size_test(self):
         node1.flush()
 
         table_name = 'standard1'
-        output = node1.nodetool('cfstats', True)[0]
+        output = node1.nodetool('cfstats')[0]
```
mambocab: I know this is literally all I do, but: this is why it'd be nice to return a `namedtuple` with fields `stdout`, `stderr`, and `returncode` (to conform with [the Popen object interface](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.stdin). This code could be `output = node1.nodetool('cfstats').stdout`, which is way more readable imo.

bootstrap_test.py
```diff
@@ -283,9 +282,8 @@ def resumable_bootstrap_test(self):
 
         debug("Check data is present")
         # Let's check stream bootstrap completely transferred data
-        stdout, stderr = node3.stress(['read', 'n=100k', 'no-warmup', '-schema',
-                                       'replication(factor=2)', '-rate', 'threads=8'],
-                                      capture_output=True)
+
+        stdout, stderr, rc = node3.stress(['read', 'n=100k', 'no-warmup', '-schema', 'replication(factor=2)', '-rate', 'threads=8'])
```
mambocab: I'm surprised that our linters aren't mad about this unused `rc` variable. Throughout, you can replace unused variables with `_`, just to have a placeholder variable to unpack into. But, if our linter's not mad, I don't know if that's worth the effort. Up to you.

commitlog_test.py
smccarthy788: The change was prompted by failure [1](https://cassci.datastax.com/job/ptnapoleon-trunk-dtest/7/testReport/commitlog_test/TestCommitLog/ignore_failure_policy_test/), [2](https://cassci.datastax.com/job/ptnapoleon-trunk-dtest/7/testReport/commitlog_test/TestCommitLog/stop_failure_policy_test/), [3](https://cassci.datastax.com/job/ptnapoleon-trunk-dtest/7/testReport/commitlog_test/TestCommitLog/die_failure_policy_test/), and [4](https://cassci.datastax.com/job/ptnapoleon-trunk-dtest/7/testReport/commitlog_test/TestCommitLog/stop_commit_failure_policy_test/), which are all the same failure really.

The tests using `_provoke_commitlog_failure` want some kind of failure to happen, and check for this failure by grepping a node's log for specific messages. We ignore output from `stress` by routing it to `devnull`, and print `stress`'s errors to stdout.

With the new API we were checking to see if `stress` ran into any failures, and raising a `ToolError` when it did. Since the tests grep the logs for errors, I  used `stress_process` to just skip error checking, since we check for the specific error later.

mambocab: I might be confused -- does the new `stress` method print to `stdout`? It'd be nice if it didn't IMO.

And your explanation makes sense. I this could be condensed to `self.node1.stress_process(<...>).wait()`, and should have a comment to document why we do that.

cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -549,7 +549,7 @@ def custom_null_indicator_template(self, indicator=None, copy_from_options=None)
         if indicator:
             cmds += " WITH NULL = '{d}'".format(d=indicator)
         debug(cmds)
-        self.node1.run_cqlsh(cmds=cmds, show_output=True, cqlsh_options=['--debug'])
+        self.node1.run_cqlsh(cmds=cmds, cqlsh_options=['--debug'])
```
mambocab: Dunno if it's important, but this is a behavioral change -- IIRC, `show_output` printed the `stdout`, so it might be good to do that

mambocab: But, I'm not super-attached to printing the output.

commitlog_test.py
smccarthy788: `stress` will always output to `subprocess.PIPE`, which will in turn be stored in `(out, err, rc)`.

cqlsh_tests/cqlsh_copy_tests.py
smccarthy788: So I changed it to assign `out` and then `debug(out)`. I think this is a happy middle ground.

offline_tools_test.py
```diff
@@ -34,18 +35,20 @@ def sstablelevelreset_test(self):
 
         # test by trying to run on nonexistent keyspace
         cluster.stop(gently=False)
-        (output, error, rc) = node1.run_sstablelevelreset("keyspace1", "standard1", output=True)
-        self.assertIn("ColumnFamily not found: keyspace1/standard1", error)
-        # this should return exit code 1
-        self.assertEqual(rc, 1, msg=str(rc))
+        try:
+            (output, error, rc) = node1.run_sstablelevelreset("keyspace1", "standard1")
```
mambocab: You can unwrap this tuple.

mambocab: Eh, never mind. Not worth the effort to clean up this file cosmetically.

commitlog_test.py
```diff
@@ -157,8 +157,7 @@ def _provoke_commitlog_failure(self):
 
         self._change_commitlog_perms(0)
 
-        p = self.node1.stress_process(['write', 'n=1M', 'no-warmup', '-col', 'size=FIXED(1000)', '-rate', 'threads=25'])
-        p.wait()
+        self.node1.stress_process(['write', 'n=1M', 'no-warmup', '-col', 'size=FIXED(1000)', '-rate', 'threads=25']).communicate()
```
mambocab: Nice. This should still have a comment explaining why we use this over the normal `stress` method.

mambocab:
+1 after my nits are addressed, looks good.

smccarthy788:
Had to make some changes to rebuild_test.py to deal with the rebase. Running a multiplex here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/194/

Once that's done, I think I'll merge.

mambocab:
Just to confirm, did you rebase onto current master to make sure you didn't miss anything since you created the branch?

mambocab:
Oh, you had to address changes in `rebuild_test` - yeah, those were the recent changes I was thinking of. :+1:, never mind!

smccarthy788:
@mambocab, I addressed your notes. Ended up changing a lot of files to deal with the unused variables, so I'm running one last CI check here:

http://cassci.datastax.com/view/Dev/view/smccarthy788/job/smccarthy788-trunk-dtest/2/

Assuming that comes back clean, I'll merge. After that there are three(?) follow up tickets to file?
1) For the bootstrap URI issue.
2) For the rebuild_test issue.
3) To test if the 
> errors are of the form `Operation .+error`

mambocab:
Sounds right, and looks great!

mambocab:
This needs to be rebased :/

PR #1149, opened 2016-07-28
add known failure annotation for CASSANDRA-12337

jkni:


scrub_test.py
```diff
@@ -254,6 +254,9 @@ def test_scrub_static_table(self):
         users = self.query_users(session)
         self.assertEqual(initial_users, users)
 
+    @known_failure(failure_source='cassandra',
+                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-12337',
+                   flaky=True)
```
ptnapoleon: It's with a heavy heart, that I tell you that unless you set a `notes` in this known_failure, you will see `notes='windows'` whenever we check this test programatically, because of `nosetests` reasons which are bad. So, you may want to set notes to some string here.

ptnapoleon:
+1, though with caveat

mambocab:
Merged, thanks!

PR #1150, opened 2016-07-28
default to empty notes field

mambocab:
This will make double-annotation with known_failure work as expected with the notes parameter - if you tag it twice, but omit notes in the outer decorator, the notes will be an empty string.

@knifewine can you review please?

### DIFF
--- a/tools.py
+++ b/tools.py
@@ -366,7 +366,7 @@ def require(require_pattern, broken_in=None):
         return tagging_decorator
 
 
-def known_failure(failure_source, jira_url, flaky=False, notes=None):
+def known_failure(failure_source, jira_url, flaky=False, notes=''):
     """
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Wow I REAALLY thought empty string for notes didn't work, or I would have done this already. What a bummer.

knifewine:
If these cases work fine then I'm +1:

single decorator with notes
single decorator without notes
double, outside has notes, inside no notes
double, outside no notes, inside has notes
double, both have notes
double, neither have notes

ptnapoleon:
Jim? Merging this would be nice.

mambocab:
Did a manual test as described in the new commit message. Read it, ya ingrates, it's a good lil manual test thing. It looks good, I'm merging.

PR #1153, opened 2016-07-29
add known_failure annotation

c-kodman:


c-kodman:
@mambocab if you could review this please

mambocab:
We'll keep an eye on this test, but I think that the issue described has already been resolved.

PR #1155, opened 2016-07-29
Tests for CASSANDRA-8523

pauloricardomg:


replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+        # insert initial data
+        node1.stress(['write', 'n=1000', "no-warmup", '-schema', 'replication(factor=3)'],
+                     whitelist=True)
+
+        # query and save initial data
+        session = self.patient_exclusive_cql_connection(node1)
+        session.default_timeout = 45
+        stress_table = 'keyspace1.standard1'
+        query = SimpleStatement('select * from %s' % stress_table, consistency_level=ConsistencyLevel.TWO)
+        expected_data = rows_to_list(session.execute(query))
```
mambocab: Style nit: these 3 lines could be condensed into

```python
query = SimpleStatement('select * from keyspace1.standard1', consistency_level=ConsistencyLevel.TWO)
expected_data = rows_to_list(session.execute(query))
```

or

```python
expected_data = rows_to_list(session.execute(
    SimpleStatement('select * from keyspace1.standard1', consistency_level=ConsistencyLevel.TWO)
))
```

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+            cluster.remove(node3)
+
+        # Upgrade only node2 to current version
+        if mixed_versions:
+            debug("Upgrading node2 to current version")
+            node2.stop(gently=True, wait_other_notice=True)
+            node2.set_install_dir(install_dir=default_install_dir)
+            node2.start(wait_other_notice=True, wait_for_binary_proto=True)
+            # Also change cluster dir so new node is started in new version
+            cluster.set_install_dir(install_dir=default_install_dir)
```
mambocab: Hmm. If this line is necessary, that looks like a CCM bug.

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+        # write additional data during replacement - if the new replace is used, extra writes should be
+        # forwarded to the replacement node
+        debug("Writing data to node1 - should replicate to replacement node")
+        node1.stress(['write', 'n=2k', "no-warmup", '-schema', 'replication(factor=3)'],
+                     whitelist=True)
+
+        # update expected_data if writes should be redirected to replacement node
+        if not same_address and not mixed_versions:
+            query = SimpleStatement('select * from %s' % stress_table, consistency_level=ConsistencyLevel.TWO)
+            expected_data = rows_to_list(session.execute(query))
```
mambocab: Since this logic is duplicated above, I think it's worth pulling out into a utility function.

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+            node2.stop(gently=True, wait_other_notice=True)
+            node2.set_install_dir(install_dir=default_install_dir)
+            node2.start(wait_other_notice=True, wait_for_binary_proto=True)
+            # Also change cluster dir so new node is started in new version
+            cluster.set_install_dir(install_dir=default_install_dir)
+
+        # start node4 on write survey mode so it does not finish joining ring
+        node4_address = '127.0.0.3' if same_address else '127.0.0.4'
+        debug("Starting node 4 to replace node 3 with address {}".format(node4_address))
+        node4 = Node('node4', cluster, True, (node4_address, 9160), (node4_address, 7000),
```
mambocab: I think you can use `tools.new_node` here. If you'd rather not, could you please use keyword arguments for the `Node` constructor for readability?

replace_address_test.py
mambocab: Ohhhh, I understand -- when you say `new node` in the comment, you mean `node4` below. Gotcha.

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+        if not same_address and not mixed_versions:
+            query = SimpleStatement('select * from %s' % stress_table, consistency_level=ConsistencyLevel.TWO)
+            expected_data = rows_to_list(session.execute(query))
+
+        debug("Joining replaced node")
+        node4.nodetool("join")
+
+        if not same_address:
+            for node in nodes_in_current_version:
+                node.watch_log_for("Node /127.0.0.4 will complete replacement of /127.0.0.3 for tokens", timeout=10)
```
mambocab: Could you, here and throughout, replace literal IP addresses with inserting `nodeX.address()` in the string? `.address` returns a string, so string formatting or concatenation is fine with me style-wise.

My logic with this request is that this will be a little more readable -- we'll know why we use those particular addresses -- and more robust to any future changes to how CCM assigns addresses.

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
         Tester.__init__(self, *args, **kwargs)
         self.allow_log_errors = True
 
+    def _test_insert_data_during_replace(self, same_address, mixed_versions=False):
```
mambocab: Could you add a docstring with an `@jira_ticket` annotation please?

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+            num_tokens = 1
+        else:
+            # a little hacky but grep_log returns the whole line...
+            num_tokens = int(node3.get_conf_option('num_tokens'))
+
+        debug("testing with num_tokens: {}".format(num_tokens))
+
+        # Change system_auth keyspace replication factor to 2, otherwise replace will fail
+        session = self.patient_cql_connection(node1)
+        update_auth_keyspace_replication(session)
```
mambocab: The use of this function might be clearer with a different name -- maybe `prepare_system_auth_table_for_replace` or similar?

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+        node4.start(jvm_args=["-Dcassandra.replace_address_first_boot=127.0.0.3",
+                              "-Dcassandra.write_survey=true"], wait_for_binary_proto=True,
+                    wait_other_notice=False)
+
+        if same_address:
+            node4.watch_log_for("Writes will not be forwarded to this node during replacement",
+                                timeout=60)
+
+        nodes_in_current_version = [node2, node4] if mixed_versions else [node1, node2, node4]
+        if not same_address:
```
mambocab: This `if` block and the one above would be clearer as an `if: ... else:` clause.

replace_address_test.py
```diff
@@ -38,6 +47,111 @@ def __init__(self, *args, **kwargs):
...
+            for node in nodes_in_current_version:
```
mambocab: Why don't we expect `node1` to have this error? Is this feature new since 2.2.4?

mambocab:
I keep getting lost while trying to read `_test_insert_data_during_replace`. I'm not sure what the issue is for me, but it may be helped by changing the names of the nodes to indicate their function, like `replaced_node` and `new_node`. I'll mess with it and push a commit for you to cherry-pick if I think it helps readability.

replace_address_test.py
```diff
@@ -611,3 +764,76 @@ def multi_dc_replace_with_rf1_test(self):
...
+        t.join()
+        self.assertFalse(node3.is_running())
+
+        node1.watch_log_for("FatClient /127.0.0.3 has been silent for 30000ms, removing from gossip", timeout=60)
+        node1.watch_log_for("Node /127.0.0.3 failed during replace.", timeout=60, filename='debug.log')
+
+        debug("replace failed - wiping node")
+
+        # wipe any data for node2
+        self._cleanup(node3)
```
mambocab: This line and the comment above it don't match.

replace_address_test.py
```diff
@@ -611,3 +764,76 @@ def multi_dc_replace_with_rf1_test(self):
...
+        # write some data, enough for the bootstrap to fail later on
+        debug("Inserting Data...")
+        node1.stress(['write', 'n=10K', 'no-warmup', '-schema', 'replication(factor=2)'])
+        cluster.flush()
+
+        session = self.patient_cql_connection(node1)
+        update_auth_keyspace_replication(session)
+        session.default_timeout = 45
+        stress_table = 'keyspace1.standard1'
+        query = SimpleStatement('select * from %s LIMIT 1' % stress_table, consistency_level=ConsistencyLevel.TWO)
```
mambocab: Why only check one value here?

Same comment as above, since this logic is repeated, this could be a utility function.

replace_address_test.py
```diff
@@ -611,3 +764,76 @@ def multi_dc_replace_with_rf1_test(self):
...
+        initial_data = rows_to_list(session.execute(query))
+
+        node2.stop(gently=False)
+
+        # Add a new node to replace node3 to ensures that it is not a seed
+        node3 = Node('node3', cluster, True, ('127.0.0.3', 9160), ('127.0.0.3', 7000), '7400', '0',
```
mambocab: Same comment as above -- this could be changed to use `tools.new_node`, or to use keyword arguments.

mambocab:
> I'll mess with it and push a commit for you to cherry-pick if I think it helps readability.

Here's my branch with renamed node variables: https://github.com/riptano/cassandra-dtest/compare/master...mambocab:pull-1155_rename-nodes

I think it helps track the uses of nodes through the test, and I've had Russ give it a once-over, so I'm pretty confident it helps readability. If you agree, could you cherry-pick it onto this PR?

replace_address_test.py
mambocab: I remember you mentioning this elsewhere -- is this necessary if you don't remove `node3` from the ccm cluster?

replace_address_test.py
```diff
@@ -207,12 +344,12 @@ def replace_first_boot_test(self):
         debug("Verifying tokens migrated sucessfully")
         moved_tokens = node4.grep_log("Token .* changing ownership from /127.0.0.3 to /127.0.0.4")
         debug("number of moved tokens: {}".format(len(moved_tokens)))
-        self.assertEqual(len(moved_tokens), num_tokens)
+        self.assertTrue(len(moved_tokens) % num_tokens == 0)
```
mambocab: This should use `self.assertEqual` rather than `self.assertTrue(x == y)`.

Is there some particular number of log lines we expect? If so, this should be more like

```python
self.assertEqual(len(moved_tokens), n * num_tokens)
```

mambocab:
Looks great, mostly style nits from me. Let me know if you want me to take on any particular changes and I'll provide a commit to cherry-pick. Thanks for the tests!

pauloricardomg:
Addressed comments as well as did a refactoring to improve reuse and efficiency of replace_address_test. The most relevant changes are:
* Extracted common methods between tests: `_setup`, `_do_replace`, `_stop_node_to_replace`, `_insert_data`, `_fetch_initial_data`, `_verify_replacement`, `_verify_data`, `_wait_for_replacement`, `_verify_tokens_migrated_successfully`
* Extracted cluster nodes to class variables: `self.query_node`, `self.replaced_node` and `self.replaced_node`
* Made suite faster and more deterministic with:
 * Reduce number of keys inserted/queried with stress
 * Use byteman to interrupt streaming
 * Reduce cassandra.ring_delay_ms from default of 30s to 10s and cassandra.broadcast_interval_ms from default of 60s to 10s.
* Removed `Try to restart node 3 (should fail)` check since it's not meaningful, and filled [CASSANDRA-12485](https://issues.apache.org/jira/browse/CASSANDRA-12485) for a proper check

I removed the `@attr('resource-intensive')` annotations so test are not ignore on CI:
* [2.1](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-2.1-8523-dtest/3/) 
* [2.2](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-2.2-8523-dtest/16/)
* [3.0](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-3.0-8523-dtest/8/)
* [trunk](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunk-8523-dtest/11/)

pauloricardomg:
@mambocab friendly ping in case you missed the notification, since both the ticket and this are quite large so rebases can get painful. I will try to be faster in the next follow-up. :)

mambocab:
Thanks for the ping! Reviewing now.

byteman/stream_failure.btm
```diff
@@ -10,7 +10,7 @@ METHOD prepare
 AT INVOKE startStreamingFiles
 BIND peer = $0.peer
 # set flag to only run this rule once.
-IF peer.equals(InetAddress.getByName("127.0.0.3")) AND NOT flagged("done")
+IF NOT flagged("done")
```
mambocab: Will this affect how this script is used in `boostrap_test`?

https://github.com/riptano/cassandra-dtest/blob/54abff9296191aa0b47c4ce871594cf06feb65d8/bootstrap_test.py#L255

replace_address_test.py
```diff
@@ -19,232 +18,304 @@ class NodeUnavailable(Exception):
...
+
+    @known_failure(failure_source='systemic',
+                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11652',
+                   flaky=True,
+                   notes='windows')
+    def replace_stopped_node_test(self):
+        """
+        Test that we can replace a node that is not shutdown gracefully.
+        """
+        self._test_replace_node()
```
mambocab: Nit: could you pass `gently=False` here, just for clarity and in case the `_test_replace_node` defaults change?

replace_address_test.py
```diff
@@ -19,232 +18,304 @@ class NodeUnavailable(Exception):
...
+
+        # we redo this check because restarting node should not result
+        # in tokens being moved again, ie number should be same
+        if not same_address:
+            self._verify_tokens_migrated_successfully(previous_log_size)
+
+        self._verify_data(initial_data)
+
+    def replace_active_node_test(self):
+        self.ignore_log_patterns.extend([r'Exception encountered during startup'])
```
mambocab: This pattern actually modifies the class's `ignore_log_patterns` :/

```python
In [1]: class ModifyAttributeDemo(object):
   ...:     xs = ['initial population']
   ...:     

In [2]: m0 = ModifyAttributeDemo()

In [3]: m0.xs.extend(['added to m0'])

In [4]: m1 = ModifyAttributeDemo()

In [5]: m1.xs
Out[5]: ['initial population', 'added to m0']
```

This means that the patterns you ignore on one test will be ignored on subsequent tests. The way to work around this is to do something like

```python
self.ignore_log.patterns = list(self.ignore_log_patterns) + [...]
```

or

```python
self.ignore_log.patterns = self.ignore_log_patterns[:] + [...]
```

I have a branch that, in theory, will keep us from doing this in the future, but it's fairly low-priority, and breaks stuff for reasons I don't understand, so I haven't pursued it further:

https://github.com/riptano/cassandra-dtest/compare/master...mambocab:immutable-ignore_log_patterns#diff-4cbdcd9896ce5009acd37ed830ce8959R308

replace_address_test.py
```diff
@@ -19,232 +18,304 @@ class NodeUnavailable(Exception):
...
-        Check that tokens are migrated and that data is replicated properly.
-        """
-        debug("Starting cluster with 3 nodes.")
-        cluster = self.cluster
-        cluster.populate(3).start()
-        node1, node2, node3 = cluster.nodelist()
+        n = len(self.cluster.nodelist())  # n may be a list of nodes per dc
+        self.cluster.set_batch_commitlog(enabled=True)
+        self.query_node = self.cluster.nodelist()[0]
+        self.replaced_node = self.cluster.nodelist()[n - 1]
```
mambocab: Are you trying to get the last element here? You can use negative indices to do that:

```python
>>> [1, 2, 3, 4][-1]
4
```

If you do so, looks like you could remove the assignment of `n` a few lines above.

replace_address_test.py
```diff
@@ -311,240 +372,162 @@ def unsafe_replace_test(self):
...
+                                        wait_for_binary_proto=True)
+            # check if we reset bootstrap state
+            self.replacement_node.watch_log_for("Resetting bootstrap progress to start fresh", from_mark=mark)
+        elif mode == 'resume':
+            debug("Resuming failed bootstrap")
+            self.replacement_node.nodetool('bootstrap resume')
+            # check if we skipped already retrieved ranges
+            self.replacement_node.watch_log_for("already available. Skipping streaming.")
+            self.replacement_node.watch_log_for("Resume complete")
+        else:  # mode == 'wipe'
```
mambocab: I think I'd prefer this be `elif mode == 'wipe'` and have an

```python
else:
    raise RuntimeError('invalid mode value {mode}'.format(mode))
```

block below, just to make sure things are called correctly in the future.

mambocab:
What are your thoughts on this failure? http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunk-8523-dtest/12/testReport/junit/replace_address_test/TestReplaceAddress/unsafe_replace_test/

mambocab:
The other builds look clean though. Let me know what you think of that error. Most of my other comments are nits.

Once you address my `ignore_log_errors` comment, could you kick off another set of runs on cassci?

byteman/stream_failure.btm
pauloricardomg: no, because in that case the script is only submitted to node1, which will only stream to the bootstrapping node (node3), so the script will have the same effect without the IP filter.

pauloricardomg:
Thanks for the follow-up. The `unsafe_replace_test` [failure](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunk-8523-dtest/12/testReport/junit/replace_address_test/TestReplaceAddress/unsafe_replace_test/) was due to waiting other's node notice in the refactor, which was not present in the original test, so I fixed this.

I updated branch with suggested nits and resubmitted tests:
* [2.1unpatched](http://cassci.datastax.com/job/pauloricardomg-2.2-8523-dtest/18/)
* [2.2](http://cassci.datastax.com/job/pauloricardomg-2.2-8523-dtest/18/)
* [3.0](http://cassci.datastax.com/job/pauloricardomg-3.0-8523-dtest/11/)
* [trunk](http://cassci.datastax.com/job/pauloricardomg-trunk-8523-dtest/14/)

ptnapoleon:
@mambocab this is +1 after travis nit, yes?

mambocab:
Yep, +1. Thanks, Paulo!

iamaleksey:
Patch committed.

ptnapoleon:
Paulo, can you rebase and merge? Thanks

pauloricardomg:
Rebased, now merging. Thanks all!

PR #1156, opened 2016-08-01
add known_failure annotation

c-kodman:
@mambocab


mambocab:
It's looking like this may be a cross-cutting C* issue that could fail basically any test. I'm going to close this for now.

PR #1157, opened 2016-08-01
Ignore log error cassandra 12340

mambocab:
@knifewine Could I get a quick review? This should successfully ignore the error in [CASSANDRA-12340](https://issues.apache.org/jira/browse/CASSANDRA-12340), yeah?

I forgot to remove the `known_failure` annotation, so I'll need to do that before merge, if we merge this.

### DIFF
--- a/upgrade_supercolumns_test.py
+++ b/upgrade_supercolumns_test.py
@@ -11,7 +11,7 @@ from thrift_bindings.v22.ttypes import (CfDef, Column, ColumnOrSuperColumn,
                                         Mutation, SlicePredicate, SliceRange,
                                         SuperColumn, TimedOutException)
 from thrift_tests import get_thrift_client
-from tools import RerunTestException, requires_rerun, since, known_failure
+from tools import RerunTestException, known_failure, requires_rerun, since
 
 
 @since('2.0', max_version='2.1.x')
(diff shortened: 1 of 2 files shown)


knifewine:
+1

mambocab:
Ping @knifewine for re-review -- I've added a new commit moving the ignore pattern to apply to all (or at least more) upgrade tests. I'm not sure where else I might put this.

upgrade_supercolumns_test.py
```diff
@@ -27,6 +28,14 @@ def __init__(self, *args, **kwargs):
...
             r'Can\'t send migration request: node.*is down',
         ]
+        if LooseVersion(CASSANDRA_VERSION_FROM_BUILD) < '2.2':
+            _known_teardown_race_error = (
+                'ScheduledThreadPoolExecutor$ScheduledFutureTask@[0-9a-f]+ '
+                'rejected from org.apache.cassandra.concurrent.DebuggableScheduledThreadPoolExecutor'
+            )
+            # don't alter ignore_log_patterns on the class, just the obj for this test
+            self.ignore_log_patterns += [_known_teardown_race_error]
+
```
knifewine: won't this cause the class mutable list to get appended multiple times with the same _known_teardown_race_error, once each time __init__ is called?

mambocab: No, `self.ignore_log_patterns` is set anew each `__init__`, a couple lines above

knifewine:
+1

mambocab:
Thanks, merging.

PR #1158, opened 2016-08-01
CASSANDRA-11687

yukim:
First of all, my fix in #1146 was not doing the job. Even though nodetool is invoked in background first, the second call can be invoked first. So I reverted the changes made.

Now, I noticed that we were trying to spawn Thread to run rebuild in parallel, but actually it wasn't. The first rebuild is called before the thread is created, resulting sequential rebuild call instead of parallel.

@mambocab @pauloricardomg can you run multiplexer on this one?

### DIFF
--- a/rebuild_test.py
+++ b/rebuild_test.py
@@ -78,21 +78,55 @@ class TestRebuild(Tester):
         session.execute("ALTER KEYSPACE system_auth WITH REPLICATION = {'class':'NetworkTopologyStrategy', 'dc1':1, 'dc2':1};")
         session.execute('USE ks')
 
-        # rebuild dc2 from dc1 in background
-        mark = node2.mark_log()
-        node2.nodetool('rebuild dc1', False, False)
+        self.rebuild_errors = 0
+        # make it possible for thread to raise errors
+        self.thread_exc_info = None
+
+        # rebuild dc2 from dc1
+        def rebuild():
+            try:
+                node2.nodetool('rebuild dc1')
+            except NodetoolError as e:
+                if 'Node is still rebuilding' in e.message:
+                    self.rebuild_errors += 1
+                else:
+                    raise e
+
+        def call_and_register_exc_info_on_exception(func):
+            """
+            Closes over self to catch any exceptions raised by func and
+            register them at self.thread_exc_info
+            Based on http://stackoverflow.com/a/1854263
+            """
+            try:
+                func()
+            except Exception:
+                import sys
+                self.thread_exc_info = sys.exc_info()
+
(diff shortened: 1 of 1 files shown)


pauloricardomg:
Submitted multiplexer [100x run](https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/207/parameters/)

rebuild_test.py
```diff
@@ -77,22 +78,60 @@ def simple_rebuild_test(self):
...
+                """
+                Closes over self to catch any exceptions raised by func and
+                register them at self.thread_exc_info
+                Based on http://stackoverflow.com/a/1854263
+                """
+                try:
+                    self.func()
+                except Exception:
+                    import sys
+                    self.thread_exc_info = sys.exc_info()
```
mambocab: Here, `self` is the `Runner` object, not the test case. As a result, below, `self.thread_exc_info` will always be `None`.

mambocab:
Could you explain your last commit? I don't see the problem with the previous code, or how your commit fixes it?

yukim:
When we did:
```python
cmd1 = Thread(target=call_and_register_exc_info_on_exception(rebuild))
```
before, Thread is created _after_ we run `rebuild()`, since `call_and_register_exec_info_on_exception` runs it before passing to Thread constructor.
Thus, we were executing `rebuild` sequentially, which was not intended.

pauloricardomg:
multiplexer job failed for some unknown reason, but it seems [from the logs](https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/207/artifact/test_stdout.000.txt) it was able to run 92x successfully, so I think this should be ok.

+1, can you just change the CL of the last query to `LOCAL_ONE` and remove the known_failure annotation?

Thanks

mambocab:
> Thread is created after we run rebuild(), since call_and_register_exec_info_on_exception runs it before passing to Thread constructor.

:doh: Yep, I see it now, thanks.

https://github.com/riptano/cassandra-dtest/pull/1158#discussion_r73198419 < this comment still needs to be addressed, though. The error-checking code:

```python
if self.thread_exc_info is not None:
    raise self.thread_exc_info[1], None, self.thread_exc_info[2]
```

will never find errors raised in the thread. This could be fixed by just removing `thread_exc_info` from the test object and using `cmd1.thread_exc_info` instead of `self.thread_exc_info`.

pauloricardomg:
@yukim friendly ping in case you missed my comment above ;-)

pauloricardomg:
+1, thanks!

mambocab:
Thanks to both of you for the work on this test!

PR #1162, opened 2016-08-02
add log ignore for RejectedExecution during upgrade

knifewine:
for cassandra-12364

### DIFF
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -45,6 +45,7 @@ class UpgradeTester(Tester):
         self.ignore_log_patterns = [
             # Normal occurance. See CASSANDRA-12026. Likely won't be needed after C* 4.0.
             r'Unknown column cdc during deserialization',
+            r'RejectedExecutionException.*ThreadPoolExecutor has shut down',  # see  CASSANDRA-12364
         ]
         super(UpgradeTester, self).__init__(*args, **kwargs)
 
--- a/upgrade_tests/cql_tests.py
+++ b/upgrade_tests/cql_tests.py
@@ -1971,9 +1971,6 @@ class TestCQL(UpgradeTester):
             query = "SELECT blog_id, timestamp FROM test WHERE author = 'bob'"
             assert_all(cursor, query, [[1, 0], [1, 3], [0, 0]])
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-12364',
-                   flaky=True)
     def refuse_in_with_indexes_test(self):
         """ Test for the validation bug of #4709 """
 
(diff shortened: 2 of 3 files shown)


knifewine:
@mambocab to review, please.

upgrade_tests/upgrade_base.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, *args, **kwargs):
         self.ignore_log_patterns = [
             # Normal occurance. See CASSANDRA-12026. Likely won't be needed after C* 4.0.
             r'Unknown column cdc during deserialization',
+            r'RejectedExecutionException.*ThreadPoolExecutor has shut down',  # see  CASSANDRA-12364
```
mambocab: I don't know where these exceptions might get raised, so I'm worried that it'd happen in more places than just teardown. Maybe append `.*at org.apache.cassandra.db.ColumnFamilyStore.waitForFlushes`? It sounds like that's the place where this happens.

mambocab:
I couldn't tell from @pcmanus's comment on the Jira ticket whether this was an upgrade-specific error or not. So, this lgtm after my nit, but might not cover all the cases we want it to. @pcmanus, should this error be ignored everywhere, or just in upgrade tests?

upgrade_tests/upgrade_base.py
knifewine: I'm actually unclear as to whether we should ignore any RejectedExecutionException or try to be more specific like you suggested. I wasn't getting this to repro locally so not 100% if the more specific case would have us covered (I also forget whether we can span multiple lines with the ignore regex, if that will work). If we are in fact being too general then the same pattern in upgrade_through_versions_test.py will need a fixup too.

knifewine: note on the u_t_v tests: looks like CASSANDRA-7727 was what prompted a similar pattern to be ignored in those tests. Looks to be a RejectedExecutionException but with a different backtrace than the one noted on 12364.

ptnapoleon:
Bump, @mambocab. Since Sylvain is out, could you investigate with someone else?

knifewine:
We're still seeing failures on CI due to not ignoring logs properly. 

http://cassci.datastax.com/view/cassandra-3.9/job/cassandra-3.9_dtest_upgrade/lastCompletedBuild/testReport/upgrade_tests.paging_test/TestPagingWithDeletionsNodes3RF3_Upgrade_current_3_x_To_indev_3_x/test_single_cell_deletions/

Unfortunately, this failure would not be caught by the more specific pattern suggested above, and I still think the more generous regex in my pr would be fine here. At the very least maybe we can merge in this version and spin off another issue to investigate making things more specific at a later time.

Getting this in would help get the CI failure baseline down even more and get rid of these as lingering issues (which are really just noise).

knifewine:
@ptnapoleon @mambocab bump. thoughts here?

pcmanus:
It's not an upgrade specific thing and could be triggered anytime a node is shutdown (gently). Basically, it's a race in C* gentle shutdown with no real consequence other that log pollution. It's still pollution and we should fix it where it happens, and we have already done so in other places in the past, but we never spend the cycles finding all the places where this can happen and fixing them. And for the reasons above, it's not high on our priority list. Hence the suggestion of ignoring for test purposes.

As for when to ignore, ideally we should ignore it every time we shut down a node gently, but as I imagine it might be a pain to do that exactly, I don't think it's we're taking much risk in just ignoring it all the time (note that I'm mostly saying to ignore in term of failing the whole test, we could still log some form of warning so we have some reminder that this is not ideal).

ptnapoleon:
I'm +1. I think the risk of this covering up a different error is small enough.

PR #1163, opened 2016-08-03
add known_failure annotation

c-kodman:
@mambocab


mambocab:
Let's let this one marinate a little longer -- looks like the underlying ticket may be closed soon.

PR #1164, opened 2016-08-03
add known_failure annotation

c-kodman:
@mambocab


mambocab:
Let's also let this sit for the moment, as it may be closed after tonight's 2.2 run.

PR #1165, opened 2016-08-03
Update INSTALL.md for most recent method of installing CQL

jkni:
This is how we install on CI. The Google code repo no longer exists.


knifewine:
+1

PR #1168, opened 2016-08-04
wait for keyspace1 to be initialized before running repair

krummas:


repair_tests/incremental_repair_test.py
```diff
@@ -49,8 +46,12 @@ def sstable_marking_test(self):
         node2.flush()
 
         node3.start(wait_other_notice=True)
+        if node3.get_cassandra_version() < '2.2':
+            log_file = 'system.log'
+        else:
+            log_file = 'debug.log'
+        node3.watch_log_for("Initializing keyspace1.standard1", filename=log_file)
         time.sleep(3)
```
mambocab: Could this `sleep` be unnecessary after this change?

repair_tests/incremental_repair_test.py
```diff
@@ -49,8 +46,12 @@ def sstable_marking_test(self):
         node2.flush()
 
         node3.start(wait_other_notice=True)
+        if node3.get_cassandra_version() < '2.2':
+            log_file = 'system.log'
+        else:
+            log_file = 'debug.log'
+        node3.watch_log_for("Initializing keyspace1.standard1", filename=log_file)
```
mambocab: @ptnapoleon and/or @knifewine do we think that log-watching is the best way to block on keyspace init? I feel like it might be better to do a spin lock on the local schema, but I'm not sure.

ptnapoleon: If we're sure this log message only occurs after the appropriate schema state, then this is fine. I'm indifferent

repair_tests/incremental_repair_test.py
krummas: I'd rather keep it - I reduced it to 1 second

repair_tests/incremental_repair_test.py
mambocab: @krummas can you confirm that this message is logged after initialization is complete?

krummas: as I mentioned, there is some less grep friendly output after this, that is why I want to keep the sleep

pauloricardomg:
+1

PR #1170, opened 2016-08-04
CASSANDRA-12039 Added tests for pre-join 2i callback.

sbtourist:


secondary_indexes_test.py
```diff
@@ -1019,3 +1019,101 @@ def upgrade_to_version(self, tag, nodes=None):
...
+            self.assertTrue(len(node2.grep_log('Startup complete, but write survey mode is active, not becoming an active ring member.')))
+            self.assertFalse(len(node2.grep_log('Executing pre-join')))
+
+            node2.nodetool("join")
+            self.assertTrue(len(node2.grep_log('Leaving write survey mode and joining ring at operator request')))
+            self.assertTrue(len(node2.grep_log('Executing pre-join post-bootstrap tasks')))
+
+        self._base_test(write_survey_and_join)
+
+
```
ptnapoleon: You have loose whitespace down here, can you remove that?

secondary_indexes_test.py
```diff
@@ -1019,3 +1019,101 @@ def upgrade_to_version(self, tag, nodes=None):
...
+        def write_survey_and_join(cluster, token):
+            node2 = new_node(cluster)
+            node2.set_configuration_options(values={'initial_token': token})
+            node2.start(jvm_args=["-Dcassandra.write_survey=true"], wait_for_binary_proto=True)
+            self.assertTrue(len(node2.grep_log('Startup complete, but write survey mode is active, not becoming an active ring member.')))
+            self.assertFalse(len(node2.grep_log('Executing pre-join')))
+
+            node2.nodetool("join")
+            self.assertTrue(len(node2.grep_log('Leaving write survey mode and joining ring at operator request')))
+            self.assertTrue(len(node2.grep_log('Executing pre-join post-bootstrap tasks')))
```
ptnapoleon: Do we even need to take the `len()` here? I think `None` is returned if nothing is found, which is also falsey.

ptnapoleon:
@mambocab , I like this, other than my nits. What do you think?

secondary_indexes_test.py
mambocab: `grep_log` always returns a list: https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L345-L352

We'd get better error reporting with `self.assertTrue(node2.grep_log(...))` or `self.assertNotEqual([], node2.grep_log(...))`.

secondary_indexes_test.py
```diff
@@ -1019,3 +1019,101 @@ def upgrade_to_version(self, tag, nodes=None):
...
+
+    def __init__(self, *args, **kwargs):
+        # Ignore these log patterns:
+        self.ignore_log_patterns = [
+            # Ignore streaming error during bootstrap
+            r'Exception encountered during startup',
+            r'Streaming error occurred'
+        ]
+        Tester.__init__(self, *args, **kwargs)
+        self.allow_log_errors = True
```
mambocab: Is this intentional? Why use both `ignore_log_patterns` and `allow_log_errors`?

mambocab:
These look very good on the whole. I'm confused about the log-ignoring configuration, and I think the assertions involving `len` should be changed as I noted in [this comment](https://github.com/riptano/cassandra-dtest/pull/1170#discussion_r73590198), but address those and I'm +1.

secondary_indexes_test.py
sbtourist: Sure.

secondary_indexes_test.py
sbtourist: I just followed the same idiom used in other dtests. I'll change as suggested.

mambocab: Oof, good to know. We should fix those. Sorry about the bad style elsewhere.

secondary_indexes_test.py
sbtourist: My mistake, let me remove `allow_log_errors`.

sbtourist:
@mambocab, @ptnapoleon,

I've applied your suggestions and run another round, but got many apparently unrelated failures: https://cassci.datastax.com/view/Dev/view/sbtourist/job/sbtourist-CASSANDRA-12039-3.0-dtest/11

Any ideas?

ptnapoleon:
Rebasing your branch onto master should fix the issue

ptnapoleon:
-1 while resume_test continues to hang

mambocab:
This also needs to be rebased onto the restructured `tools` module.

ptnapoleon:
Closing in favor of #1316

PR #1088, opened 2016-07-12
Fix enabling of jmx auth on windows

beobal:

PR #1175, opened 2016-08-04
Fix enabling of jmx auth on windows

jkni:
PR for issue at [CASSANDRA-11730](https://issues.apache.org/jira/browse/CASSANDRA-11730). This edits the Windows config files appropriately to set up JMX auth.

### DIFF
--- a/jmx_auth_test.py
+++ b/jmx_auth_test.py
@@ -4,15 +4,12 @@ from ccmlib.node import ToolError
 
 from dtest import Tester
 from jmxutils import apply_jmx_authentication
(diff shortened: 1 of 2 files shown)


mambocab:
This is a resubmission of #1088? It looks fine and passes on my local Linux env. I haven't run it in a Windows env.

jkni:
Depends on what you mean by resubmission - it is a revised version of that same PR, but the content is slightly different.

mambocab:
Ah yeah, fair. But, I'm +1 if you're confident it works on Windows.

PR #1176, opened 2016-08-04
special case for 2.2/3.0 mixed version cluster paging

knifewine:
Due to CASSANDRA-10880 there's a special case that needs to be handled with protocol version.

This mixed cluster combination of 2.2/3.0 needs to work over proto v3, rather than v4 (despite that fact that both support v4).

### DIFF
--- a/upgrade_tests/paging_test.py
+++ b/upgrade_tests/paging_test.py
@@ -187,6 +187,13 @@ class PageAssertionMixin(object):
 class BasePagingTester(UpgradeTester):
 
     def prepare(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


knifewine:
@mambocab to review.

ptnapoleon:
Do we need to do this for 2.1/3.0 as well?

knifewine:
@ptnapoleon I think 2.1 has a max proto of v3, so I don't think it would be helpful there?

ptnapoleon:
Makes sense, ill let jim get back to review

mambocab:
LGTM, +1

knifewine:
thanks guys.

PR #1177, opened 2016-08-04
page wait() small improvements

knifewine:
Add a small loop sleep, and improve debug message when paging times out.

I spotted these small differences comparing the regular paging tests to the upgrade paging tests, and thought they could be useful in the upgrade suite.

At some point we should extract the paging utility classes to a single place, so that both paging suites can pull it in without code duplication, so this is a small band-aid for now.


knifewine:
@ptnapoleon to review please.

ptnapoleon:
+1

PR #1178, opened 2016-08-05
Fix tests post-CASSANDRA-10783

pcmanus:


ptnapoleon:
This test is still going to run on versions without 10783, so we need to filter the expected error message based on cluster version

pcmanus:
You're right, updated the patch to deal with that.

ptnapoleon:
+1, Thanks

PR #1184, opened 2016-08-05
remove redundant argument

smccarthy788:
@ptnapoleon or @mambocab or @knifewine, for your consideration.

The `--node` argument is redundant here, as it is specified inside of ccmlib/node.py here:
https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L1226-L1229

Running a quick multiplex here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/218/


ptnapoleon:
+1 if CI is happy

knifewine:
+1

PR #1185, opened 2016-08-08
Fix stress profile in local_quorum_bootstrap

ptnapoleon:
Moved all stress commands within the context management
of the profile tmp file. Previously we deleted file,
then attempted to read from it again
@smccarthy788 to review


smccarthy788:
+1

PR #1190, opened 2016-08-08
special case proto version for paging in 2.1/3.0 clusters

knifewine:


knifewine:
@ptnapoleon was correct on #1176 that we need to handle 2.1/3.0 clusters with this logic as well. This is because using proto v4 on the 3.0 nodes when mixed with 2.1 will still cause paging problems.

ptnapoleon:
+1

knifewine:
thanks @ptnapoleon !

PR #1191, opened 2016-08-08
add large attribute to test methods

smccarthy788:
@ptnapoleon, @mambocab, please review.

### DIFF
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -13,6 +13,7 @@ from dtest import DISABLE_VNODES, Tester, debug, MultiError
 from tools import (create_c1c2_table, insert_c1c2, insert_columns,
                    known_failure, query_c1c2, rows_to_list, since)
 from nose.tools import assert_greater_equal
+from nose.plugins.attrib import attr
 
 
 class TestHelper(Tester):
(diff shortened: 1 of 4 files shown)


mambocab:
It worries me that Travis doesn't see the problem here -- you need to import `nose.plugins.attrib.attr` before using it.

mambocab:
Does trying to import the changed modules work?

mambocab:
Also:

\*rap air horns\* **BIKESHED TIIIIIME**

@ptnapoleon @knifewine is `large` the name we want to use for this class of test? We could also do something like `many-nodes` or `large-instance-size`.

ptnapoleon:
Do we need it to be more specific than large? I don't care between the choices presented, honestly

mambocab:
I'm of 2 minds. We don't want vocabulary that only people in DS TE understand, but we also don't want need to be completely explicit all the time. I think I'm +0.5 for a longer one, no real preference between the longer options. @knifewine tiebreaker?

knifewine:
Yeah, 'large' is pretty ambiguous. I lean towards large-instance-size, but it's still a bit confusing since tests can may get run locally by people in which case 'instance' is a weird word too.

knifewine:
large-instance-size or maybe resource-intensive-test ?

mambocab:
I like `resource-intensive`. Any objections to having Sean use that?

ptnapoleon:
+1 to resource-intensive

mambocab:
Cool. This needs to be rebased but then we can merge :+1:

ptnapoleon:
+1

PR #1192, opened 2016-08-09
Add tests for decommission

kdmu:
Add test for testing concurrent decommission is not allowed and decommission is resumable. 
Byteman script injection is used here.

### DIFF
--- a/topology_test.py
+++ b/topology_test.py
@@ -50,6 +50,96 @@ class TestTopology(Tester):
         # described in 9912. Do not remove it.
         time.sleep(10)
 
+    def concurrent_decommission_not_allowed_test(self):
+        """
+        Test concurrent decommission is not allowed
+        """
+        cluster = self.cluster
+        cluster.set_configuration_options(values={'stream_throughput_outbound_megabits_per_sec': 1})
+        cluster.populate(2).start(wait_other_notice=True)
+        node1, node2 = cluster.nodelist()
+
+        session = self.patient_cql_connection(node2)
+        self.create_ks(session, 'ks', 1)
+        self.create_cf(session, 'cf', columns={'c1': 'text', 'c2': 'text'})
+        insert_c1c2(session, n=10000, consistency=ConsistencyLevel.ALL)
+
+        mark = node2.mark_log()
+
+        def decommission():
+            node2.nodetool('decommission')
+
+        # Launch first decommission in a external thread
+        t = Thread(target=decommission)
+        t.start()
+
+        # Make sure first decommission is initialized before second decommission
+        node2.watch_log_for('DECOMMISSIONING', filename='debug.log')
+
+        # Launch a second decommission, should fail
+        with self.assertRaises(ToolError):
+            node2.nodetool('decommission')
+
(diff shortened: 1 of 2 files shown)


ptnapoleon:
@knifewine, do you want to review?

topology_test.py
```diff
@@ -50,6 +50,89 @@ def simple_decommission_test(self):
...
+        node2.watch_log_for('DECOMMISSIONED', from_mark=mark)
+        session = self.patient_cql_connection(node1)
+        session.execute('USE ks')
+        for n in xrange(0, 10000):
+            query_c1c2(session, n, ConsistencyLevel.ONE)
+
+    @since('3.10')
+    def resumable_decommission_test(self):
+        """
+        @jira-ticket CASSANDRA-12008
```
ptnapoleon: Can you make this `@jira_ticket`?

topology_test.py
```diff
@@ -50,6 +50,89 @@ def simple_decommission_test(self):
...
+        node1, node2, node3 = cluster.nodelist()
+
+        session = self.patient_cql_connection(node2)
+        self.create_ks(session, 'ks', 2)
+        self.create_cf(session, 'cf', columns={'c1': 'text', 'c2': 'text'})
+        insert_c1c2(session, n=10000, consistency=ConsistencyLevel.ALL)
+
+        # Execute first rebuild, should fail
+        with self.assertRaises(ToolError):
+            script = ['./decommission_failure_inject.btm']
```
ptnapoleon: Can you put the `.btm` file in a subdirectory named `byteman`? We'd like to start moving these there, to prevent cluttering the top-level

topology_test.py
```diff
@@ -50,6 +50,89 @@ def simple_decommission_test(self):
...
+        # Decommission again
+        mark = node2.mark_log()
+        node2.nodetool('decommission')
+
+        # Check decommision is done and we skipped transfereed ranges
+        node2.watch_log_for('DECOMMISSIONED', from_mark=mark)
+        node2.grep_log("Skipping transferred range .* of keyspace ks, endpoint /127.0.0.3", filename='debug.log')
+
+        # Check data is correctly forwarded to node1 and node3
+        session = self.patient_cql_connection(node1)
```
ptnapoleon: Both this, and the next session need to use `patient_exclusive_cql_connection`, if you only want them querying from the given node. Additionally, that still doesn't prevent other nodes from giving the responses, even at `CL.ONE`, so you should take down the other nodes in the cluster, when verifying the data on only one node

topology_test.py
kdmu: Sorry, you mean add "@jira_ticket" to cocurrent_not_allowed_test?

ptnapoleon: I just meant, to edit the `jira-ticket` to `jira_ticket` on this line.

kdmu:
@ptnapoleon Changed data check-up procedure and tested with trunk, it seems work. Please let me know if there's any error, thanks.

ptnapoleon:
+1 Thanks @kdmu !

PR #1193, opened 2016-08-09
Added missing known_failure import

stef1927:
`cql_tracing_test.py` currently doesn't run because `known_failure` was added without an import for it.


ptnapoleon:
I pushed this fix before I saw your PR. In the future, feel completely free to push these kind of things to directly to master.

stef1927:
Thanks for the advice, in future I will push this kind of things directly to master.

PR #1196, opened 2016-08-09
Update jmx_auth_test on Windows for config file changes

jkni:
This pull request is intended to fix [CASSANDRA-11730](https://issues.apache.org/jira/browse/CASSANDRA-11730).

I originally PRed a fix to this issue as PR #1175 but the config file contents concurrently changed in [CASSANDRA-12109](https://issues.apache.org/jira/browse/CASSANDRA-12109), so the replace was no longer working correctly. This now passes correctly on trunk.

@mambocab to review probably, since he reviewed the original.


mambocab:
+1, based on visual inspection of this change and of 9eae8d340cb00fcabcab9b1c8c0d85943eac061b

PR #1198, opened 2016-08-10
handle missing attr

knifewine:
this is just to get upgrade tests running for the evening, then we should figure out if there's a better fix.


knifewine:
@mambocab looks like we had a small problem with 27be252c

PR #1204, opened 2016-08-10
add resource-intensive attr to flapping test

mambocab:
See CASSANDRA-12415.

This shouldn't close CASSANDRA-12415 until the new attribute successfully skips the test. @smccarthy788 and @ptnapoleon to review.

### DIFF
--- a/replace_address_test.py
+++ b/replace_address_test.py
@@ -6,6 +6,7 @@ import tempfile
 from cassandra import ConsistencyLevel, ReadTimeout, Unavailable
 from cassandra.query import SimpleStatement
 from ccmlib.node import Node, NodeError
+from nose.plugins.attrib import attr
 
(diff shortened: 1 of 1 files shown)


smccarthy788:
+1

ptnapoleon:
-1. If we think 4 nodes is too many, we should skip every 4 node test in this suite

mambocab:
Hmm, fair. Closing, then?

ptnapoleon:
I mean, you could just add them in this PR?

mambocab:
Oh, I thought you meant the opposite of that.

Do the other 4-node tests flap?

ptnapoleon:
Well, they haven't yet? But, since this test is parameterized, and there are other tests using the exact same set up, I guarantee they will flap in the same way?

PR #1208, opened 2016-08-10
Don't perform int division on sstable count

ptnapoleon:
anyone to review

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -143,7 +143,12 @@ class TestCompaction(Tester):
         if not hasattr(self, 'strategy') or self.strategy == "LeveledCompactionStrategy":
             size_factor = 1
         else:
-            size_factor = len(node1.get_sstables('keyspace1', 'standard1')) / len(node1.data_directories())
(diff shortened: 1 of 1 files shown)


mambocab:
This is weird -- there will never be a fractional number of dirs.

mambocab:
I'd prefer `from __future__ import division` here:

```python
>>> 11 / 2
5
>>> from __future__ import division
>>> 11 / 2
5.5
>>> 
```

If that's not possible because of other uses of `/` in the module, I'd prefer

```python
sstable_count = len(node1.get_sstables('keyspace1', 'standard1'))
dir_count = len(node1.data_directories())
debug("sstable_count is: {}".format(sstable_count))
debug("dir_count is: {}".format(dir_count))
size_factor = sstable_count / float(dir_count)
```

so it's clear that we're converting to `float` to get a `float` out of division.

Other than that, +1.

ptnapoleon:
Floats can represent whole numbers, but okay, made the second change

PR #1209, opened 2016-08-10
Add pull repair test for CASSANDRA-9876

geoffxy:
This adds a dtest for pull repairs, which are introduced in [CASSANDRA-9876](https://issues.apache.org/jira/browse/CASSANDRA-9876).

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -721,6 +721,42 @@ class TestRepair(BaseRepairTest):
 
         self._parameterized_range_repair(repair_opts=['-pr'])
 
+    @no_vnodes()
+    def pull_repair_test(self):
+        """
+        Test repair using the --pull option
+        @jira_ticket CASSANDRA-9876
+        * Launch a three node cluster
+        * Insert some data at RF 2
+        * Shut down node2, insert more data, restore node2
+        * Issue a pull repair on a range that only belongs to node1
+        * Verify that nodes 1 and 2, and only nodes 1+2, are repaired
+        * Verify that node1 only received data
+        * Verify that node2 only sent data
+        """
+        cluster = self.cluster
+        cluster.set_configuration_options(values={'hinted_handoff_enabled': False})
+        cluster.set_batch_commitlog(enabled=True)
+        debug("Starting cluster..")
+        cluster.populate(3).start(wait_for_binary_proto=True)
+
+        node1, node2, node3 = cluster.nodelist()
+
+        node1_address = node1.network_interfaces['binary'][0]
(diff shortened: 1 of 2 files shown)


repair_tests/repair_test.py
```diff
@@ -721,6 +721,41 @@ def partitioner_range_repair_test(self):
...
+    def pull_repair_test(self):
+        """
+        Test repair using the --pull option
+        * Launch a three node cluster
+        * Insert some data at RF 2
+        * Shut down node2, insert more data, restore node2
+        * Issue a pull repair on a range that only belongs to node1
+        * Verify that nodes 1 and 2, and only nodes 1+2, are repaired
+        * Verify that node1 only received data
+        * Verify that node2 only sent data
```
ptnapoleon: Can you add `@jira_ticket CASSANDRA-9876` on its own line in this docstring?

ptnapoleon:
+1 after nit

geoffxy:
I added in the JIRA ticket reference into the docstring and amended the commit. Thanks!

geoffxy:
@ptnapoleon Requesting another review - I made some changes to `deprecated_repair_test.py` to fix the test failures caused by changes in 9876.

The `self.assertEqual(opt["pull_repair"], "false", opt)` assertions are because the pull repair option won't be supported through the deprecated API. So all repairs carried out through the deprecated API should have that option be set to its default value (`false`).

ptnapoleon:
Lgtm!

jkni:
-1 from me - the further updates for fix the failures caused by CASSANDRA-9876 seem to break the tests on 2.2/3.0. We will need to version gate those changes to only apply for 3.10 and above. Can you do this @geoffxy? If not, let me know. Thanks!

geoffxy:
Absolutely - I'll update the PR to fix the changes. My bad for missing that!

ptnapoleon:
@jkni, do you want to double check this?

pauloricardomg:
submitted test runs for [2.2](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-22copy-dtest/1/) and [trunk](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunkcopy-dtest/26/), let's see how that goes.

jkni:
I'll check through these too - thanks for the 2.2/trunk runs.

pauloricardomg:
CI looks good.

jkni:
Looks good to me - +1 to merging

ptnapoleon:
Merging then, to get rid of the dtest failures

PR #1210, opened 2016-08-10
Filter new alias error message with < 3.8

ptnapoleon:
10783 was committed in 3.8, so we should NOT expect 3.8
 to have the old error

@knifewine to review


knifewine:
+1

PR #1211, opened 2016-08-10
Write a blocker function to ensure that an index is built

c-kodman:
Not sure if I got every instance where index_is_built could be used.

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -104,10 +104,15 @@ LOG = logging.getLogger('dtest')
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
-def block_index(index_name):
-    #takes index_name and checks if index is built
-    #returns boolean
-    #
+def index_is_built(keyspace, table_name, idx_name, session):
+    # checks if an index has been built
+    index_query = (
+        """SELECT * FROM system_schema.indexes WHERE keyspace_name = '{}' AND table_name = '{}' AND index_name = '{}'""".format(keyspace, table_name, idx_name)
+        if self.cluster.version() > '3.0' else
+        """SELECT * FROM system."IndexInfo" WHERE table_name = '{}' AND index_name = '{}.{}'""".format(keyspace, table_name, idx_name)
+    )
+    return len(list(session.execute(index_query))) == 1
+
 
 def get_sha(repo_dir):
     try:
@@ -116,7 +121,7 @@ def get_sha(repo_dir):
         if os.environ.get('LOCAL_GIT_REPO') is not None:
             prefix = 'local:'
         return "{}{}".format(prefix, output)
-    except CalledProcessError, e:
+    except CalledProcessError as e:
         if re.search('Not a git repository', e.message) is not None:
             # we tried to get a sha, but repo_dir isn't a git repo. No big deal, must just be working from a non-git install.
             return None
--- a/dtest.py
+++ b/dtest.py
@@ -104,9 +104,9 @@ LOG = logging.getLogger('dtest')
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
-def index_is_built(cluster, session, keyspace, table_name, idx_name):
+def index_is_built(node, session, keyspace, table_name, idx_name):
     # checks if an index has been built
-    full_idx_name = idx_name if cluster.version() > '3.0' else '{}.{}'.format(table_name, idx_name)
+    full_idx_name = idx_name if node.version() > '3.0' else '{}.{}'.format(table_name, idx_name)
     index_query = """SELECT * FROM system."IndexInfo" WHERE table_name = '{}' AND index_name = '{}'""".format(keyspace, full_idx_name)
     return len(list(session.execute(index_query))) == 1
 
--- a/dtest.py
+++ b/dtest.py
@@ -104,7 +104,7 @@ LOG = logging.getLogger('dtest')
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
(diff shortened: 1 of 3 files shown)


dtest.py
```diff
@@ -103,14 +103,24 @@
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
+def index_is_built(keyspace, table_name, idx_name, session):
+    # checks if an index has been built
+    index_query = (
+        """SELECT * FROM system_schema.indexes WHERE keyspace_name = '{}' AND table_name = '{}' AND index_name = '{}'""".format(keyspace, table_name, idx_name)
+        if self.cluster.version() > '3.0' else
```
ptnapoleon: I'm not sure this can work, since you made this a function and not a method. @mambocab would it be better to make this a method, or just take a cluster?

secondary_indexes_test.py
```diff
@@ -314,7 +310,7 @@ def index_is_built():
             index_sstables_dirs.append(index_sstables_dir)
 
         node1.nodetool("rebuild_index keyspace1 standard1 ix_c0")
-        while not index_is_built():
+        while not index_is_built('keyspace1', 'standard1', 'ix_c0', 'session'):
```
ptnapoleon: You should be passing `session`, not a string

dtest.py
```diff
@@ -103,14 +103,24 @@
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
+def index_is_built(keyspace, table_name, idx_name, session):
```
ptnapoleon: Logically it makes more sense (to me and this is a nit), to take session first

tools.py
```diff
@@ -11,14 +11,14 @@
...
-import assertions
 from cassandra import ConsistencyLevel
 from cassandra.concurrent import execute_concurrent_with_args
 from cassandra.query import SimpleStatement
 from ccmlib.node import Node
 from nose.plugins.attrib import attr
-from nose.tools import assert_equal, assert_in, assert_true, assert_is_instance
+from nose.tools import assert_equal, assert_in, assert_is_instance, assert_true
 
+import assertions
```
ptnapoleon: I think this change might be wrong

c-kodman: I only used pyclean and isort, I did not manually make these changes

mambocab: It should be correct -- it's in a section with other local imports, separated from 3rd party imports.

dtest.py
mambocab: That follows the idiom we use elsewhere. +1, the first arg should be a session here.

c-kodman:
http://cassci.datastax.com/job/c-kodman-trunk-dtest/31/

dtest.py
mambocab: It won't work, yes.

Of the two options presented, I'm not sure I have a preference. There's nothing keeping us from implementing it as a function (for testability, ease of use outside test methods, etc.) and implementing a method on top of that. It also could belong in CCM.

We also could keep the current API (module argument order as mentioned above) and use the session to check for the existence of `system_schema.indexes`. If it's there, use the first query, if not, use the second. That should be reliable enough, right?

dtest.py
```diff
@@ -103,14 +103,24 @@
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
+def index_is_built(cluster, session, keyspace, table_name, idx_name):
+    # checks if an index has been built
+    index_query = (
+        """SELECT * FROM system_schema.indexes WHERE keyspace_name = '{}' AND table_name = '{}' AND index_name = '{}'""".format(keyspace, table_name, idx_name)
+        if cluster.version() > '3.0' else
```
mambocab: How will this handle mixed-version clusters?

tools.py
ptnapoleon: tools.py has a recursive import problem, and I think it was wrong intentionally? I'll look into it and comment the file as such

dtest.py
ptnapoleon: Hmm, we should only take a node then, since the system tables are node specific anyway. We could even take a version string?

c-kodman:
http://cassci.datastax.com/job/c-kodman-trunk-dtest/32/

dtest.py
mambocab: Maybe it's a `Tester` method that takes a `ccmlib.node.Node`?

c-kodman:
http://cassci.datastax.com/job/c-kodman-trunk-dtest/

tools.py
c-kodman: fwiw, I changed it back.

ptnapoleon:
You have only one failure, but it worries me because its in a test you changed. Does this test fail otherwise?

c-kodman:
Yeah, this failure has to be due to me.  Looking into it.

c-kodman:
https://cassci.datastax.com/job/c-kodman-trunk-dtest/34/

c-kodman:
build was successful.  Ready for review

dtest.py
```diff
@@ -104,14 +104,21 @@
 logging.getLogger('cassandra').setLevel(logging.INFO)
 
 
+def index_is_built(cluster, session, keyspace, table_name, idx_name):
```
ptnapoleon: This still doesn't handle a mixed version cluster, and thus should take a node, not a cluster

ptnapoleon:
Can you run a new CI build, since these changes aren't trivial?

c-kodman:
https://cassci.datastax.com/job/c-kodman-trunk-dtest/35/

c-kodman:
Build passed.  Ready to merge.

PR #1213, opened 2016-08-10
Stabilize resumable_bootstrap_test by using byteman

yukim:
Although It's faily stable recently, [the most recent failure](http://cassci.datastax.com/job/cassandra-3.9_novnode_dtest/30/testReport/bootstrap_test/TestBootstrap/resumable_bootstrap_test/) indicates there still be a chance that the test fails because bootstrap streaming is not interrupted as expected.

This patch uses byteman instead of Thread and watch_for_log for killing node to fail bootstrap stream.
This is more reliable and since we can eliminate node start/stop and decrease the size of SSTable, test runs much faster (2x to 3x).

Byteman support is added to 2.2 also so this should be fine.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -12,7 +12,7 @@ from ccmlib.node import NodeError
 
 from assertions import assert_almost_equal, assert_one, assert_not_running
 from dtest import Tester, debug, DISABLE_VNODES
-from tools import (InterruptBootstrap, KillOnBootstrap, known_failure,
+from tools import (KillOnBootstrap, known_failure,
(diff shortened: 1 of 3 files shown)


ptnapoleon:
I'm +1 if you are, @mambocab

mambocab:
Looks good, but I'd like to see a multiplexer run first. Started one here:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/240

ptnapoleon:
That multiplex passed. Let's merge, @mambocab ?

mambocab:
SGTM. Thanks, Yuki!

PR #1214, opened 2016-08-11
Update tests after changing behaviour of LWTLs on non-existing partit…

ifesdjeen:
…ions vs empty statics

for [CASSANDRA-12060](https://issues.apache.org/jira/browse/CASSANDRA-12060)

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -800,13 +800,13 @@ class LWTTester(CQLTester):
             CREATE TABLE lwt_with_static (a int, b int, s int static, d text, PRIMARY KEY (a, b))
         """)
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I see the ticket has patches for 2.1+. Then we can rely on every major C* version to have the same behavior in these tests?

ifesdjeen:
Yes, that was the only reason to make a change to 2.1: to keep things consistent across the versions (even though the behaviour was consistent, just different prior to that change). 

I'll wait for review and merge that when C* ticket is reviewed.

ptnapoleon:
https://issues.apache.org/jira/browse/CASSANDRA-12724

ifesdjeen:
Looks like this particular commit is already merged in.

PR #1080, opened 2016-07-06
Sstable popen

mambocab:
This should enable debugging for [CASSANDRA-12137](https://issues.apache.org/jira/browse/CASSANDRA-12137). Depends on https://github.com/pcmanus/ccm/pull/510/files for proper error handling, so don't merge yet.

PR #1221, opened 2016-08-11
print stdout and err from sstableloader

mambocab:
Running the whole file here:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/238/

Serves the same purpose as #1080, but hopefully actually works.


mambocab:
Trying again here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/251/

ptnapoleon:
Lgtm, let's see what CI does

mambocab:
Oh snaaaap. Merging.

PR #1222, opened 2016-08-11
Use cqlsh instead of stress to verify data count

ptnapoleon:
So this fixes the test, as in, makes it pass without removing coverage of C*. However, the stress behavior is still a thing. Do we want to file a bug for that? @mambocab / @jkni?

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -6,6 +6,7 @@ from unittest import skip
 from cassandra import ConsistencyLevel
 from cassandra.query import SimpleStatement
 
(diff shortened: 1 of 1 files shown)


jkni:
I'm investigating the stress behavior right now - let's hold on merging this until I can make a call as to whether it is a bug.

mambocab:
This seems like a better assertion anyway, as far as readability goes. I'm +1 whether there's a stress bug or not.

ptnapoleon:
Joel says offline that stress is fine, we just didn't whitelist properly. Merging

PR #1225, opened 2016-08-12
Added more debug information for CASSANDRA-12430

stef1927:
As discussed on the ticket, we need more debug information to understand why the test failed.


ptnapoleon:
+1

PR #1228, opened 2016-08-12
Wait for all pools on cluster connect

jkni:
This was a tweak suggested by @aholmberg that restores an old behavior of the Python driver. Previously, the driver waited for all pools on cluster connect. Now, the default behavior is to return after successfully connecting to one. Note that this change would break tests for users who have not updated their driver.

@aholmberg, what are the consequences if we don't enable this? Tests should continue to function correctly, right?

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -503,7 +503,7 @@ class Tester(TestCase):
         cluster = PyCluster([node_ip], auth_provider=auth_provider, compression=compression,
                             protocol_version=protocol_version, load_balancing_policy=load_balancing_policy, default_retry_policy=FlakyRetryPolicy(),
                             port=port, ssl_options=ssl_opts, connect_timeout=10)
-        session = cluster.connect()
+        session = cluster.connect(wait_for_all_pools=True)
 
         # temporarily increase client-side timeout to 1m to determine
         # if the cluster is simply responding slowly to requests


aholmberg:
Almost all test should function correctly, and probably be marginally faster if they connect many sessions.

However, any test that assumed every pool was available for load balancing would be impacted. For example, if a sequence of queries executed immediately, and expected them to load balance across all available nodes (and not just the ones connected at that time). Or if a test inspected pool state immediately after connect.
By way of example: [here](https://github.com/datastax/python-driver/commit/d9ef7fc18ead9b2177ba112f097a3dc6c47fcf33#diff-7f7f70e3cbea6a2eceaa0e42ad451342R808) we wait for all pools because the test wants to verify quiescent pool state following the connection.

jkni:
I agree that we should favor robustness at this point. I'm not aware of any tests that rely on quiescent pool state at this moment. @mambocab or @ptnapoleon might have more complete knowledge there.

ptnapoleon:
I'm also drastically in favor of robustness and thus support the change.

mambocab:
I don't know of any tests that depend on some particular pool state.

I'm curious if there are failures we've seen that this addresses, or if this is just a proactive change in response to changing default driver behavior.

jkni:
Proactive. It might be something worth exploring more once we're really confident in the stability to the test suite.

jkni:
Oh, also I should note that I've done a run on trunk with this PR and saw no regressions.

knifewine:
I'm +1 just on the basis of a trunk trial run with no regressions.

mambocab:
> Proactive.

Do you think you could amend the commit to note this?

ptnapoleon:
I don't know that this is proactive? I just updated to the newest driver and _need_ this change to run tests properly.

aholmberg:
@ptnapoleon are you running with latest released or the `cassandra-test` branch? I would not expect this change to be required for `cassandra-test`, which includes some updates beyond 3.6.0.

ptnapoleon:
Obviously the latest released, since that's the wrong thing to be doing ;)

aholmberg:
Sorry, we had some issues with 3.6 + dtests. We originally rolled `cassandra-test` back to 3.5. Now advancing it with issues resolved, in order to incorporate the `USE_BETA_FLAG` and error map decoding for https://issues.apache.org/jira/browse/CASSANDRA-12311.

This is totally not confusing. Try to keep up. ;-)

ptnapoleon:
What is holding us back from merging this?

jkni:
I don't think anything - let's go ahead and merge.

PR #1229, opened 2016-08-12
Don't execute schema change in mixed-version cluster

thobbs:
Done for CASSANDRA-12192


thobbs:
We should be able to merge this after review without waiting for the cassandra ticket to be resolved.  There are some additional improvements that we may make to C*, but this test fix can be independent of that.

ptnapoleon:
+1, sorry to miss a pretty obvious test issue, but good to see some C* improvements are coming out of it

PR #1232, opened 2016-08-15
add known_failure annotation

c-kodman:


ptnapoleon:
+1 to the batch changes, but dont merge in the annotation for bug_5732 test yet

PR #1237, opened 2016-08-15
Wait for thrift port to open before connecting

ptnapoleon:


jkni:
Makes sense to me - we start the Native Server before the Thrift Server, so that could definitely race.

thrift_tests.py
```diff
@@ -77,6 +77,7 @@ def post_initialize_cluster(cls):
             node1.set_configuration_options(values={'initial_token': "a".encode('hex')})
 
         cluster.start(wait_for_binary_proto=True)
+        cluster.nodelist()[0].watch_log_for("Listening for thrift clients")  # Wait for the thrift port to open
         time.sleep(0.1)
```
mambocab: Will we still need this sleep?

jkni: Don't think so - it was probably an attempt to work around this issue.

ptnapoleon: I imagine so? I have documentation saying that the log message "Listening for X clients" actually occurs slightly before the action. Now, I suppose I could go look at the source, but the `0.1s` isn't hurting anyone?

ptnapoleon:
Merging with the "lgtm" from joel, and an offline "+1" from Hatch.

knifewine:
belated +1

ptnapoleon:
Turns out Russ didn't give me a +1, and I just can't read offline messages properly. Alas.

PR #1238, opened 2016-08-15
skip replace_address_test tests on small instances

mambocab:
This should have all the tests that I tagged as `resource-intensive` in #1212 run only on large instances. My mistake for thinking that actually skipping those should be blocked on other work.

Addresses https://issues.apache.org/jira/browse/CASSANDRA-12415. @ptnapoleon could you please review? Basically, all the tests I tagged as `resource-intensive` should be mentioned in the test-select list, and each of those that is `@since`d should be in there twice, once for `replace_address_test.py`, once for `tools.py`.


ptnapoleon:
+1

mambocab:
Cheers. Merging.

PR #1246, opened 2016-08-16
Mark pending range as resource intensive

ptnapoleon:
@smccarthy788 to review


smccarthy788:
Correctly imports `@attr` and add it to test. +1

PR #1249, opened 2016-08-17
Tests for CASSANDRA-12403

stef1927:
Some new tests for the slow query log, introduced by [CASSANDRA-12403](https://issues.apache.org/jira/browse/CASSANDRA-12403).

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -10,7 +10,7 @@ from cassandra.policies import FallthroughRetryPolicy
 from cassandra.protocol import ProtocolException
 from cassandra.query import SimpleStatement
 
-from assertions import assert_invalid, assert_one, assert_unavailable, assert_none, assert_all
+from assertions import assert_invalid, assert_one, assert_unavailable, assert_none, assert_all, assert_length_equal
 from dtest import Tester, canReuseCluster, freshCluster
 from thrift_bindings.v22.ttypes import \
     ConsistencyLevel as ThriftConsistencyLevel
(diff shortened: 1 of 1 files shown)


cql_tests.py
```diff
@@ -785,6 +785,192 @@ def materialized_view_test(self):
...
+
+        debug(session.execute(SimpleStatement("SELECT * from test1 where id = 1",
+                                              consistency_level=ConsistencyLevel.ONE,
+                                              retry_policy=FallthroughRetryPolicy())))
+
+        debug(session.execute(SimpleStatement("SELECT * from test1 where id = 1",
+                                              consistency_level=ConsistencyLevel.ONE,
+                                              retry_policy=FallthroughRetryPolicy())))
+
+        debug(session.execute(SimpleStatement("SELECT * from test1 where token(id) > 0",
```
ptnapoleon: Why are we wrapping these all in debug? What do we expect to see that wouldn't be printed by the python driver's logger?

cql_tests.py
```diff
@@ -785,6 +785,192 @@ def materialized_view_test(self):
...
+                                              retry_policy=FallthroughRetryPolicy())))
+
+        time.sleep(1)  # do our best to ensure logs had a chance to appear
+
+        self._check_logs(node, "operations were slow", 'system.log', 0)
+        self._check_logs(node, "SELECT \* FROM ks.test2", 'debug.log', 0)
+
+    def _check_logs(self, node, pattern, filename, num_expected):
+        ret = node.grep_log(pattern, filename=filename)
+        self.assertEquals(num_expected, len(ret),
```
ptnapoleon: You could just use `assert_length_equal` from `assertions.py` here

cql_tests.py
stef1927: I removed the debug wrappers. How do we enable the driver's logger?

cql_tests.py
stef1927: Replaced, thanks.

cql_tests.py
ptnapoleon:
```
import logging
# set python-driver log level to INFO by default for dtest
```

is in `dtest.py`. You could just set that to `logging.DEBUG` I believe, for the duration of this test

ptnapoleon:
Left the driver logging reply in-line, after that +1, thanks Stefania!

stef1927:
Thanks Philip, I'm merging the PR as is without changing the driver log level. We don't need the driver to log at DEBUG level for this test, I merely wanted to know for my own benefit. The intent of the debug statements was to show the query results but that's not strictly necessary for this test.

PR #1253, opened 2016-08-17
Only alter system_auth on 2.2+

ptnapoleon:
@mambocab to review


mambocab:
This code might be a little more self-explanatory if it just checked that that keyspace existed before running the `ALTER`. That would also let these tests successfully run with auth enabled, if we imagine a future where we run dtests in a big feature matrix like that.

But, that's a readability nit and a nit about preparing for a future that might not happen. I'm +1 on the code as-is if you don't want to change it.

PR #1255, opened 2016-08-17
node.get_versions in upgrade should return a LooseVersion

ptnapoleon:
I need to run a lot of CI to be sure this doesnt break everything

@mambocab to review

### DIFF
--- a/upgrade_tests/upgrade_base.py
+++ b/upgrade_tests/upgrade_base.py
@@ -191,7 +191,7 @@ class UpgradeTester(Tester):
         return node1.version()
 
     def get_node_versions(self):
-        return [n.get_cassandra_version() for n in self.cluster.nodelist()]
+        return [LooseVersion(n.get_cassandra_version()) for n in self.cluster.nodelist()]
 
     def node_version_above(self, version):
         return min(self.get_node_versions()) >= version
@@ -201,7 +201,7 @@ class UpgradeTester(Tester):
         Used in places where is_upgraded was used to determine if the node version was >=2.2.
         """
         node_versions = self.get_node_versions()
-        self.assertLessEqual(len(set(node_versions)), 2)
+        self.assertLessEqual(len({v.vstring for v in node_versions}), 2)
         return max(node_versions) if is_upgraded else min(node_versions)
 
     def tearDown(self):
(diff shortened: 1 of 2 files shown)


ptnapoleon:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-dtest/1/
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-upgrade/1/

upgrade_tests/upgrade_base.py
```diff
@@ -201,7 +201,7 @@ def get_node_version(self, is_upgraded):
         Used in places where is_upgraded was used to determine if the node version was >=2.2.
         """
         node_versions = self.get_node_versions()
-        self.assertLessEqual(len(set(node_versions)), 2)
+        self.assertLessEqual(len(node_versions), 2)
```
mambocab: This will always be equivalent to `assertLessEqual(len(self.cluster.nodelist()), 2)`. I believe `assertLessEqual(len({v.vstring for v in node_versions}), 2)` should do it.

ptnapoleon: What is the need to differentiate from `self.cluster.nodelist()`?

mambocab: If there are 3 or more nodes, the test calling this will just fail. The goal here is to make sure that, if this is a mixed-version cluster, there are at most 2 unique versions. The rewritten assertion checks that there are at most 2 nodes.

ptnapoleon:
@mambocab ready for re-review. The newest CI builds from those jobs show no bad results of this change (though not all dtests ran)

mambocab:
I'm confident enough to +1. Have you run it against something older than trunk? That'd make me even more confident -- even a small local run against one of the formerly-buggy tests would be great.

knifewine:
I'd love to see this run in some test jobs before it's committed.

ptnapoleon:
Also running on 2.2 as well then: http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-2.2-upgrade-upgrade/

ptnapoleon:
So that was all passing. Am I good to merge? ^

knifewine:
+1

PR #1257, opened 2016-08-19
Fix infinite loop in 2i tests

ptnapoleon:
@mambocab et al to review

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -106,7 +106,7 @@ logging.getLogger('cassandra').setLevel(logging.INFO)
 
 def index_is_built(node, session, keyspace, table_name, idx_name):
     # checks if an index has been built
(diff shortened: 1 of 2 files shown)


mambocab:
I'd prefer the time-checking part be executed as an assertion in the loop. If I read this correctly, after 30s, the loop will exit, whether the index is built or not. The test should error out in that case.

ptnapoleon:
That is a great idea. I'd rather merge this as-is, and then fix all instances of this loop logic to the new logic in a second PR, that has CI run on it. I admit this is bad practice, but I really want this fix in ASAP

knifewine:
while/else might be a way to go as well

ptnapoleon:
Oh clever, thanks Russ

PR #1258, opened 2016-08-19
Add test for CASSANDRA-12279

pauloricardomg:
this needs to wait for CASSANDRA-12279 to be merged

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -9,6 +9,8 @@ from cassandra.query import SimpleStatement
 from dtest import FlakyRetryPolicy, Tester, debug
 from tools.data import insert_c1c2, query_c1c2
 from tools.decorators import known_failure, no_vnodes, since
+from threading import Thread
+from ccmlib.node import ToolError
 
 
 def _repair_options(version, ks='', cf=None, sequential=True):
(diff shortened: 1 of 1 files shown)


ptnapoleon:
@knifewine can you review, please

knifewine:
The basic flow of the test looks reasonable to me, except one small issue. Since this test has the potential to hang test runs (but shouldn't once patched), I think it's preferable to run the attempted repair in the background and check to make sure it exits in a reasonable amount of time. The reason being if this somehow regressed in the future it would be difficult to track down the hanging test later, so we should be careful to prevent that.

I think this could be accomplished using 'nodetool_process' rather than the blocking 'nodetool' method. nodetool_process will return a subprocess object that can be checked a few times to see that it exits as expected, and if it doesn't return in a timely manner the test should fail itself (and probably make an attempt to terminate the subprocess too).

pauloricardomg:
I agree we should add protection against hanging. I created a new thread to trigger the blocking nodetool command and expect it to complete within 60s, please let me know what do you think of this approach. I submitted some CI runs:

* [2.2](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-2.2-12279-dtest/4/)
* [3.0](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-3.0-12279-dtest/5/)
* [trunk](http://cassci.datastax.com/view/Dev/view/paulomotta/job/pauloricardomg-trunk-12279-dtest/5/)

I didn't add a @since annotation with 2.2, because the test is still valid for 2.1 to make sure it will not hang, even though the `Unknown keyspace/cf pair` message is not logged.

knifewine:
+1 if CI is happy

thanks!

pauloricardomg:
CI looks good and original ticket was committed. Thanks for the review!

PR #1267, opened 2016-08-22
restrict C-12403 test to 3.10+

ptnapoleon:
Anyone to review


knifewine:
+1 if 3.10 is the right version

stef1927:
Thanks guys, I totally forgot the `@since`, 3.10 is definitely the right version.

PR #1268, opened 2016-08-22
use get_sstable_data_files method in ccm

smccarthy788:
@ptnapoleon, please review.

This should only be merged after https://github.com/pcmanus/ccm/pull/540


ptnapoleon:
+1

PR #1269, opened 2016-08-23
CASSANDRA-12092: fixTestAccuracy.test_simple_strategy_counters

stef1927:
This should hopefully fix all flakiness in this test, and possibly in other consistency tests as well, more details in [CASSANDRA-12092](https://issues.apache.org/jira/browse/CASSANDRA-12092).

### DIFF
--- a/consistency_test.py
+++ b/consistency_test.py
@@ -77,7 +77,7 @@ class TestHelper(Tester):
         nodes = self.nodes
         rf = self.rf
 
-        cluster.set_configuration_options(values={'hinted_handoff_enabled': False})
+        cluster.set_configuration_options(values={'hinted_handoff_enabled': False, 'dynamic_snitch': False})
         cluster.populate(nodes)
 
         if isinstance(nodes, int):
@@ -125,7 +125,9 @@ class TestHelper(Tester):
                 firstname text,
                 lastname text,
                 age int
-            ) WITH COMPACT STORAGE""")
+            ) WITH COMPACT STORAGE
+              AND dclocal_read_repair_chance = 0 AND read_repair_chance = 0
+              AND speculative_retry =  'NONE'""")
 
     def insert_user(self, session, userid, age, consistency, serial_consistency=None):
         text = "INSERT INTO users (userid, firstname, lastname, age) VALUES ({}, 'first{}', 'last{}', {}) {}"\
@@ -158,7 +160,8 @@ class TestHelper(Tester):
             CREATE TABLE counters (
                 id int PRIMARY KEY,
                 c counter
-            )
+            ) WITH dclocal_read_repair_chance = 0 AND read_repair_chance = 0
+              AND speculative_retry =  'NONE'
         """)
 
     def update_counter(self, session, id, consistency, serial_consistency=None):
@@ -169,23 +172,13 @@ class TestHelper(Tester):
 
     def query_counter(self, session, id, val, consistency, check_ret=True):
         statement = SimpleStatement("SELECT * from counters WHERE id = {}".format(id), consistency_level=consistency)
-        res = session.execute(statement)
-        ret = rows_to_list(res)
+        ret = rows_to_list(session.execute(statement))
         if check_ret:
             self.assertEqual(ret[0][1], val, "Got {} from {}, expected {} at {}".format(ret[0][1],
                                                                                         session.cluster.contact_points,
                                                                                         val,
                                                                                         consistency_value_to_name(consistency)))
         return ret[0][1] if ret else 0
-
-    def read_counter(self, session, id, consistency):
-        """
-        Return the current counter value. If we find no value we return zero
-        because after the next update the counter will become one.
-        """
-        statement = SimpleStatement("SELECT c from counters WHERE id = {}".format(id), consistency_level=consistency)
-        ret = rows_to_list(session.execute(statement))
-        return ret[0][0] if ret else 0
 
 
 class TestAvailability(TestHelper):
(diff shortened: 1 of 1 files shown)


knifewine:
Is there any potential for these dtest changes to mask a legitimate bug?

If bug masking isn't likely, I'm +1

knifewine:
The full suite ran fine locally for me as well.

stef1927:
It depends what the aim of the test should be. It is true that if we disable the dynamic snitch, read repair and speculative retry then we are not testing these features, so we need to ask ourselves whether we have other tests for them.

At the moment `TestAccuracy` is testing consistency combinations that do not ensure strong consistency, and in these cases we check that the test reads the up-to-date value at least from the number of replicas corresponding to the write CL but, in order to test this, we need to change all those settings to ensure that the test always reads from the replica it contacted and, as we've seen, this is not easy. 

Maybe we should abandon this, and accept that we can only test the guarantees offered by strong consistency and that we cannot easily poke into the Cassandra internals with a distributed test. Alternatively, we could use these settings only when required, by using different tables when we don't have strong consistency, or at a minimum for other testers.

knifewine:
@stef1927 Could it be reasonable to either a) add a second class that uses the more aggressive settings and put only the problematic tests there, or b) add a couple new helper methods to TestAccuracy for setting up slightly differently for just those tests ?

That way we could keep the general case testing basically the same and set things up better for the test(s) that really need differing settings to prevent flakiness. I think that would at least help us know that we couldn't mask problems as easily, and we'd benefit from testing a bit more configuration space.

stef1927:
Yes it's possible to use those settings only when required, I've implemented this in 9c78f09. We could even go further and split out only the problematic combinations but that would increase the test time since for each test method we need to restart the cluster. 

I'm multiplexing the counters test against 2.1 [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/34/) as well as running the entire consistency tests against trunk [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/33/).

stef1927:
The first build was aborted during the scp, restarted [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/35/).

stef1927:
The multiplexed build succeeded but the two builds running the full tests in `consistency_test.py` aborted twice in `TestAvailability:test_network_topology_strategy`. I've restarted a build without resource intensive tests [here](https://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/37/). I'm not sure if it's just as simple as picking a large instance on EC2 for resource intensive tests, I normally run my multiplexed tests on Openstack.

stef1927:
The latest build without the resource intensive tests was successful. Note that all links above are now page not found because I ran several multiplexed jobs yesterday for two other tickets.

Could I have a review of the latest changes please?

knifewine:
If the tests are all running well with the changes, I'm +1.

consistency_test.py
```diff
@@ -55,6 +55,29 @@ def _required_nodes(self, cl, rf_factors, dc):
...
 
+    def _get_num_nodes(self, idx, rf_factors, write_cl, read_cl):
+            """
+            Given a node index, identify to which data center we are connecting and return
+            number of nodes we write to, read from and whether R + W > N
+            """
+            nodes = [self.nodes] if isinstance(self.nodes, int) else self.nodes
+
+            dc = 0
+            for i in xrange(1, len(nodes)):
```
mambocab: Could you add a comment explaining how this loop works? It looks like it picks a datacenter, but I'm not sure how or for what.

consistency_test.py
```diff
@@ -55,6 +55,29 @@ def _required_nodes(self, cl, rf_factors, dc):
             ConsistencyLevel.LOCAL_ONE: 1,
         }[cl]
 
+    def _get_num_nodes(self, idx, rf_factors, write_cl, read_cl):
```
mambocab: This method's name could better reflect its purpose. Maybe something about consistency? I'm not sure, it's hard for me to summarize what it does.

consistency_test.py
```diff
@@ -55,6 +55,29 @@ def _required_nodes(self, cl, rf_factors, dc):
...
+
+            if write_cl == ConsistencyLevel.EACH_QUORUM:
+                write_nodes = sum([self._required_nodes(write_cl, rf_factors, i) for i in range(0, len(nodes))])
+            else:
+                write_nodes = self._required_nodes(write_cl, rf_factors, dc)
+
+            read_nodes = self._required_nodes(read_cl, rf_factors, dc)
+            strong_consistency = read_nodes + write_nodes > sum(rf_factors)
+
+            return write_nodes, read_nodes, strong_consistency
```
mambocab: Could you add, at the top level of this module, a `namedtuple` with fields named `write_nodes`, `read_nodes`, and `strong_consistency`, and use it for the return value of this method?

mambocab:
Oh, feel free to ignore my comments on `_get_num_nodes`, I just noticed that you're moving existing code.

mambocab:
Yeah, I'm also +1 if you're happy with CI results. Thanks!

stef1927:
Thanks for the review. Indeed `_get_num_nodes()` was just moved but why not make it better, so I've implemented @mambocab suggestions in f8a3149.

I was happy with the CI results for the flaky test and for those not labeled as resource intensive, the other ones I couldn't run. Either way, after the new commit we need to re-run them so I've started a build [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/57), I've launched it on EC2 2xlarge so hopefully the resource intensive tests will run as well.

stef1927:
Latest CI run went well, all tests passed. Waiting for a +1 from @mambocab for the implementation of his suggestions (f8a3149) and then I think we can merge.

mambocab:
Looks great, thank you! Merging.

PR #1270, opened 2016-08-23
patch for CASSANDRA-12524

smccarthy788:
@ptnapoleon & @mambocab, please review.

I've not been able to run this locally, so I'm running a multiplex here:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/271/

### DIFF
--- a/upgrade_tests/paging_test.py
+++ b/upgrade_tests/paging_test.py
@@ -6,9 +6,11 @@ from unittest import SkipTest, skipUnless
 from cassandra import ConsistencyLevel as CL
 from cassandra import InvalidRequest
 from cassandra.query import SimpleStatement, dict_factory, named_tuple_factory
+from ccmlib.node import TimeoutError
 from nose.tools import assert_not_in
 
 from dtest import RUN_STATIC_UPGRADE_MATRIX, debug, run_scenarios
+from distutils.version import LooseVersion
 from tools.assertions import assert_read_timeout_or_failure
 from tools.data import rows_to_list
 from tools.datahelp import create_rows, flatten_into_set, parse_data_into_dicts
@@ -1555,12 +1557,23 @@ class TestPagingWithDeletions(BasePagingTester, PageAssertionMixin):
             stmt = SimpleStatement("select * from paging_test", fetch_size=1000, consistency_level=CL.ALL)
             assert_read_timeout_or_failure(cursor, stmt)
 
-            patterns = [r"Scanned over.* tombstones during query 'SELECT \* FROM ks.paging_test.* query aborted",  # new pattern
-                        "Scanned over.* tombstones in ks.paging_test.* query aborted"]  # old pattern
+            self._grep_for_aborted_query_patterns(nodes, 50)
 
-            failed = any([n.grep_log(m) for n, m in itertools.product(nodes, patterns)])
+    def _grep_for_aborted_query_patterns(self, nodes, timeout):
+        old_pattern = r"Scanned over.* tombstones in ks.paging_test.* query aborted"
+        new_pattern = r"Scanned over.* tombstones during query 'SELECT \* FROM ks.paging_test.* query aborted"
 
-            self.assertTrue(failed, "Cannot find tombstone failure threshold error in log for {} node".format(("upgraded" if is_upgraded else "old")))
+        end_time = time.time() + timeout
+        while True:
+            if time.time() > end_time:
+                raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
+                                   " Unable to find: " + old_pattern + "or " + new_pattern + "in any node log within " + str(50) + "s")
+            for node in nodes:
+                pattern = old_pattern if LooseVersion(node.get_cassandra_version()) < LooseVersion('3.0') else new_pattern
+
+                if node.grep_log(pattern, "system.log"):
+                    return True
+            time.sleep(1)
 
 
 topology_specs = [
--- a/dtest.py
+++ b/dtest.py
@@ -763,7 +763,7 @@ class Tester(TestCase):
             time.sleep(1)
 
         raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
-                           " Unable to find: " + pattern + " in any node log within " + str(timeout) + "s")
+                           " Unable to find: " + repr(pattern) + " in any node log within " + str(timeout) + "s")
 
     def get_jfr_jvm_args(self):
         """


upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1726,17 @@ def test_failure_threshold_deletions(self):
...
+            # "Scanned over.* tombstones in ks.paging_test.* query aborted"]  # old pattern
 
-            failed = any([n.grep_log(m) for n, m in itertools.product(nodes, patterns)])
+            # Combine these two patterns into a single pattern to pass to wait_for_any_log()
 
-            self.assertTrue(failed, "Cannot find tombstone failure threshold error in log for {} node".format(("upgraded" if is_upgraded else "old")))
+            pattern = r"^Scanned over.* tombstones (during query 'SELECT \* FROM|tombstones in) ks\.paging_test.* query aborted$"
+
+            try:
+                failed = self.wait_for_any_log(nodes, pattern, 50)
```
ptnapoleon: Why do you need to do this? The TimeoutError being raised should fail the test

smccarthy788: It would, but I thought this would be more clear. We're explicitly saying this is a real failure case of this test, not just an environmental issue.

upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1726,17 @@ def test_failure_threshold_deletions(self):
             stmt = SimpleStatement("select * from paging_test", fetch_size=1000, consistency_level=CL.ALL)
             assert_read_timeout_or_failure(cursor, stmt)
 
-            patterns = [r"Scanned over.* tombstones during query 'SELECT \* FROM ks.paging_test.* query aborted",  # new pattern
-                        "Scanned over.* tombstones in ks.paging_test.* query aborted"]  # old pattern
+            # [r"Scanned over.* tombstones during query 'SELECT \* FROM ks.paging_test.* query aborted",  # new pattern
+            # "Scanned over.* tombstones in ks.paging_test.* query aborted"]  # old pattern
 
-            failed = any([n.grep_log(m) for n, m in itertools.product(nodes, patterns)])
+            # Combine these two patterns into a single pattern to pass to wait_for_any_log()
```
mambocab: No need for the whitespace above and below this.

Is this a situations where, on a mixed-version cluster, we could expect either log message? If we can figure out which message to expect based on C* version, we probably should, and check for the one we actually think will be there.

smccarthy788:
@ptnapoleon, bump. Multiplex was successful.

ptnapoleon:
Well, I'm +1, but like Jim said, it might be better to explicitly check for the correct pattern based on the node's version, rather than filtering for either? Franky, it's not a huge deal, so I'll leave it up to you.

smccarthy788:
updated to explicitly check the message based on node version.

Multiplexing a 2.x -> 3.x upgrade here:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/277

3.x -> indev_3.x here:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/276

upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1727,22 @@ def test_failure_threshold_deletions(self):
             stmt = SimpleStatement("select * from paging_test", fetch_size=1000, consistency_level=CL.ALL)
             assert_read_timeout_or_failure(cursor, stmt)
 
-            patterns = [r"Scanned over.* tombstones during query 'SELECT \* FROM ks.paging_test.* query aborted",  # new pattern
-                        "Scanned over.* tombstones in ks.paging_test.* query aborted"]  # old pattern
+            self.__grep_nodes_for_patterns(nodes, 50)
 
-            failed = any([n.grep_log(m) for n, m in itertools.product(nodes, patterns)])
+    def __grep_nodes_for_patterns(self, nodes, timeout):
```
mambocab: This should start with a single underscore.

upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1727,22 @@ def test_failure_threshold_deletions(self):
...
+        for _ in range(timeout):
```
mambocab: I'd prefer having the timeout logic be more like

```python
end_time = time.time() + timeout
while True:
    if time.time() > end_time:
        raise TimeoutError(...)
    ...
```

Either that, or rename `timeout` to be `retries`.

upgrade_tests/paging_test.py
mambocab: Also, the name's a little generic. `_grep_for_aborted_query_patterns`, maybe?

upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1727,23 @@ def test_failure_threshold_deletions(self):
...
-
-            failed = any([n.grep_log(m) for n, m in itertools.product(nodes, patterns)])
+            self.__grep_nodes_for_patterns(nodes, 50)
+
+    def _grep_for_aborted_query_patterns(self, nodes, timeout):
+        end_time = time.time() + timeout
+        while True:
+            if time.time() > end_time:
+                raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
+                                   " Unable to find: " + repr(pattern) + " in any node log within " + str(50) + "s")
```
smccarthy788: One thing we haven't considered is that only the pattern for the node with the greatest version will be reported here.

Maybe just hard code the two patterns into the exception? @mambocab?

mambocab: Good catch. I'd pull out the 2 patterns, then use them in both the checking and error-raising logic.

Incidentally, the timeout is incorrectly chardcoded here.

upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1727,26 @@ def test_failure_threshold_deletions(self):
...
 
-            failed = any([n.grep_log(m) for n, m in itertools.product(nodes, patterns)])
+    def _grep_for_aborted_query_patterns(self, nodes, timeout):
+        old_pattern = r"Scanned over.* tombstones in ks.paging_test.* query aborted"
+        new_pattern = r"Scanned over.* tombstones during query 'SELECT \* FROM ks.paging_test.* query aborted"
 
-            self.assertTrue(failed, "Cannot find tombstone failure threshold error in log for {} node".format(("upgraded" if is_upgraded else "old")))
+        end_time = time.time() + timeout
+        while True:
+            if time.time() < end_time:
```
mambocab: The conditional here is backwards, should be `time.time() > end_time`, right?

upgrade_tests/paging_test.py
```diff
@@ -1725,12 +1727,26 @@ def test_failure_threshold_deletions(self):
...
+                raise TimeoutError(time.strftime("%d %b %Y %H:%M:%S", time.gmtime()) +
+                                   " Unable to find: " + old_pattern + "or " + new_pattern + "in any node log within " + str(50) + "s")
+            for node in nodes:
+                if LooseVersion(node.version()) < LooseVersion('3.0'):
+                    pattern = old_pattern
+                else:
+                    pattern = new_pattern
+
+                if node.grep_log(pattern, "system.log"):
```
mambocab: This would be more succinct if you inlined the calculation of `pattern`. You could even inline the for loop into an `any(<comprehension>)` if you wanted, but that's up to you.

smccarthy788:
Multiplexing a 2.x -> 3.x upgrade here:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/287

3.x -> indev_3.x here:
http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/288

smccarthy788:
@ptnapoleon & @mambocab, the multiplex run were successful.

ptnapoleon:
This will eventually be replaced by the new ccm code you're writing, yes?

smccarthy788:
yes

ptnapoleon:
I'm +1, then

ptnapoleon:
Make sure to close the ticket with me as reviewer and remove the annotation.

PR #1277, opened 2016-08-24
make meta_tests a module

smccarthy788:
@mambocab, for your review.

This will group the test files in the meta_tests directory under a single module. Their results will be grouped together by nose.


mambocab:
Great, thanks! +1.

PR #1286, opened 2016-08-30
Fix for CASSANDRA-12479

stef1927:
This fix increases the COPY TO page timeout for `test_bulk_round_trip_non_prepared_statements` in order to fix the operation timed out errors, see [CASSANDRA-12479](https://issues.apache.org/jira/browse/CASSANDRA-12479). It also shows the COPY command's output for all bulk round trip tests as I suspect other tests may be failing with timeouts, see for example [CASSANDRA-12559](https://issues.apache.org/jira/browse/CASSANDRA-12559). Finally, I've removed the known failures of some solved tickets.

If CASSANDRA-12559 is also failing for the same reason, we may want to increase the COPY TO page timeouts for all bulk round trip tests.

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -1282,9 +1282,6 @@ class CqlshCopyTest(Tester):
 
         os.unlink(err_file.name)
 
-    @known_failure(failure_source='test',
-                   jira_url='https://issues.apache.org/jira/browse/CASSANDRA-12382',
-                   flaky=True)
     def test_reading_with_multiple_files(self):
         """
         Test importing multiple CSV files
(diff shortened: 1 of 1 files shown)


cqlsh_tests/cqlsh_copy_tests.py
```diff
@@ -2418,7 +2412,9 @@ def run_copy_to(filename):
             if copy_to_options:
                 copy_to_cmd += ' WITH ' + ' AND '.join('{} = {}'.format(k, v) for k, v in copy_to_options.iteritems())
             debug(copy_to_cmd)
-            ret.append(self.node1.run_cqlsh(cmds=copy_to_cmd, cqlsh_options=['--debug']))
+            result = self.node1.run_cqlsh(cmds=copy_to_cmd, cqlsh_options=['--debug'])
+            debug(result[0])
```
knifewine: could we get a little bit of text here so we know what this means when its output?

knifewine:
+1 after nit, sounds like multiplex was already done and looked good.

stef1927:
Thanks for the review, nit fixed and multiplex was indeed already done, so merging.

PR #1287, opened 2016-08-30
check stderr has no errors

smccarthy788:
@ptnapoleon, for your review.

We first questioned the validity of the error message in line 405 during cstar-97. I am unable to find it in the cassandra source code. Now that `stress` should return the correct return code, this test will automatically fail if stress fails, but as an extra precaution, I've checked stderr is clean, and left in the original error message check.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -10,7 +10,7 @@ from cassandra import ConsistencyLevel
 from cassandra.concurrent import execute_concurrent_with_args
 from ccmlib.node import NodeError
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
I must be missing something, where do we fail if stress's rc != 0?

smccarthy788:
`node.stress()` returns `handle_external_tool_process()`

https://github.com/pcmanus/ccm/blob/master/ccmlib/node.py#L1972

ptnapoleon:
Oh wow, that is a behavior I didn't realize was occurring inside ccm. +1

PR #1289, opened 2016-08-31
explain where byteman files should reside

smccarthy788:
@ptnapoleon, all the byteman files I can find are already in the byteman directory, so I just updated CONTRIBUTING.md to state where future .btm files should reside.


knifewine:
+1

PR #1294, opened 2016-08-31
remove unnecessary init methods

knifewine:


knifewine:
I get the same test counts comparing master to this branch, so that's a good sign things are working nicely.

Trying a trunk run here to see how things look: http://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-trunk_dtest_init_cleanup/1/

ptnapoleon:
This looks really nice but I have to defer to @mambocab for review here

knifewine:
CI run looks unhappy, so I need to sort that out before we continue here.

dtest.py
```diff
@@ -302,7 +302,11 @@ class Tester(TestCase):
     def __init__(self, *argv, **kwargs):
         # if False, then scan the log of each node for errors after every test.
         self.allow_log_errors = False
-        self.cluster_options = kwargs.pop('cluster_options', None)
+
```
mambocab: I don't think I like the "__init__ takes a `cluster_options` kwarg` behavior in the first place -- could we just do away with it? At a skim I don't see anywhere it's used.

mambocab:
Those are some weeeird errors on CI. Lemme know if you want help digging.

dtest.py
knifewine: yeah, seems like a weird leftover thing. doing away with it might be the ticket.

knifewine:
kicked off another CI build with some fixes in branch: http://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-trunk_dtest_init_cleanup/2/

knifewine:
another fix commited, and another run going here:
https://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-trunk_dtest_init_cleanup/3/

knifewine:
CI appears happy now, last run zero test failures.

knifewine:
@mambocab with CI happy now want to have a last look over this?

read_failures_test.py
```diff
@@ -13,14 +13,12 @@ class TestReadFailures(Tester):
     Tests for read failures in the replicas, introduced as a part of
     @jira_ticket CASSANDRA-12311.
     """
+    ignore_log_patterns = (
+        "Scanned over [1-9][0-9]* tombstones"  # This is expected when testing read failures due to tombstones
+    )
```
mambocab: This needs a `,` after the string to denote that it's a tuple.

rebuild_test.py
```diff
@@ -143,9 +140,9 @@ def resumable_rebuild_test(self):
 
         Test rebuild operation is resumable
         """
-        self.ignore_log_patterns = self.ignore_log_patterns[:] + [r'Error while rebuilding node',
-                                                                  r'Streaming error occurred on session with peer 127.0.0.3',
-                                                                  r'Remote peer 127.0.0.3 failed stream session']
+        self.ignore_log_patterns = list(self.ignore_log_patterns)[:] + [r'Error while rebuilding node',
```
mambocab: Since we already construct a new list with the `list` call, the `[:]` is unnecessary.

sstable_generation_loading_test.py
```diff
@@ -21,13 +22,13 @@ class BaseSStableLoaderTest(Tester):
...
     compact = False
     jvm_args = None
+    allow_log_errors = True
+    cluster_options = ImmutableMapping({'start_rpc': True})
 
     def __init__(self, *argv, **kwargs):
         if self.jvm_args is None:
             self.jvm_args = []
-        kwargs['cluster_options'] = {'start_rpc': True}
         Tester.__init__(self, *argv, **kwargs)
```
mambocab: Does this do any meaningful work with `Tester.__init__` being removed? I'd rather get this PR in, but maybe in a followup PR or ticket, we could remove this method entirely and set `jvm_args` to `()` on the class.

tools/misc.py
```diff
@@ -67,3 +68,23 @@ def generate_ssl_stores(base_dir, passphrase='cassandra'):
...
+
+    def __getitem__(self, key):
+        return self._data[key]
+
+    def __iter__(self):
+        for k in self._data.keys():
+            yield k
+
+    def __len__(self):
+        return len(self._data)
```
mambocab: Could you add this `__repr__`?

```python
def __repr__(self):
    return '{cls}({data})'.format(cls=self.__class__.__name__, data=self._data)
```

or similar as you see fit.

tools/misc.py
```diff
@@ -67,3 +68,23 @@ def generate_ssl_stores(base_dir, passphrase='cassandra'):
...
+    Useful at class level to prevent mutability problems (such as a method altering the class level mutable)
+    """
+    def __init__(self, init_dict):
+        self._data = init_dict.copy()
+
+    def __getitem__(self, key):
+        return self._data[key]
+
+    def __iter__(self):
+        for k in self._data.keys():
```
mambocab: If I had to guess from https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping, I think that `return iter(self._data)` might be slightly better here, as it doesn't depend on a `keys` implementation. Though, of course, we can usually assume there will be one, since we (should?) only ever use builtin `dict`s.

mambocab:
Looks great, very exciting. My nit on `read_failures_test` is the only hard blocker IMO. Could you address that, then multiplex a run on that module? Then feel free to merge with other nits addressed or not as you see fit.

read_failures_test.py
knifewine: good catch, fixed.

rebuild_test.py
knifewine: ah nice, fixed.

sstable_generation_loading_test.py
knifewine: updated to use jvm_args as a tuple, removed init entirely.

tools/misc.py
knifewine: oh, good idea.

tools/misc.py
knifewine: updated.

knifewine:
Doing another full CI run here since I made a few changes: https://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-trunk_dtest_init_cleanup/5/

knifewine:
latest CI is happy (1 failure but it's a known_failure). rebased and I think we're good to merge.

knifewine:
@mambocab can I get a +1 if you're happy here? (rather than multiplex I just did a full suite run, but I can multiplex too if you think it's needed).

ptnapoleon:
I'm +1 on the change happening, and the CI run, but I'll let Jim give it a final onceover tomorrow

mambocab:
Looks great, I'm +1. Sorry I missed it before it fell out of mergability :/

knifewine:
thanks for reviews! merging.

PR #1296, opened 2016-08-31
generic log filtering context manager

knifewine:


knifewine:
@mambocab would love some comment when you've got a few!

knifewine:
trying a full trunk dtest run with this in place, over here: http://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-trunk_dtest_log_contextman/1/

tools/context.py
```diff
@@ -0,0 +1,65 @@
...
+Home for functionality that provides context managers, and anything related to
+making those context managers function.
+"""
+import logging
+import os
+from contextlib import contextmanager
+
+from six import print_
+
+ALLOW_NOISY_LOGGING = os.environ.get('ALLOW_NOISY_LOGGING', '').lower() in ('yes', 'true')
```
mambocab: I think this should be gotten from the env in `dtest` and imported here, just to keep var-grabbing consolidated.

tools/context.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@contextmanager
+def silencing_of(log_id, expected_strings=None):
```
mambocab: NANANANA NANANANA NANANANA NANANANA BIKE SHEEEED

I think it could be clearer that this affects logging -- `silence_logger`? `filter_logger`? `log_filter`, so its use looks like `with log_filter('cassandra.cluster', [...]):`?

dtest.py
```diff
@@ -536,8 +515,11 @@ def patient_cql_connection(self, node, keyspace=None,
         if is_win():
             timeout *= 2
 
-        logging.getLogger('cassandra.cluster').addFilter(expect_control_connection_failures)
-        try:
+        with silencing_of(
```
mambocab: Just for readability, I think this could be

```python
expected_log_lines = ('Control connection failed to connect, shutting down Cluster:', '[control connection] Error connecting to ')
with silencing_of('cassandra.cluster', expected_log_lines):
    ...
```

mambocab:
+1 after my nits, looks awesome

tools/context.py
knifewine: discussed offline and env var will go into tools/ for now, to prevent circular imports.

tools/context.py
knifewine: log_filter seems reasonable.

dtest.py
knifewine: cool, done.

knifewine:
CI looks good, and confirmed with a little bit of local smoke testing. Merging, thanks @mambocab for the context manager idea and help along the way!

PR #1297, opened 2016-09-01
New tests for CASSANDRA-12423

stef1927:
These are 2 new thrift tests that @thobbs and I wrote for CASSANDRA-12423 to test range tombstones with end of component (eoc) set to zero. Once the ticket is committed, the tests will pass on all versions, at the moment they only work on 2.x but fail in 3.0+. There is also a new storage engine upgrade test to ensure that we can upgrade an sstable with such tombstones.

Please don't merge until the ticket is committed.

### DIFF
--- a/thrift_tests.py
+++ b/thrift_tests.py
@@ -122,6 +122,26 @@ class ThriftTester(ReusableClusterTester):
             cls.client.system_add_keyspace(ks)
 
 
+def i64(n):
+    return _i64(n)
+
+
+def i32(n):
+    return _i32(n)
+
+
+def i16(n):
+    return _i16(n)
+
+
+def composite(item1, item2=None, eoc='\x00'):
+    packed = _i16(len(item1)) + item1 + eoc
+    if item2 is not None:
+        packed += _i16(len(item2)) + item2
+        packed += eoc
+    return packed
+
+
 def _i64(n):
     return struct.pack('>q', n)  # big endian = network order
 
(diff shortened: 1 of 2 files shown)


upgrade_tests/storage_engine_upgrade_test.py
```diff
@@ -380,6 +386,41 @@ def upgrade_with_range_and_collection_tombstones_test(self):
...
+        client.transport.close()
+
+        session.execute("INSERT INTO rt (id, c1, c2, v) VALUES (1, 'asd', '', 0) USING TIMESTAMP 1470761451368658")
+        session.execute("INSERT INTO rt (id, c1, c2, v) VALUES (1, 'asd', 'asd', 0) USING TIMESTAMP 1470761449416613")
+
+        session = self._do_upgrade()
+
+        ret = list(session.execute('SELECT * FROM rt'))
+        debug('Results: {}'.format(ret))
+        self.assertEquals(2, len(ret))
```
ptnapoleon: You can use `assert_length_equals` from `tools.assertions` here, instead of the above two lines

ptnapoleon:
+1 with nit

upgrade_tests/storage_engine_upgrade_test.py
stef1927: Fixed, thank you.

thrift_tests.py
```diff
@@ -2480,3 +2554,37 @@ def test_compact_storage_get(self):
...
+        """
+        Insert a range tombstone with EOC=0 for a compact storage table. Insert 2 rows that
+        are just outside the range and check that they are present.
+
+        @jira_ticket CASSANDRA-12423
+        """
+        node1 = self.cluster.nodelist()[0]
+        session = self.patient_cql_connection(node1)
+
+        session.execute("USE \"Keyspace1\"")
```
mambocab: Feel free to merge with or without this change, but you can avoid escaping the `"`s by delimiting this string with `'`s: `'USE "Keyspace1"'`

stef1927: Fixed, thank you.

stef1927:
The 3 tests introduced by this PR will fail for 3.9 but will pass for 2.x, 3.0 and trunk. I'm not sure if there is a way to version gate the tests to avoid 3.9 or if we should perhaps revert this P.R. until 3.9 is released and we no longer run the tests for 3.9, cc @jkni.

ptnapoleon:
I'll version gate the test, in a manner consistent with how I'm gating elsewhere. Thanks for pointing this out in advance!

PR #1300, opened 2016-09-02
change imports for new module structure

mambocab:
Should fix the import problem causing [this test](http://cassci.datastax.com/job/trunk_dtest_upgrade/43/testReport/nose.failure/Failure/runTest/) to fail.


knifewine:
+1
didn't realize we had both opened PR's for the same change but you got it first!

mambocab:
Whoops, my bad! Thanks, Russ, merging.

PR #1304, opened 2016-09-02
move away from custom wait for compaction to using ccm implementation

knifewine:


knifewine:
tested relevant tests locally against trunk and 2.1.

@mambocab can you review?

compaction_test.py
```diff
@@ -172,7 +174,9 @@ def sstable_deletion_test(self):
         for x in range(0, 100):
             session.execute('delete from cf where key = ' + str(x))
 
-        block_on_compaction_log(node1, ks='ks', table='cf')
+        node1.flush()
+        node1.compact()
```
mambocab: This changes behavior a bit -- the new implementation compacts everything, not just `ks.cf`. Is that intentional? Is preserving that behavior important?

mambocab:
Running locally, but code looks good -- feel free to merge after my comment is addressed.

compaction_test.py
knifewine: Yeah, was intentional, as far as I could see the tests behave fine while doing a full compaction, so didn't see the need to keep the complexity.

ccm's Node.compact() implementation doesn't currently allow setting specific ks/table, though we could update that or call directly to Node.nodetool with custom params.

IMO we're fine to just do a wholesale compaction for now but can change it up too.

mambocab: Gotcha -- yeah, sounds good to me. I don't imagine there's enough data in other tables to make compaction take noticably longer across everything, yeah?

> ccm's Node.compact() implementation doesn't currently allow setting specific ks/table

Side note: that's surprising. I'll file a ccm ticket.

knifewine: Actually, on second thought it's trivial to just use nodetool here, so I'll do that and preserve compacting only ks.cf

Tested and it works just fine.

knifewine:
thanks for the feedback, fixed it up to keep behavior the same as before. will squash and merge once travis is happy.

mambocab:
:+1:

PR #1307, opened 2016-09-12
known_failure decorator tracks annotations as a list of dicts.

smccarthy788:
Update `collect_known_failure.py` to reflect this.

@mambocab, please review.

Tests with two annotations will be represented by `collect_known_failure.py` as:

```bash
{"failure_annotations": [{"failure_source": "test", "notes": "", "known_flaky": true, "jira_url": 
"https://issues.apache.org/jira/browse/CASSANDRA-12444"}, {"failure_source": "cassandra", "notes": "", "known_flaky": true, "jira_url": "https://issues.apache.org/jira/browse/CASSANDRA-12457"}], "name": "TestUpgrade_current_3_x_To_indev_3_x.rolling_upgrade_test", "module": "upgrade_tests.upgrade_through_versions_test"} ... ok
```

Something we should think about is any place where we access the `known_failure` attribute thinking there will be the `failure_source` value. I'm not aware of anything like this in dtest, but maybe it is used in Jenkins?

### DIFF
--- a/tools/decorators.py
+++ b/tools/decorators.py
@@ -68,16 +68,14 @@ def known_failure(failure_source, jira_url, flaky=False, notes=''):
     'systemic'.
 
     To run all known failures, use the functionality provided by the nosetests
-    attrib plugin, using the known_failure and known_flaky attributes:
+    attrib plugin, using the known_failure attributes:
 
         # only run tests that are known to fail
         $ nosetests -a known_failure
         # only run tests that are not known to fail
         $ nosetests -a !known_failure
         # only run tests that fail because of cassandra bugs
-        $ nosetests -a known_failure=cassandra
-        # only run tests that fail because of cassandra bugs, but are not flaky
-        $ nosetests -a known_failure=cassandra -a !known_flaky
+        $ nosetests -A "'cassandra' in [d['failure_source'] for d in known_failure]"
 
     Known limitations: a given test may only be tagged once and still work as
     expected with the attrib plugin machinery; if you decorate a test with
@@ -90,11 +88,17 @@ def known_failure(failure_source, jira_url, flaky=False, notes=''):
         assert_in(failure_source, valid_failure_sources)
         assert_is_instance(flaky, bool)
 
-        tagged_func = attr(known_failure=failure_source,
-                           jira_url=jira_url)(f)
-        if flaky:
-            tagged_func = attr('known_flaky')(tagged_func)
+        try:
+            existing_failure_annotations = f.known_failure
+        except AttributeError:
+            existing_failure_annotations = []
 
-        tagged_func = attr(failure_notes=notes)(tagged_func)
+        new_annotation = [{'failure_source': failure_source, 'jira_url': jira_url, 'notes': notes, 'flaky': flaky}]
+
+        failure_annotations = existing_failure_annotations + new_annotation
+
+        tagged_func = attr(known_failure=failure_annotations)(f)
+
         return tagged_func
+
     return wrapper
--- a/bin/collect_known_failures.py
+++ b/bin/collect_known_failures.py
@@ -36,18 +36,12 @@ class PrintJiraURLPlugin(nose.plugins.Plugin):
             cls=test.test,
         )
 
-        jira_url = get_attr_for_current_method(attr_name='jira_url')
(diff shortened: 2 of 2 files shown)


tools/decorators.py
```diff
@@ -155,11 +155,20 @@ def wrapper(f):
...
+        except AttributeError:
+            existing_failure_annotations = []
+
+        new_annotation = [{'failure_source': failure_source, 'jira_url': jira_url, 'notes': notes, 'known_flaky': flaky}]
+
+        failure_annotations = existing_failure_annotations + new_annotation
+
+        tagged_func = attr(known_failure=failure_annotations)(f)
+
         if flaky:
```
mambocab: I think we might want to remove this as well. I'll flesh my thoughts out further in a top-level comment

mambocab:
> Tests with two annotations will be represented by collect_known_failure.py as:

I'm surprised this doesn't present as a json list.

mambocab:
> Something we should think about is any place where we access the known_failure attribute thinking there will be the failure_source value. I'm not aware of anything like this in dtest, but maybe it is used in Jenkins?

I'm not sure I follow -- afiak, Jenkins doesn't consume this information, except maybe with the `nosetests -a` option. That's something to look into.

Speaking of `nosetests -a`, this change makes the documentation for `known_failures` incorrect, and may affect other documentation. This should be addressed.

Finally, @ptnapoleon/@knifewine, do we want to use this as an opportunity to fix the failure annotation API and names? There we could maybe improve the annotation interface for people using `nosetests -a`, though I can't think of a particularly useful application here. Is "run all the flaky tests" an interesting command? We could also remove `failure_source` and change `known_flaky` just to `flaky`.

Maybe these changes are out of scope or should be done later. I just know we've talked about some of this and it might be good to get it all out of the way in one go.

mambocab:
> I'll dig around for other cases

Someone with ssh access to CassCI should be able to grep for it, I think.

> there's a documented case where we want to run `nosetests -a known_failure=test`

Yeah, that'd be broken here. One solution, which I don't like much, is to require clients to call something like

```sh
nosetests -A "'test' in [d['failure_source'] for d in known_failure]"
```

We could track failure sources both in the list of dicts and in a `failure_sources` attribute to make this a bit easier:

```sh
nosetests -A "'test' in failure_sources"
```

But, I don't know that we can get around using `-A` without doing something much, much grosser. (To be clear, we probably should not do something much, much grosser.) I don't know if I like doubling up that data to make it easier to query via `-a`.

Just out of curiosity, what's the use case where we run with `-a known_failure=test`? I remember talking with folks about removing the failure type annotation entirely, so I'm surprised someone's using it.

> there's an edge case where one annotation is flaky and the other isn't. Is the test flaky?

I guess it's not? If one of the errors makes it fail all the time, then it's a hard failure. So, for a given test, `flaky` should be `all(d['flaky'] for d in known_failure)`.

smccarthy788:
>  what's the use case where we run with `-a known_failure=test`
It's just mentioned in the `known_failure` docs.

mambocab:
> It's just mentioned in the known_failure docs.

Ah, cool -- I think that's just a usage example. Unless you know of somewhere it's

mambocab:
Ping -- have you found any actual uses of `-a known_failure=...` or similar?

smccarthy788:
I have found no uses of `-a known_failures=...` etc.

mambocab:
Niiice. I think we can break the whole `-a` API then. @ptnapoleon ?

I think that also means removing the `if flaky` block that adds the extra attribute, yeah?

ptnapoleon:
Go ahead and break the API

smccarthy788:
@mambocab, what do you think of these changes?

mambocab:
Code looks good! Ran it locally and I saw the line

```
{"failure_annotations": [{"failure_source": "test", "notes": "windows", "flaky": false, "jira_url": "https://issues.apache.org/jira/browse/CASSANDRA-11236"}, {"failure_source": "cassandra", "notes": "", "flaky": true,               "jira_url": "https://issues.apache.org/jira/browse/CASSANDRA-12337"}], "name": "TestScrubIndexes.test_standalone_scrub", "module": "scrub_test"} ... ok 
```

Which is what I had hoped for. Merge when you're ready.

PR #1308, opened 2016-09-12
remove require annotation from tools

knifewine:


ptnapoleon:
Lgtm, but could we see a CI run, to confirm all the skips are correct?

knifewine:
rather than CI I just ran nose locally and can see the skips working as expected.

example output here: https://gist.github.com/knifewine/772972a40e153a9cf0cad15d6fafb760

I didn't run every single upgrade test method through this process but they don't differ in how they are skipped.

knifewine:
looks to work fine going through run_dtests.py script too: https://gist.github.com/knifewine/c07a2a65a05c4cc05d67eb3792c02c27

ptnapoleon:
+1!

PR #1309, opened 2016-09-13
skeleton for additional supercolumn coverage

smccarthy788:
@ptnapoleon, @mambocab, @knifewine: I'm opening this PR to have a place to discuss implementing additional super column coverage.

In short, we’re adding coverage for:
- Reading super columns in all supported C* versions via thrift & CQL
- Upgrading clusters with pre-existing super columns through every version

I think the best way to do this is some kind of `upgrade_through_versions` test:
- Create supercolumns
- Upgrade the cluster through versions:
    - Verify & modify data at each version

Initially I tried to implement this in `upgrade_through_versions_test.py`, but I was unsure how to start with a 1.2 C* cluster.

This follows the conventions established in `upgrade_supercolumns_test.py`, but is blocked on [CASSANDRA-12616](https://issues.apache.org/jira/browse/CASSANDRA-12616)

### DIFF
--- a/upgrade_supercolumns_test.py
+++ b/upgrade_supercolumns_test.py
@@ -19,7 +19,6 @@ COMPARATORS = ['BytesType', 'AsciiType', 'UTF8Type', 'IntegerType', 'LongType', 
                'TimestampType']
 
 
-@since('2.0')
 class TestSCUpgrade(Tester):
     """
     Tests upgrade between 1.2->2.0 for super columns (since that's where we
@@ -47,7 +46,7 @@ class TestSCUpgrade(Tester):
         cluster = self.cluster
 
         # Forcing cluster version on purpose
-        cluster.set_install_dir(version="1.2.16")
+        cluster.set_install_dir(version="1.2.19")
         if "memtable_allocation_type" in cluster._config_options:
             cluster._config_options.__delitem__("memtable_allocation_type")
         cluster.populate(num_nodes).start()
@@ -126,7 +125,7 @@ class TestSCUpgrade(Tester):
         self.verify_with_thrift()
         self.verify_with_cql()
 
-        versions = ['git:2.1', 'git:2.2', 'git:3.0', 'git:3.9', 'git:trunk']
+        versions = ['git:cassandra-2.1', 'git:cassandra-2.2', 'git:cassandra-3.0', 'git:cassandra-3.9', 'git:cassandra-trunk']
 
         for version in versions:
             session.cluster.shutdown()
--- a/upgrade_supercolumns_test.py
+++ b/upgrade_supercolumns_test.py
@@ -15,7 +15,11 @@ from tools.decorators import known_failure, since
 from tools.flaky import RerunTestException, requires_rerun
 
 
-@since('2.0', max_version='2.1.x')
+COMPARATORS = ['BytesType', 'AsciiType', 'UTF8Type', 'IntegerType', 'LongType', 'DoubleType', 'BooleanType', 'DateType',
+               'TimestampType']
+
+
+@since('2.0')
 class TestSCUpgrade(Tester):
     """
     Tests upgrade between 1.2->2.0 for super columns (since that's where we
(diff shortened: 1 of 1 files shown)


upgrade_supercolumns_test.py
```diff
@@ -39,6 +43,114 @@ def __init__(self, *args, **kwargs):
 
         Tester.__init__(self, *args, **kwargs)
 
+    def prepare(self, num_nodes=2):
+        cluster = self.cluster
+
+        # Forcing cluster version on purpose
+        cluster.set_install_dir(version="1.2.16")
```
ptnapoleon: We can go ahead and use 1.2.19 here, when this test was first written, 16 was the newest patch release, but that's no longer the case

upgrade_supercolumns_test.py
```diff
@@ -39,6 +43,114 @@ def __init__(self, *args, **kwargs):
...
+                    ThriftConsistencyLevel.ONE)
+
+    def restart_nodes(self, nodelist):
+        for node in nodelist:
+            node.flush()
+            time.sleep(.5)
+            node.stop(wait_other_notice=True)
+            self.set_node_to_current_version(node)
+            node.start(wait_other_notice=True, wait_for_binary_proto=True)
+            time.sleep(.5)
```
ptnapoleon: These `.5` sleeps are either unnecessary, or not long enough

upgrade_supercolumns_test.py
```diff
@@ -39,6 +43,114 @@ def __init__(self, *args, **kwargs):
...
+        self.create_super_column_family(client, comparator_type='UTF8Type', subcomparator_type='UTF8Type',
+                                        key_validation_class='UTF8Type', default_validation_class='UTF8Type')
+        session.cluster.control_connection.wait_for_schema_agreement()
+
+        self.add_super_column_data(client)
+
+        self.verify_with_thrift()
+        self.verify_with_cql()
+
+        versions = ['git:2.1', 'git:2.2', 'git:3.0', 'git:3.9', 'git:trunk']
```
ptnapoleon: Why aren't these in the form `git:cassandra-2.1` and so on?

upgrade_supercolumns_test.py
```diff
@@ -39,6 +43,114 @@ def __init__(self, *args, **kwargs):
...
+
+        for version in versions:
+            session.cluster.shutdown()
+            client.transport.close()
+            self.upgrade_to_version(version)
+            time.sleep(.5)
+
+            self.restart_nodes([node1, node2])
```
ptnapoleon: This and the call to 134 are probably redundant, or worse, actively non-functional. I'm not seeing the `upgrade_to_version` in this PR though, are we just using the existing one in the test?

upgrade_supercolumns_test.py
```diff
@@ -15,7 +15,11 @@
 from tools.flaky import RerunTestException, requires_rerun
 
 
-@since('2.0', max_version='2.1.x')
+COMPARATORS = ['BytesType', 'AsciiType', 'UTF8Type', 'IntegerType', 'LongType', 'DoubleType', 'BooleanType', 'DateType',
+               'TimestampType']
+
+
+@since('2.0')
```
ptnapoleon: No need for this, we never run on 1.2 anyway

upgrade_supercolumns_test.py
ptnapoleon: Do we want to separately track the upgrades from 2.1 -> 3.0 and 2.1 -> 2.2 -> 3.0?

upgrade_supercolumns_test.py
```diff
@@ -39,6 +43,114 @@ def __init__(self, *args, **kwargs):
...
+        ksdef = KsDef()
+        ksdef.name = 'test'
+        ksdef.strategy_class = 'SimpleStrategy'
+        ksdef.strategy_options = {'replication_factor': '2'}
+        ksdef.durable_writes = True
+        ksdef.cf_defs = []
+
+        client.system_add_keyspace(ksdef)
+
+    def create_super_column_family(self, client, comparator_type, subcomparator_type, key_validation_class, default_validation_class):
```
ptnapoleon: We 100 million % should just do this once, ever, and then use the static data we saved off. 1.2 is never going to see changes ever again, so generating a bunch of supercolumn sstables, and then re-using those in an upgrade is gong to be faster/easier, without us ever needing to debug driver connections to 1.2

smccarthy788: I think I've only used it once?

upgrade_supercolumns_test.py
smccarthy788: 1) Because they're wrong.
2) Probably.

upgrade_supercolumns_test.py
smccarthy788: There is an `upgrade_to_version` method defined at the bottom of the file. As with the `sleep()`s, I've mostly just been following the conventions established in this test.

upgrade_supercolumns_test.py
ptnapoleon: No, I don't mean per test run. I mean, ever. If those sstables will literally never change, we ought to manually create them. It saves us a lot of work in the test

upgrade_supercolumns_test.py
ptnapoleon: Okay. The `upgrade_to_version` at the bottom of the file already does restarts, so you shouldn't need to do that twice. Try removing all the sleeps and see what happens? This test is very old, so those are from when the log methods were much more brittle

upgrade_supercolumns_test.py
mambocab: Do we want to use the manifest and filter or manipulate it for our purposes?

smccarthy788:
I'm going to close this PR while working on this.

PR #1311, opened 2016-09-13
add default pattern to node grepping

smccarthy788:
@ptnapoleon, @mambocab: please review.

Running a multiplex here: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/309/


knifewine:
I'm a bit out of the loop on the ccm changes, but did want to note that in a couple places the LogPatternToVersion code checks the truthiness of default_pattern, which for an empty string will be False. Just wanted to mention in case that's not intended, disregard otherwise.

smccarthy788:
I think when we are using the empty string as `default_pattern`, it is ok everywhere except when we try to get the `patterns` attribute. In that case we can't distinguish from the case where `default_pattern=None`. I think it's that's a pretty simple fix.

The multiplex was successful.

ptnapoleon:
I don't know what default_pattern here means, but +1

smccarthy788:
`default_pattern` is a base case for the pattern fetching. E.g. when the pattern didn't exist until 2.2, and was changed at 3.0, `default_pattern` is the empty string.

PR #1317, opened 2016-09-14
Increase the page timeout for CASSANDRA-12430

stef1927:
Another COPY TO test that requires a higher page timeout, see [CASSANDRA-12430](https://issues.apache.org/jira/browse/CASSANDRA-12430).

Maybe we should review all cqlsh COPY TO tests and ensure they have a higher page timeout. It seems that the default, 10 seconds per 1000 rows, is not enough on Jenkins and causes rare failures. As an alternative, we could also consider doubling the default value in production code, I don't think this would cause issues.

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -119,6 +119,41 @@ class CqlshCopyTest(Tester):
         self.session.execute('DROP KEYSPACE IF EXISTS ks')
         self.ks = 'ks'
         self.create_ks(self.session, self.ks, 1)
+        self.cqlshrc = self.create_default_cqlsh_properties()
+
+    def create_default_cqlsh_properties(self):
+        """
+        Create a cqlshrc file containing some common copy options that are useful for
+        all tests:
+
+            - fix the page size and page timeout for COPY TO to 1000 rows and 60 seconds respectively,
+            since the default, 10 seconds per 1000 rows, can cause rare timeouts on slow Jenkins machines.
+
+        Note that any COPY options specified in the COPY command will override the options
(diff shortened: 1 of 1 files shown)


mambocab:
Code looks good, +1.

> we could also consider doubling the default value in production code

A third option is to provide a `cqlshrc` to all `run_cqlsh` calls, which would be an easier audit than all `COPY TO` calls IMO. Do you have a preference between that and changing the default timeout? In principle I don't like the idea of changing product behavior to make a flaky test env work, but this isn't particularly crucial behavior.

stef1927:
I also would prefer not to change production code for the tests. Using a _.cqlshrc_ file is a good idea. It needs to be in the home .cassandra folder or passed in as an option. Since `--debug` is also a useful option, I've wrapped `run_cqlsh` for copy tests to always use a _.cqlshrc_ file and `--debug`, see 6f4f82a.

I've started a multiplex run [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/68/).

stef1927:
I've aborted the multiplex run above because it was taking way too long, and I've restarted a single run [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/69/). I think there is at least one test failing due to the extra debug output.

stef1927:
I fixed it by adding a debug parameter defaulting to true, see 067fdad.

Tests running [here](http://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/71/).

mambocab:
Awesome. CI looks great, code looks great I'm +1

stef1927:
Thanks! Squashed and merged.

PR #1320, opened 2016-09-15
remove use of output files

smccarthy788:
@ptnapoleon, @mambocab: Please review.

Multiplexing here: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/314/

### DIFF
--- a/repair_tests/incremental_repair_test.py
+++ b/repair_tests/incremental_repair_test.py
@@ -1,4 +1,3 @@
-import os
 import time
 from collections import Counter
 from re import findall
@@ -54,17 +53,8 @@ class TestIncRepair(Tester):
         else:
             node3.nodetool("repair -par -inc")
 
-        with open('sstables.txt', 'w') as f:
-            node1.run_sstablemetadata(output_file=f, keyspace='keyspace1')
-            node2.run_sstablemetadata(output_file=f, keyspace='keyspace1')
-            node3.run_sstablemetadata(output_file=f, keyspace='keyspace1')
-
-        with open("sstables.txt", 'r') as r:
-            output = r.read().replace('\n', '')
-
-        self.assertNotIn('Repaired at: 0', output)
-
-        os.remove('sstables.txt')
+        for out in (node.run_sstablemetadata(keyspace='keyspace1').stdout for node in cluster.nodelist()):
+            self.assertNotIn('Repaired at: 0', out)
 
     @known_failure(failure_source='test',
                    jira_url='https://issues.apache.org/jira/browse/CASSANDRA-11268',
@@ -177,14 +167,10 @@ class TestIncRepair(Tester):
         node2.run_sstablerepairedset(keyspace='keyspace1')
         node2.start(wait_for_binary_proto=True)
 
-        with open('initial.txt', 'w') as f:
-            node2.run_sstablemetadata(output_file=f, keyspace='keyspace1')
-            node1.run_sstablemetadata(output_file=f, keyspace='keyspace1')
+        initialOut1 = node1.run_sstablemetadata(keyspace='keyspace1').stdout
+        initialOut2 = node2.run_sstablemetadata(keyspace='keyspace1').stdout
 
-        with open('initial.txt', 'r') as r:
-            initialoutput = r.read()
-
-        matches = findall('(?<=Repaired at:).*', initialoutput)
+        matches = findall('(?<=Repaired at:).*', '\n'.join([initialOut1, initialOut2]))
         debug("Repair timestamps are: {}".format(matches))
 
         uniquematches = set(matches)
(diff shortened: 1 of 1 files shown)


repair_tests/incremental_repair_test.py
```diff
@@ -54,17 +53,11 @@ def sstable_marking_test(self):
...
+        out2 = node2.run_sstablemetadata(keyspace='keyspace1').stdout
+        out3 = node3.run_sstablemetadata(keyspace='keyspace1').stdout
 
-        with open("sstables.txt", 'r') as r:
-            output = r.read().replace('\n', '')
-
-        self.assertNotIn('Repaired at: 0', output)
-
-        os.remove('sstables.txt')
+        self.assertNotIn('Repaired at: 0', [out1, out2, out3])
```
ptnapoleon: I don't think this will work?

smccarthy788: I didn't expect it to either, but it passes locally.

ptnapoleon: Oh that's because it's a negative assertion. The semantics of `in` change when checking a list instead of a string. So, instead of doing substring search, we're now checking that none of `out1`, `out2`, or `out3` are literally the string `Repaired at: 0`, which they never will be, because they're making longer than that. It would be better instead to do something like
```
for out in (out1, out2, out3):
    self.assertNotIn('Repaired at: 0', out)
```

smccarthy788:
Running a new multiplex here : https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/315/

repair_tests/incremental_repair_test.py
```diff
@@ -224,10 +210,7 @@ def sstable_repairedset_test(self):
 
         self.assertGreaterEqual(max(matchcount), 2)
 
-        self.assertNotIn('Repaired at: 0', finaloutput)
-
-        os.remove('initial.txt')
-        os.remove('final.txt')
+        self.assertNotIn('Repaired at: 0', [finalOut1, finalOut2])
```
ptnapoleon: This needs to either check both of these vars, or the concatenation of them

smccarthy788: Good catch.

repair_tests/incremental_repair_test.py
```diff
@@ -177,14 +171,10 @@ def sstable_repairedset_test(self):
...
-            node2.run_sstablemetadata(output_file=f, keyspace='keyspace1')
-            node1.run_sstablemetadata(output_file=f, keyspace='keyspace1')
+        initialOut1 = node1.run_sstablemetadata(keyspace='keyspace1').stdout
+        initialOut2 = node2.run_sstablemetadata(keyspace='keyspace1').stdout
 
-        with open('initial.txt', 'r') as r:
-            initialoutput = r.read()
-
-        matches = findall('(?<=Repaired at:).*', initialoutput)
+        matches = findall('(?<=Repaired at:).*', initialOut1 + initialOut2)
```
mambocab: I'm not sure, but I think this should maybe be passed `initialOut1 + '\n' + initialOut2`, to make sure that if 1 ends with the searched-for string, and 2 starts with it, we don't accidentally treat them as one. Hopefully 1 is newline-terminated, but just in case.

mambocab: Same goes for all the other instances below. If there are more than 2, or if you think it looks cleaner, you can do this easily with `'\n'.join([...])`.

repair_tests/incremental_repair_test.py
```diff
@@ -384,15 +367,10 @@ def sstable_marking_test_not_intersecting_all_ranges(self):
...
-            output = r.read()
-
-        self.assertNotIn('Repaired at: 0', output)
+        out1 = node1.run_sstablemetadata(keyspace='keyspace1').stdout
+        out2 = node2.run_sstablemetadata(keyspace='keyspace1').stdout
+        out3 = node3.run_sstablemetadata(keyspace='keyspace1').stdout
+        out4 = node4.run_sstablemetadata(keyspace='keyspace1').stdout
 
-        os.remove('final.txt')
+        for out in (out1, out2, out3, out4):
```
mambocab: If this is the only these `out*` variables are used, you could do:

```python
for out in (node.run_sstablemetadata(keyspace='keyspace1').stdout for node in self.cluster.nodelist()):
    self.assertNotIn('Repaired at: 0', out)
```

Maybe pull that comprehension out to a list with a name. But IMO fewer intermediate variables is always good if they're only used in one place.

smccarthy788:
@mambocab, I have addressed your review comments. 

Multiplexing here : https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/318/

mambocab:
Code looks good, I'm +1.

mambocab:
(Assuming CI is good.)

PR #1318, opened 2016-09-14
Loose versions

ptnapoleon:
@mambocab to review. Depends on https://github.com/pcmanus/ccm/pull/548

CI at http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-dtest/20/testReport/
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgrade-upgrade/2/

PR #1322, opened 2016-09-15
Fix version comparison

ifesdjeen:
String comparison only works for versions 1 to 9. For 10 we have to use `LooseVersion` thanks @jkni for pointing it out!


ptnapoleon:
So #1318 will supercede this fix, and CI for that should be done in an hour? Would you mind if I just merged that, instead of this (and then needing to rebase 1318 again)?

ifesdjeen:
Sure, absolutely. Thanks!

PR #1327, opened 2016-09-16
Parameterize error message in snitch_test after typo fix in C*

jkni:
This typo was fixed in https://github.com/apache/cassandra/commit/2dc2bb808634be9aaf5ee1d65eb968f8c5514305, but the test was never updated accordingly.


ptnapoleon:
So, feel super free to say "No" and merge, but watch_log_for takes a regex, not just a string, so you could change 
```
node1.watch_log_for("Intiated reconnect to an Internal IP 
```
to
```
node1.watch_log_for("In.?tiated reconnect to an Internal IP 
```
instead of using an additional var

jkni:
That's a much better approach. I pushed an updated version.

ptnapoleon:
+1

PR #1244, opened 2016-08-16
CASSANDRA-12148 cdc hard-link support

josh-mckenzie:

PR #1332, opened 2016-09-20
CASSANDRA-12148 cdc hard-link support

mambocab:
Supercedes #1244 -- see that ticket for discussion.

At time of writing, still needs to be rebased.

### DIFF
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -515,12 +515,10 @@ class TestCDC(Tester):
         generation_node.stop()
         generation_session.cluster.shutdown()
 
-        source_cdc_indexes = []
         # We can rely on the existing _cdc.idx files to determine which .log files contain cdc data.
         source_path = os.path.join(generation_node.get_path(), 'cdc_raw')
-        for seg in os.listdir(source_path):
-            if '_cdc' in seg:
-                source_cdc_indexes.append(ReplayData(source_path, seg))
+        source_cdc_indexes = [ReplayData(source_path, name)
+                              for name in source_path if '_cdc' in name]
 
         # create a new node to use for cdc_raw cl segment replay
         loading_node = self._init_new_loading_node(ks_name, cdc_table_info.create_stmt)
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -47,11 +47,7 @@ def _insert_rows(session, table_name, insert_stmt, values):
 
 
 def _move_commitlog_segments(source_dir, dest_dir, verbose=True):
-    for source_filename in os.listdir(source_dir):
-        # skip any cdc index files and only move .log files
-        if '_cdc' in source_filename:
-            continue
-
+    for source_filename in [name for name in os.listdir(source_dir) if '_cdc' not in name]:
         source_path, dest_path = (os.path.join(source_dir, source_filename),
                                   os.path.join(dest_dir, source_filename))
         if verbose:
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -519,6 +519,7 @@ class TestCDC(Tester):
         source_path = os.path.join(generation_node.get_path(), 'cdc_raw')
         source_cdc_indexes = [ReplayData.load(source_path, name)
                               for name in source_path if name.endswith('_cdc.idx')]
+        self.assertGreater(len(source_cdc_indexes), 0)
 
         # create a new node to use for cdc_raw cl segment replay
         loading_node = self._init_new_loading_node(ks_name, cdc_table_info.create_stmt)
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -47,7 +47,7 @@ def _insert_rows(session, table_name, insert_stmt, values):
 
 
 def _move_commitlog_segments(source_dir, dest_dir, verbose=True):
-    for source_filename in [name for name in os.listdir(source_dir) if '_cdc' not in name]:
+    for source_filename in [name for name in os.listdir(source_dir) if not name.endswith('_cdc.idx')]:
         source_path, dest_path = (os.path.join(source_dir, source_filename),
                                   os.path.join(dest_dir, source_filename))
         if verbose:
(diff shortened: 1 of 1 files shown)


mambocab:
Rebased. Testing locally.

mambocab:
@josh-mckenzie I'm seeing failures locally, running on CI here:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/321/

mambocab:
Job failed with an authentication error. Trying again:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/322/

cdc_test.py
```diff
@@ -494,22 +498,27 @@ def test_cdc_data_available_in_cdc_raw(self):
...
         generation_node.drain()
         debug('stopping')
         # stop the node and clean up all sessions attached to it
         generation_node.stop()
         generation_session.cluster.shutdown()
 
+        # We can rely on the existing _cdc.idx files to determine which .log files contain cdc data.
+        source_path = os.path.join(generation_node.get_path(), 'cdc_raw')
+        source_cdc_indexes = [ReplayData.load(source_path, name)
+                              for name in source_path if '_cdc' in name]
```
josh-mckenzie: That'd be technically more granular, correct. I don't know the likelihood of _cdc ever showing up in a regular commitlog name, but no harm in it.

mambocab: Cool. I'm less concerned here about granularity and more about clarity -- `name.endswith('_cdc.idx')` looks a lot more like "`name` is a CDC index file" than `'_cdc' in name` to me.

mambocab:
Restarting here to account for new code on Josh's branch: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/324/

mambocab:
Post-rebase run here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/326/

mambocab:
Rebased `37e98c1...82e0745` to deal with changes in imports.

mambocab:
@josh-mckenzie My most recent commits add the assertions that we talked about, and passes locally. Running on CassCI here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/336/

mambocab:
Hmm, do we still expect that failure? Re-running here: http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/345/

josh-mckenzie:
Hm. "In first set but not in second" seems suspect, but I need to think about it a bit more. We might be seeing the reverse of the previous symptom, where hard-links are created for CLS that don't have CDC data in them which would then not propagate over to the 2nd node.

mambocab:
Hmm -- if that's the case, I can't think of any assertions we can reliably make about non-CDC CLs.

josh-mckenzie:
I don't know if there's a way for us to force a complete deterministic flush of *all* CLS to sstables without poking a hole in the DB, since even a drain/shutdown can race with the system/schema tables being mutated. Is this a 100% reproduction of failure or just intermittent on multiplexed run?

mambocab:
I see it consistently locally.

mambocab:
Rebased 03dd88f88d76caa473b83bb73e82f3f6c280ae8a onto `master` at 6c43337.

josh-mckenzie:
I've thought about this today and I have an idea. It's kind of terrible but it might work.

If our goal is to prove whether or not a CommitLog with CDC data inside it is correctly flagged and moved into the cdc folder, I believe a chain of 3 cassandra clusters should prove whether or not the replay logic is giving us the data we expect. We use the 3rd cluster as a ghetto "CDC reader" to confirm the CDC data we originally put in cluster 1 was replayed and linked in cluster 2, then read by and is available in cluster 3. So, a specific flow would be:

1. Cluster 1 generates the data, flushes, drains, shuts down
2. We create a 2nd cluster, start and create cdc-enabled schema matching cluster 1, then shut it down
3. We move the CDC files from cluster 1 into cluster 2's commitlog folder
4. We restart cluster 2 and allow replay to finish
5. We confirm the data is in the tables in cluster 2
6. We shut down cluster 2 (flush, drain, shut down)
7. We spin up a 3rd cluster, start and create cdc-enabled schema matching cluster 1, then shut it down
8. We move the CDC files from cluster 2 into cluster 3 **(These are the files that are created by replay on cluster 2)**
9. We start cluster 3, wait for replay to finish, then confirm the CDC data from cluster 1 is in cluster 3

That should tell us, definitively, whether or not the replay process is placing the data we're looking for into the CDC logs in cluster 2 after replay.

@mambocab: See any major flaws in the above reasoning? @ptnapoleon: this is janky enough, and has gone on long enough, that I wouldn't mind a 2nd sanity check from you as well. I think Jim and I both are thinking ourselves in circles on this one.

If neither of you see anything immediately flawed in the above testing plan, I'll see about creating a PR to your branch that does the above @mambocab. Shouldn't require major code-changes since it's really just repeating the steps you've already coded up w/perhaps some src/dest dir changes, and we'll need to see how long it takes.

ptnapoleon:
It sounds functional, if unreasonably slow.

josh-mckenzie:
Slow, yeah. Alternative would be to write our own out-of-process cdc tester (see [here](https://issues.apache.org/jira/browse/CASSANDRA-11575)) but that's not ready yet.

mambocab:
I've lost a lot of context for this -- I don't follow what the second transfer buys us that the first doesn't. What bug would make data consistently be transferred to cluster 2, but not to cluster 3? Do we expect replay to mark or organize commitlogs differently than normal writing will?

josh-mckenzie:
Cluster 1->2 tests if replay puts the data into memtables/sstables
Cluster 2->3 tests if replay *correctly links that data in the cdc folder*

Code-path for treatment of CommitLogs is unique to replay, yes. That clear it up?

mambocab:
Got it, yep 👍

PR #1338, opened 2016-09-21
deps for pcmanus/ccm#554

knifewine:
changes necessary to match pcmanus/ccm#554 (local: slug changes to support multiple local versions).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ futures
 six
 -e git+https://github.com/datastax/python-driver.git@cassandra-test#egg=cassandra-driver
 
-ccm==2.4.0
+ccm==2.4.3
 cql
 decorator
 docopt
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ futures
 six
 -e git+https://github.com/datastax/python-driver.git@cassandra-test#egg=cassandra-driver
 
-ccm==2.4.3
+ccm==2.4.2
 cql
 decorator
 docopt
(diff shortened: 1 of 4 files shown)


knifewine:
first testing iteration over here, emulating the trunk dtest upgrade job: https://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-dtest_upgrade_local_slugs/1/

knifewine:
another test job for confirming that dev autojobs work ok, since these are scripted a bit differently than the 'main' jobs: https://cassci.datastax.com/view/Dev/view/knifewine/job/knifewine-upgrade_test_309-upgrade/1/

ptnapoleon:
+1 (can merge after ccm merge?) We also need to change requirements right?

mambocab:
I'm also +1, and yeah -- I think this'll need to bump the ccm dep in requirements, and wait on publishing the new ccm version.

knifewine:
Yeah, I guess that would mean a ccm version bump. We'll want this and the ccm change to happen at the same time to keep CI from doing weird things.

knifewine:
Updated requirements.txt to use ccm 2.4.3 which I bumped up in the ccm pr. If you guys are happy here, we just need to wait for that new ccm version to be published and this can be merged I think.

ptnapoleon:
It is confusing, I know, but you want 2.4.2, not 2.4.3

requirements.txt
```diff
@@ -5,7 +5,7 @@ futures
 six
 -e git+https://github.com/datastax/python-driver.git@cassandra-test#egg=cassandra-driver
 
-ccm==2.4.0
```
ptnapoleon: Whatever happened to ccm 2.4.1?

knifewine:
corrected to 2.4.2

knifewine:
think we're good to go, here, mergin'

PR #1340, opened 2016-09-22
bounce cluster post-startup on cl-replaying tests

mambocab:
Should address [CASSANDRA-11811](https://issues.apache.org/jira/browse/CASSANDRA-11811). I'll run CI, if it runs clean I'll get Joel to review.

### DIFF
--- a/cdc_test.py
+++ b/cdc_test.py
@@ -253,6 +253,13 @@ class TestCDC(Tester):
         self.cluster.populate(1)
         self.cluster.set_configuration_options(dict(config_defaults, **configuration_overrides))
         self.cluster.start(wait_for_binary_proto=True)
+        # drain and bounce to get clean CLs. We do this to address
+        # CASSANDRA-11811, where mutations generated on startup get replayed
+        # when we replay commitlogs
+        self.cluster.drain()
+        self.cluster.stop()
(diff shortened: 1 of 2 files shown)


mambocab:
Welp, doesn't work:

http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/327/testReport/node_0_iter_001.cdc_test/TestCDC/test_cdc_data_available_in_cdc_raw_2/

@jkni Looks like I'm doing this wrong; do you have suggestions?

ptnapoleon:
I assume there is a step missing before the drain, where-in we wait for the cluster to "settle"?

jkni:
The problem was that it turns out these mutations happen on every start sequence, so bouncing the node didn't solve the problem. Instead, I tried a patch that simply ensures we advance past these commitlog segments [here](https://github.com/jkni/cassandra-dtest/tree/cdc-test-fix).

A parameterized dtest run showed no failures [here](http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/329/), so that may be a good approach. What do you think?

mambocab:
Looks like a good approach. I'll pull it out into utility functions and file a PR.

mambocab:
Closing in favor of #1349.

PR #1344, opened 2016-09-26
Use correct version check function in upgrade cql_tests

ptnapoleon:
@knifewine or @smccarthy788 to review. I ran these all locally across all relevant versions. The only failing test was failing 100% consistently, so no real need for CI.


smccarthy788:
+1

PR #1348, opened 2016-09-27
Update CQL tests to use proper DC name for CASSANDRA-12681

jeffjirsa:
CCM uses 'datacenter1'. The dtest uses 'dc1'. CASSANDRA-12681 validates the datacenter name. 

Whether or not CASSANDRA-12681 gets merged, we should use the real datacenter name.


ptnapoleon:
+1 Thanks

PR #1351, opened 2016-09-28
Removes outdated known_failure annotations

blerer:
Removing the `known_failure` annotations for `multi_collection_test` and `multi_list_set_test` as the problem behind those tests seems to have been fixed.


ptnapoleon:
+1

PR #1353, opened 2016-09-29
Fix versions in which CASSANDRA-12060 is applied

ifesdjeen:
CASSANDRA-12060 is only for 3.0.10 and 3.10 and up. 3.0.9 and 3.9 would still fail. This patch fixes it.


ptnapoleon:
+1

ifesdjeen:
Thanks!

PR #1349, opened 2016-09-27
Write to fresh CLs

mambocab:
Supercedes #1340, implementing @jkni's suggested approach with a shared utility function. @jkni can you review?

PR #1354, opened 2016-09-30
Restore system peers and local in commitlog archive tests to avoid tokens mismatch

stef1927:
Whilst reviewing [CASSANDRA-12509](https://issues.apache.org/jira/browse/CASSANDRA-12509), which drains a node on shutdown, I noticed that the failure described in [CASSANDRA-11811](https://issues.apache.org/jira/browse/CASSANDRA-11811) occurs every single time with the 12509 patch applied. I tried applying P.R. #1340 or bouncing the node even earlier, before any commitlog archiving was enabled, but to no avail. I think there is always one CL segment left even when draining.

The only fix I was able to find is in this P.R., which  restores `system.peers`, `system.local` and restarts with `-Dcassandra.load_ring_state=true` to avoid generating new tokens. The downside is that all the contents of these two tables are also restored, I'm not sure if this violates the test purpose. IMO, if commitlog archiving restores all segments for all tables then one cannot rely on it to only restore some tables (schema and user) and generate new data (like tokens) so I'm not sure how users typically use this functionality.

### DIFF
--- a/snapshot_test.py
+++ b/snapshot_test.py
@@ -219,7 +219,7 @@ class TestArchiveCommitlog(SnapshotTester):
                         [(r'^archive_command=.*$', 'archive_command={archive_command} %path {tmp_commitlog}/%name'.format(
                             tmp_commitlog=tmp_commitlog, archive_command=archive_command))])
 
-        cluster.start()
+        cluster.start(wait_for_binary_proto=True)
 
         session = self.patient_cql_connection(node1)
         create_ks(session, 'ks', 1)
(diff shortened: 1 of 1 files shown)


ptnapoleon:
@mambocab

mambocab:
[Running on CassCI here](https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/337/).

I have another PR in flight to try to deal with [CASSANDRA-11811](https://issues.apache.org/jira/browse/CASSANDRA-11811). It's #1349, and it supercedes #1340. I guess I didn't mention it on the JIRA ticket; sorry about that. I started a CassCI run today, so we'll see how it works there. What do you think of that approach? I genuinely don't know if either is what we want. Maybe Josh or some other expert on the commitlog will have something to say?

> I'm not sure if this violates the test purpose

Good point - I'm not sure either, so I'm looking into it now.

If we do merge this, this fix should be better-documented with comments, or functions with names to explain their purpose.

> I'm not sure how users typically use this functionality.

Me neither. When I was talking to Joel about #1349, he said that if a user wrote a nontrivial amount of data before a snapshot, new commitlogs would supercede any commitlogs with problematic mutations -- so, assuming that users don't snapshot commitlogs on empty or almost-empty clusters, they should be ok. I believe he plans to a look at this PR, so I'll let him weigh in when he gets to it.

stef1927:
Closing in favor of #1349.

PR #1357, opened 2016-10-05
dtest for CASSANDRA-12689

brstgt:
See here https://issues.apache.org/jira/browse/CASSANDRA-12689


brstgt:
Forgot to remove this, had problem with my local (MacOS) JAVA installation:

        expected_error = (r".*Class JavaLaunchHelper is implemented in both")
        self.ignore_log_patterns = [expected_error]

materialized_views_test.py
```diff
@@ -1733,3 +1735,63 @@ def _consistent_reads_after_write_test(self, num_partitions):
...
+            'concurrent_materialized_view_writes': 1,
+            'concurrent_writes': 1,
+        })
+        self.cluster.start(wait_for_binary_proto=True, jvm_args=[
+            "-Dcassandra.test.fail_mv_locks_count=64"
+        ])
+
+        self.nodes = self.cluster.nodes.values()
+
+        session = self.patient_exclusive_cql_connection(self.nodes[0], protocol_version=5)
```
ptnapoleon: Is 5 the correct protocol version here?

materialized_views_test.py
```diff
@@ -1733,3 +1735,63 @@ def _consistent_reads_after_write_test(self, num_partitions):
...
+
+        keyspace = "locktest"
+        session.execute("""
+            CREATE KEYSPACE IF NOT EXISTS %s
+            WITH replication = { 'class': 'SimpleStrategy', 'replication_factor': '1' }
+            """ % (keyspace))
```
ptnapoleon: Can you use the `'{}'.format()` style of string interpolation, as described in `CONTRIBUTING.md`?

materialized_views_test.py
```diff
@@ -1733,3 +1735,63 @@ def _consistent_reads_after_write_test(self, num_partitions):
...
+    @since('3.0')
+    def test_mutations_dontblock(self):
+        session = self._prepare_cluster()
+        records = 100
+        records2 = 100
+        for x in xrange(records):
+            for y in xrange(records2):
+                session.execute_async("INSERT INTO test (int1, int2, date) VALUES ({}, {}, toTimestamp(now()))".format(x, y))
+
+        assert_one(session, "SELECT count(*) FROM test WHERE int1 = 1", [records2])
```
ptnapoleon: The inserts are done asynchronously, so you'll want to block until all the futures have returned, before you check this

ptnapoleon:
Overall looks very nice, but I left a few nits

materialized_views_test.py
brstgt: Sorry for that simple question but is there a elegant way to wait for all futures like Promise.all() or do I have to loop over all futures and call future.result()?

mambocab: You can also execute them synchronously with [`execute`](https://datastax.github.io/python-driver/api/cassandra/cluster.html#cassandra.cluster.Session.execute). I doubt that'd take noticeably longer for such a small dataset.

mambocab: Oh! or this is a great case for using [`execute_concurrent_with_args`](https://datastax.github.io/python-driver/api/cassandra/concurrent.html#cassandra.concurrent.execute_concurrent_with_args). Could be something like

```python
execute_concurrent_with_args(
    session.prepare('INSERT INTO test (int1, int2, date) VALUES (?, ?, toTimestamp(now()))'),
    zip(xrange(records), xrange(records2))
)
```

With the `results_generator` given its default of `False`, this should block until all inserts have returned.

brstgt:
Is it ok like that?

ptnapoleon:
If you could add blank line at the end of the file, then +1. Thanks!

brstgt:
Hm check the last commit (405ea00). There is a blank line.

ptnapoleon:
Huh, the github UI must have been misleading me. +1

thobbs:
CASSANDRA-12689 has been committed, so I believe this should be ready to merge.  I'll let you guys take care of that just to be safe.

PR #1363, opened 2016-10-19
Make show_output=True the default for all cqlsh copy tests

stef1927:
In view of some recent failures in cqlsh copy tests that are very hard to reproduce, I propose to show the output of the cqlsh copy commands for all tests, so that in future we may be able to understand why a copy command has failed. 

I've already run the patch [here](https://cassci.datastax.com/view/Dev/view/stef1927/job/stef1927-dtest-multiplex/73/).


ptnapoleon:
+1

PR #1365, opened 2016-10-20
Change upgrade tests to expect different ordering after C-12420

ptnapoleon:
@thobbs or @guyboltonking to review


upgrade_tests/cql_tests.py
```diff
@@ -353,8 +350,8 @@ def limit_multiget_test(self):
             # Check that we do limit the output to 1 *and* that we respect query
             # order of keys (even though 48 is after 2)
```
guyboltonking: I know the change here is just to lower the bound on the version where this check takes place, but is it worth altering the comment to say that that query order _isn't_ respected for versions above 2.1.17?  Or maybe just delete the comment altogether?

ptnapoleon: Yep, good point

PR #1367, opened 2016-10-21
Fix TestCompaction.fanout_size_test

DikangGu:


compaction_test.py
```diff
@@ -518,6 +518,11 @@ def fanout_size_test(self):
         """
         @jira_ticket CASSANDRA-11550
         """
+        if not hasattr(self, 'strategy'):
+            self.strategy = 'LeveledCompactionStrategy'
```
ptnapoleon: This appears to override the set compaction strategy, and change it to LCS, in only a subset of cases. I think this might be confusing? It would be best to simply skip in all cases when we aren't using LCS, and never to change the strategy.

ptnapoleon:
Can you also remove the known failure annotation? After that, +1

PR #1368, opened 2016-10-24
Add TestNonRedundantRangeCalculation for CASSANDRA-12870

spodkowinski:
Complementary dtest for:
https://issues.apache.org/jira/browse/CASSANDRA-12870

### DIFF
--- a/pending_range_test.py
+++ b/pending_range_test.py
@@ -67,3 +67,65 @@ class TestPendingRangeMovements(Tester):
         # Check we can still execute LWT
         for i in xrange(1000):
             session.execute(lwt_query)
+
+
+class TestNonRedundantRangeCalculation(Tester):
+
+    def non_redundant_calculation_test(self):
+        """
+        @jira_ticket CASSANDRA-12281
+        """
+        cluster = self.cluster
+        # If we are on 2.1, we need to set the log level to debug or higher, as debug.log does not exist.
+        if cluster.version() < '2.2' and not TRACE:
+            cluster.set_log_level('DEBUG')
+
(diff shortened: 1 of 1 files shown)


pending_range_test.py
```diff
@@ -67,3 +67,65 @@ def pending_range_test(self):
...
+        for _, m in matches:
+            nodes = [s.strip() for s in m.group(1).split(',')]
+            debug("Nodes grouped for pending range calculation: {}".format(nodes))
+
+            kss1Found = kss1Found or kss1.issubset(nodes) and not (kss2.union(kss3.union(kss4))).intersection(nodes)
+            kss2Found = kss2Found or kss2.issubset(nodes) and not (kss1.union(kss3.union(kss4))).intersection(nodes)
+            kss3Found = kss3Found or kss3.issubset(nodes) and not (kss1.union(kss2.union(kss4))).intersection(nodes)
+            kss4Found = kss4Found or kss4.issubset(nodes) and not (kss1.union(kss2.union(kss3))).intersection(nodes)
+
+        self.assertTrue(kss1Found and kss2Found and kss3Found and kss4Found)
```
ptnapoleon: It might be better to break this up into four calls, otherwise when one of these is false, we won't know which.
```
for kssFound in [kss1Found, kss2Found, kss3Found, kss4Found]:
    self.assertTrue(kssFound)
```

spodkowinski: How would you know which boolean would cause the assertion to fail? You'd just get the line number of the assert.

ptnapoleon: Fair. Better to break it up into four sequential calls then.

PR #1370, opened 2016-10-24
Only run sstablemetadata against nodes with sstables

ptnapoleon:
@smccarthy788 to review


smccarthy788:
Yep, that's one way to do it. Do you think we should push this down into ccm, and do it for all calls of `run_sstablemetatdata()`?

+1

ptnapoleon:
Ehh, I think that ccm should throw a different error, but it should continue to throw an error. From the test's end, I'd rather just not run when we shouldn't, than to expect and catch an error.

PR #1379, opened 2016-11-09
Fix C-12874, read/write at CL QUORUM

ptnapoleon:
Anyone to review


ptnapoleon:
I didnt set the default at the prepare() call level, because some tests in this suite override it per test.

smccarthy788:
+1

PR #1381, opened 2016-11-11
Add dtest for CASSANDRA-12813

ifesdjeen:
[CASSANDRA-12813](https://issues.apache.org/jira/browse/CASSANDRA-12813)

### DIFF
--- a/upgrade_internal_auth_test.py
+++ b/upgrade_internal_auth_test.py
@@ -60,7 +60,7 @@ class TestAuthUpgrade(Tester):
         cluster.add(node3, False)
         node3.start()
 
-        time.sleep(15)
+        node3.watch_log_for('Created default superuser')
 
         node3.drain()
         node3.watch_log_for("DRAINED")
--- a/upgrade_internal_auth_test.py
+++ b/upgrade_internal_auth_test.py
@@ -3,6 +3,7 @@ from unittest import skipIf
 
 from cassandra import Unauthorized
 from ccmlib.common import is_win
+from ccmlib.node import Node
 
 from dtest import OFFHEAP_MEMTABLES, Tester, debug
 from tools.assertions import assert_all, assert_invalid
@@ -40,6 +41,55 @@ class TestAuthUpgrade(Tester):
     @skipIf(OFFHEAP_MEMTABLES, 'offheap_objects are not available in 3.0')
     def upgrade_to_30_test(self):
         self.do_upgrade_with_internal_auth("git:cassandra-3.0")
+
+    def test_upgrade_legacy_table(self):
+        """
+        Upgrade with bringing up the legacy tables after the newer nodes (without legacy tables)
+        were started.
+
+        @jira_ticket CASSANDRA-12813
+        """
+
+        cluster = self.cluster
+
+        # Forcing cluster version on purpose
+        cluster.set_install_dir(version="2.1.16")
+
+        node3 = Node('node3', cluster, False, ('127.0.0.3', 9160), ('127.0.0.3', 7000),
+                     '7400', '2000', None, binary_interface=('127.0.0.3', 9042))
+        cluster.add(node3, False)
+        node3.start()
+
+        time.sleep(15)
+
+        node3.drain()
+        node3.watch_log_for("DRAINED")
+        node3.stop(gently=True)
+
+        # Ignore errors before upgrade on Windows
+        if is_win():
+            node3.mark_log_for_errors()
+
+        self.set_node_to_current_version(node3)
+
+        # Populate the two nodes with newer version, they will get started without
+        # legacy tables. Legacy tables will get replicated later.
+        cluster.populate(2)
+
+        node1, node2, node3 = cluster.nodelist()
+        self.set_node_to_current_version(node1)
+        self.set_node_to_current_version(node2)
+
+        node1.start(wait_for_binary_proto=True)
+        node2.start(wait_for_binary_proto=True)
+        node3.start()
+
+        node1.watch_log_for('Initializing system_auth.credentials')
+        node1.watch_log_for('Initializing system_auth.permissions')
+        node1.watch_log_for('Initializing system_auth.users')
+
+        # Should succeed. Will throw an NPE on pre-12813 code.
+        self.patient_cql_connection(node1, user='cassandra', password='cassandra')
 
     def do_upgrade_with_internal_auth(self, target_version):
         """


upgrade_internal_auth_test.py
```diff
@@ -41,6 +42,54 @@ def upgrade_to_22_test(self):
     def upgrade_to_30_test(self):
         self.do_upgrade_with_internal_auth("git:cassandra-3.0")
 
+
+    def do_upgrade_legacy_table(self):
```
ptnapoleon: If this is intended to be a test method (which it appears to be), it needs to have `test` in the name, or nose wont run it.

upgrade_internal_auth_test.py
```diff
@@ -41,6 +42,54 @@ def upgrade_to_22_test(self):
...
+
+        # Forcing cluster version on purpose
+        cluster.set_install_dir(version="2.1.16")
+
+        node3 = Node('node3', cluster, False, ('127.0.0.3', 9160), ('127.0.0.3', 7000),
+                     '7400', '2000', None, binary_interface=('127.0.0.3', 9042))
+        cluster.add(node3, False)
+        node3.start()
+
+        time.sleep(15)
```
ptnapoleon: Why the sleep here?

upgrade_internal_auth_test.py
```diff
@@ -41,6 +42,54 @@ def upgrade_to_22_test(self):
...
+
+        self.set_node_to_current_version(node3)
+        cluster.populate(2)
+
+        node1, node2, node3 = cluster.nodelist()
+        self.set_node_to_current_version(node1)
+        self.set_node_to_current_version(node2)
+
+        node1.start(wait_for_binary_proto=True)
+        node2.start(wait_for_binary_proto=True, wait_other_notice=True)
```
ptnapoleon: No need to set wait_other_notice here, that's True by default.

upgrade_internal_auth_test.py
```diff
@@ -41,6 +42,54 @@ def upgrade_to_22_test(self):
...
+        node3.drain()
+        node3.watch_log_for("DRAINED")
+        node3.stop(gently=True)
+
+        # Ignore errors before upgrade on Windows
+        if is_win():
+            node3.mark_log_for_errors()
+
+        self.set_node_to_current_version(node3)
+        cluster.populate(2)
```
ptnapoleon: Is there a specific need to do this set of operations in the reverse of the expected way (creating node 3, then adding it to the cluster, then populating nodes 1 and 2)?

upgrade_internal_auth_test.py
```diff
@@ -41,6 +42,54 @@ def upgrade_to_22_test(self):
...
+
+        node1, node2, node3 = cluster.nodelist()
```
ptnapoleon: I was worried that it was wrong to assume that this `node3` matches the `node3` from above, because I thought `cluster.nodelist()` would return the nodes in the order they were created, but apparently it sorts them alphabetically, so this does work.

upgrade_internal_auth_test.py
```diff
@@ -145,7 +194,10 @@ def upgrade_to_version(self, tag, node):
 
         # Update Cassandra Directory
         debug('Updating version to tag {tag}'.format(**format_args))
-        node.set_install_dir(version=tag, verbose=True)
+        # if tag == None??
```
ptnapoleon: Why this comment?

upgrade_internal_auth_test.py
```diff
@@ -145,7 +194,10 @@ def upgrade_to_version(self, tag, node):
...
+        self.set_node_to_current_version(node)
+        # node.set_install_dir(version=tag, verbose=True)
+        # node.set_install_dir(version=tag, verbose=True)
```
ptnapoleon: This commented code should presumably be removed, no?

upgrade_internal_auth_test.py
```diff
@@ -145,7 +194,10 @@ def upgrade_to_version(self, tag, node):
 
         # Update Cassandra Directory
         debug('Updating version to tag {tag}'.format(**format_args))
-        node.set_install_dir(version=tag, verbose=True)
+        # if tag == None??
+        self.set_node_to_current_version(node)
```
ptnapoleon: Given you've changed this shared behavior, can you link CI for all affected tests?

ptnapoleon: Why change this though, when I dont see where you call this method in your test?

upgrade_internal_auth_test.py
ifesdjeen: Copied it from the same file, next test for following reasons:

```
# wait for default superuser creation
        # The log message
        # node.watch_log_for('Created default superuser')
        # will only appear on one of the three nodes, and we don't know
        # which ahead of time. Grepping all three in parallel is unpleasant.
        # See auth_test and auth_roles test for instances of this as well.
        # Should be fixed by C*-6177
        time.sleep(15)
```

upgrade_internal_auth_test.py
ifesdjeen: Yes, otherwise the two nodes will get started with legacy schema tables (they'll get replicated before statements will get prepared, which is exactly the bug we're hunting)

upgrade_internal_auth_test.py
ifesdjeen: Yes, I've triple-checked it. I had same exact concerns.

upgrade_internal_auth_test.py
ptnapoleon: I think we can watch for the log message here, since there is only one node.

ifesdjeen: Thanks! Switched to watch instead.

upgrade_internal_auth_test.py
```diff
@@ -41,6 +42,55 @@ def upgrade_to_22_test(self):
...
+        if is_win():
+            node3.mark_log_for_errors()
+
+        self.set_node_to_current_version(node3)
+
+        # Populate the two nodes with newer version, they will get started without
+        # legacy tables. Legacy tables will get replicated later.
+        cluster.populate(2)
+
+        node1, node2, node3 = cluster.nodelist()
```
mambocab: Depending on the "`nodelist()` orders nodes by name" behavior worries me, since (afaik) that's not a guarantee that `nodelist()` makes explicitly and could break in the future. I'd prefer something like

`node1, node2 = [n for n in cluster.nodelist() if n.address() != node3.address()]`.

@ptnapoleon can you confirm that this would correctly pull out the non-`node3` nodes? The result of `node.address()` should uniquely identify the nodes, right?

ifesdjeen: I only closed it to make sure it doesn't get merged before we clean up the rest of patch..

ptnapoleon:
I can confirm this test correctly fails prior to the first patch, and after patches 1 and 2.

upgrade_internal_auth_test.py
ptnapoleon: Yes, that would work, Jim

mambocab: > I only closed it to make sure it doesn't get merged before we clean up the rest of patch..

Oh, I understand, cool :+1:

ifesdjeen:
This is already good news.

ifesdjeen:
Just to make sure: fails with NPE, correct?

ptnapoleon:
I get this error:
```
NoHostAvailable: ('Unable to connect to any servers', {'127.0.0.1': AuthenticationFailed('Failed to authenticate to 127.0.0.1: Error from server: code=0100 [Bad credentials] message="Provided username cassandra and/or password are incorrect"',)})
```.

I can check the logs to see if there was an NPE.

**Comment by ptnapoleon** [comment]
Nope, no NPE in the logs after patch 2, just a lot of complaints about schema disagreement.

**Comment by ifesdjeen** [comment]
It should have been as follows: 

```
NoHostAvailable: ('Unable to connect to any servers', {'127.0.0.1': AuthenticationFailed('Failed to authenticate to 127.0.0.1: Error from server: code=0000 [Server error] message="java.lang.NullPointerException"',)})
```

So this is not exactly what I would expect...

PR #1384, opened 2016-11-14
Fix regex for large compaction warning test

ptnapoleon:
Fixes c-12119

Anyone to review


ptnapoleon:
Travis is failing, but that is unrelated to this PR. A new version of flake8 came out, which is picking up different errors. That shouldn't (and won't) be fixed in this PR.

PR #1385, opened 2016-11-15
add new supercolumn upgrade tests

smccarthy788:
@ptnapoleon, for your review.

I replaced the old super column tests seeing as the driver does not work with 1.2 anymore.

CI here: https://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/363


ptnapoleon:
+1

PR #1388, opened 2016-11-16
Update bootstrap test following pcmanus/ccm#557

beobal:
The changes to subprocess' stdout/stderr when starting a ccm node seem to have broken simultaneous_bootstrap_test. The changes I've made fix the test, but there may be better ways to acheive that.

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -647,7 +647,7 @@ class TestBootstrap(BaseBootstrapTest):
         time.sleep(.5)
         node2.watch_log_for("Starting listening for CQL clients")
 
-        node3.grep_log(bootstrap_error)
+        node3.watch_log_for(bootstrap_error)
 
         session = self.patient_exclusive_cql_connection(node2)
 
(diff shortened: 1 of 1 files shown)


bootstrap_test.py
```diff
@@ -641,14 +639,16 @@ def simultaneous_bootstrap_test(self):
...
-        self.assertFalse(node3.is_running(), msg="Two nodes bootstrapped simultaneously")
+        try:
+            node3.start(wait_other_notice=False, verbose=False)
+        except NodeError:
+            pass  # node doesn't start as expected
 
+        time.sleep(.5)
         node2.watch_log_for("Starting listening for CQL clients")
 
+        node3.grep_log(bootstrap_error)
```
pauloricardomg: this should probably be `watch_log_for`, otherwise it will not throw error if logs is not present.

ptnapoleon: Agreed

ptnapoleon: Or, you can assert that `node3.grep_log(bootstrap_error)` returns true, but watching is probably preferable.

beobal:
Thanks, committed with the nit fixed, this fixes CASSANDRA-12927.

PR #1386, opened 2016-11-15
Mark TestBootstrap as a test class

ptnapoleon:
@mambocab to review, this is a hotfix. We can think of a better way to do this later.

```
nose -sv --collect-only bootstrap_test.py
bootstrap_on_write_survey_test (bootstrap_test.TestBootstrap) ... ok
bootstrap_with_reset_bootstrap_state_test (bootstrap_test.TestBootstrap) ... ok
consistent_range_movement_false_with_replica_down_should_succeed_test (bootstrap_test.TestBootstrap) ... ok
consistent_range_movement_false_with_rf1_should_succeed_test (bootstrap_test.TestBootstrap) ... ok
consistent_range_movement_true_with_replica_down_should_fail_test (bootstrap_test.TestBootstrap) ... ok
consistent_range_movement_true_with_rf1_should_fail_test (bootstrap_test.TestBootstrap) ... ok
decommissioned_wiped_node_can_gossip_to_single_seed_test (bootstrap_test.TestBootstrap) ... ok
decommissioned_wiped_node_can_join_test (bootstrap_test.TestBootstrap) ... ok
failed_bootstrap_wiped_node_can_join_test (bootstrap_test.TestBootstrap) ... ok
killed_wiped_node_cannot_join_test (bootstrap_test.TestBootstrap) ... ok
local_quorum_bootstrap_test (bootstrap_test.TestBootstrap) ... ok
manual_bootstrap_test (bootstrap_test.TestBootstrap) ... ok
read_from_bootstrapped_node_test (bootstrap_test.TestBootstrap) ... ok
resumable_bootstrap_test (bootstrap_test.TestBootstrap) ... ok
shutdown_wiped_node_cannot_join_test (bootstrap_test.TestBootstrap) ... ok
simple_bootstrap_test (bootstrap_test.TestBootstrap) ... ok
simple_bootstrap_test_nodata (bootstrap_test.TestBootstrap) ... ok
simple_bootstrap_test_small_keepalive_period (bootstrap_test.TestBootstrap) ... ok
simultaneous_bootstrap_test (bootstrap_test.TestBootstrap) ... ok
test_cleanup (bootstrap_test.TestBootstrap) ... ok

----------------------------------------------------------------------
Ran 20 tests in 0.001s

OK
```

PR #1391, opened 2016-11-16
add plugin with selector to check __test__

mambocab:
This addresses the problem from #1386 -- if you run the tests with this plugin installed, it'll make sure every test class where `bool(cls.__test__) == False` has `__test__` set on `cls` itself, rather than inheriting it from a superclass.

I've set up a test on CassCI:

https://cassci.datastax.com/view/Dev/view/mambocab/job/mambocab-trunk-copy-dtest/10/

This test uses this branch, but with a commit removing `__test__` from some subclasses.

This should fail -- I'm not sure whether I expect it to fail in bucketization or during the test run.

Anyone to review code-wise, @ptnapoleon to review behavior. Is this what we want?

### DIFF
--- a/plugins/test_attrib_checker/test_attrib_checker.py
+++ b/plugins/test_attrib_checker/test_attrib_checker.py
@@ -1,0 +1,41 @@
+from nose import plugins, selector
+
+
+class _ClassCheckingSelector(selector.Selector):
+
+    def wantClass(self, cls):
+        default_result = super(_ClassCheckingSelector, self).wantClass(cls)
+
+        if not default_result:
+            falsy_test = hasattr(cls, '__test__') and not cls.__test__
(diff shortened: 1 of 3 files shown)


ptnapoleon:
Your CI failed, and not in a good way

plugins/test_attrib_checker/test_attrib_checker.py
```diff
@@ -0,0 +1,41 @@
...
+    Checks that any test classes skipped with `__test__ = False` are done so on
+    the class itself.
+    """
+
+    enabled = True  # always use this plugin if it's installed
+    name = 'test_attrib_checker'
+
+    def configure(self, options, conf):
+        """
+        Noop conifigure method -- this plugin is always on.
```
ptnapoleon: typo

PR #1392, opened 2016-11-17
Test decommission failure when N <= replication factor

kgreav:
pending https://issues.apache.org/jira/browse/CASSANDRA-12510

### DIFF
--- a/topology_test.py
+++ b/topology_test.py
@@ -345,6 +345,32 @@ class TestTopology(Tester):
         out = self.show_status(node2)
         self.assertFalse(null_status_pattern.search(out))
 
+    @since('3.10')
+    def stop_decommission_too_few_replicas_multi_dc_test(self):
+        """
+        Decommission should fail when it would result in the number of live replicas being less than
+        the replication factor. --force should bypass this requirement.
+        @jira_ticket CASSANDRA-12510
+        @expected_errors ToolError when # nodes will drop below configured replicas in NTS/SimpleStrategy
+        """
+        cluster = self.cluster
+        cluster.populate([2, 2]).start(wait_for_binary_proto=True)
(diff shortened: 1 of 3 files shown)


topology_test.py
```diff
@@ -345,6 +345,32 @@ def crash_during_decommission_test(self):
         out = self.show_status(node2)
         self.assertFalse(null_status_pattern.search(out))
 
+    @since('3.10')
+    def stop_decommission_too_few_replicas_multi_dc_test(self):
```
ptnapoleon: Can you also tag this test with `@attr("resource-intensive")` as done in several other test files? You'll need to import it with `from nose.plugins.attrib import attr`

topology_test.py
```diff
@@ -345,6 +345,32 @@ def crash_during_decommission_test(self):
...
+            node4.nodetool('decommission')
+
+        session.execute('DROP KEYSPACE ks')
+        create_ks(session, 'ks2', 4)
+        with self.assertRaises(ToolError):
+            node4.nodetool('decommission')
+
+        node4.nodetool('decommission --force')
+        decommissioned = node4.watch_log_for("DECOMMISSIONED", timeout=120)
+        self.assertTrue(decommissioned, "Node failed to decommission when passed --force")
```
ptnapoleon: I'm confused by these last two lines. What are you trying to do here?

kgreav: The point was to also test that --force actually decommissioned the node. Couldn't find a better way to do this as ccmlib obviously doesn't support force yet.

ptnapoleon: So sorry, @kgreav, I was reading the failure message as relating to the success condition, this is completely right.

ptnapoleon:
+1 once 12510 is committed

PR #1393, opened 2016-11-17
New test for CASSANDRA-12909

stef1927:
A new test for CASSANDRA-12909.


ptnapoleon:
+1

stef1927:
I just pushed a new commit that extends the tests a bit. I've added more CQL types to the partition key, they are all text related as far as the copy conversion is concerned. I've also added a new test that uses a non-counter table, where we only test non-prepared statements without counters, given that the execution paths are separate.

ptnapoleon:
Still looks great, +1

PR #1397, opened 2016-11-29
Make permissions_caching_test more resilient against delays

beobal:
Occasionally, `auth_test:TestAuth.permissions_caching_test` can fail unexpectedly. Investigation suggests that this may be due to the cache expiring before the test has a chance to run the query that verifies the cached permissions. This patch adds a retry in the case that the cache is found to have expired earlier than expected.


ptnapoleon:
+1

PR #1398, opened 2016-11-29
Remove potential reconnection error from auth_test

beobal:
auth_test:TestAuth.system_auth_ks_is_alterable_test can be flaky due to
the client session attempting to reconnect while the cluster is being
restarted. Because it uses the 'cassandra' user, reads to the auth
tables are made a QUORUM & so may fail if nodes are marked down.


ptnapoleon:
+1

PR #1402, opened 2016-12-06
Add max version for tests which shouldn't run for 4.0+

beobal:
The annotation in `upgrade_internal_auth_test` fixes CASSANDRA-12986, as upgrading from 2.1 -> 4.0 is not supported.
 `offline_tools_test` has also been failing recently for similar reasons, but I couldn't find a JIRA for that.


beobal:
I'm not sure exactly how best to manage the change to valid upgrade paths for `upgrade_supercolumns_test`.

ptnapoleon:
+1 on what you've got, @smccarthy788 can chime in on how to fix upgrade sc

PR #1404, opened 2016-12-08
Test upgrades to 4.0 versions, properly

ptnapoleon:
@mambocab to review

Testing here:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-upgrade/4/console

### DIFF
--- a/upgrade_tests/upgrade_manifest.py
+++ b/upgrade_tests/upgrade_manifest.py
@@ -23,11 +23,13 @@ def _get_version_family():
         version_family = '2.2.x'
     elif current_version.vstring.startswith('3.0'):
         version_family = '3.0.x'
-    elif current_version > '3.0':
+    elif '3.1' <= current_version < '4.0':
         version_family = '3.x'
-    elif current_version >= '4.0':
+    elif '4.0' <= current_version < '4.1':
+        version_family = 'trunk'
+    else:
         # when this occurs, it's time to update this manifest a bit!
-        raise RuntimeError("4.0 not yet supported on upgrade tests!")
+        raise RuntimeError("4.1+ not yet supported on upgrade tests!")
 
     return version_family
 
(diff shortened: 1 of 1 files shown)


upgrade_tests/upgrade_manifest.py
```diff
@@ -23,11 +23,10 @@ def _get_version_family():
         version_family = '2.2.x'
     elif current_version.vstring.startswith('3.0'):
         version_family = '3.0.x'
-    elif current_version > '3.0':
+    elif current_version > '3.0' and current_version < '4.0':
         version_family = '3.x'
     elif current_version >= '4.0':
-        # when this occurs, it's time to update this manifest a bit!
-        raise RuntimeError("4.0 not yet supported on upgrade tests!")
```
mambocab: I'm still pro-blowing-up-on-unanticipated-versions, I think. (I could also be talked out of it.) Do we want to restrict the new condition to being less than `'5'`, and raise in an `else` block?

upgrade_tests/upgrade_manifest.py
```diff
@@ -23,11 +23,10 @@ def _get_version_family():
         version_family = '2.2.x'
     elif current_version.vstring.startswith('3.0'):
```
mambocab: Do we want to make this `'3.0' <= current_version < '3.1'` to be more consistent?

upgrade_tests/upgrade_manifest.py
```diff
@@ -23,11 +23,10 @@ def _get_version_family():
...
         version_family = '3.0.x'
-    elif current_version > '3.0':
+    elif current_version > '3.0' and current_version < '4.0':
```
mambocab: This can be `'3.0' < current_version < '4.0'` -- and I think maybe `'3.1'` is a better lower bound? The previous `current_version.vstring.startswith('3.0')` should make it not matter, but `'3.1'` should be better for reading.

upgrade_tests/upgrade_manifest.py
ptnapoleon: Sounds reasonable, though we don't know if 4.0 will take a 4.0.x or 4.x form, or how upgrades will be supported. Blowing up on 5 is reasonable

upgrade_tests/upgrade_manifest.py
ptnapoleon: I will do `'3.1' <= current_version < '4.0'`

upgrade_tests/upgrade_manifest.py
ptnapoleon: I mean, we are either consistent with what's above it, or what's below it, so I don't know?

ptnapoleon:
I'll make your changes, re-run CI, and then merge if it's good?

upgrade_tests/upgrade_manifest.py
mambocab: Ah, sure -- I guess I mean change all the conditions to be `<version_string> <= current_version < <version_string>`.

ptnapoleon: I'd rather not do all that extra CI? I'll just fix trunk, and then clean this up after

mambocab: +1

ptnapoleon:
There are still some upgrade tests failing, but they either appear to be 12986, or a new bug.

PR #1405, opened 2016-12-08
Don't run storage upgrade tests on 4.0+

ptnapoleon:
These are for testing 2.1 -> post-8099, and thus there's no need
to run on 4.0. We don't support upgrades to 4.x from 2.1

@beobal or @pauloricardomg  to review

### DIFF
--- a/upgrade_tests/storage_engine_upgrade_test.py
+++ b/upgrade_tests/storage_engine_upgrade_test.py
@@ -435,14 +435,14 @@ class TestBootstrapAfterUpgrade(TestStorageEngineUpgrade):
         super(TestBootstrapAfterUpgrade, self).setUp(bootstrap=True, jvm_args=LEGACY_SSTABLES_JVM_ARGS)
 
 
-@since('3.0')
+@since('3.0', max_version='4')
 class TestLoadKaSStables(BaseSStableLoaderTest):
     __test__ = True
     upgrade_from = '2.1.6'
     jvm_args = LEGACY_SSTABLES_JVM_ARGS
 
 
-@since('3.0')
+@since('3.0', max_version='4')
 class TestLoadKaCompactSStables(BaseSStableLoaderTest):
     __test__ = True
     upgrade_from = '2.1.6'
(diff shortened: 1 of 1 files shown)


ptnapoleon:
See CASSANDRA-12986

pcmanus:
Fyi, in my [dtest patch](https://github.com/pcmanus/cassandra-dtest/commits/thrift_removal_tests) for CASSANDRA-11115, I've used `max_version='4'`, so could be worth being consistent (and my patch's bigger, so it's less work for you to change, unless you feel you're version is much better for some reason :)).

Other than that though, it's not entirely true that those patch are not useful for 4.0 anymore, they just shouldn't be ran from 2.1 but from 3.0. I mean, those tests basically check that most queries work in mixed versions clusters, which was particularly sensible for the 2.x->3.0 transition since there were a lot of logic involved and thus lots of risks, but is nonetheless something we should theoretically test for every possible combination of allowed mixed-version cluster (which is impractical), and at least for any mixed-version cluster with a major internode protocol change (so any 2 allowed major version to make it simple).

ptnapoleon:
'4' is better, I'm just surprised it works, since '4.0' doesn't. Sounds like these should just be integrated into the regular upgrade tests then? Any idea why we hard code 2.1.9 as the starting version for the upgrade?

ptnapoleon:
Well, integrating with the upgrade tests would test every combination of allowed mixed-version cluster. I'm reluctant to special case these though, but just testing "2.1 -> 3" and "3 -> 4" should be fine.

ptnapoleon:
@pcmanus Do you know why we hard code 2.1.9 as the initial version? If I add support for 3 -> 4 major version upgrading, which two points would you like that to come from, 3.0.latest to 4.0.latest?

pcmanus:
> Do you know why we hard code 2.1.9 as the initial version?

No clue. I assume it was easier to hard-code something initially and that never was updated/improved.

> If I add support for 3 -> 4 major version upgrading, which two points would you like that to come from, 3.0.latest to 4.0.latest?

Yes, I'd say latest to latest is the most logical to pick if we pick something.

ptnapoleon:
@guyboltonking and @smccarthy788, could you review?
Here is some CI:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-upgrade/9/testReport/

smccarthy788:
Code & CI look good. +1

PR #1407, opened 2016-12-13
Thrift removal tests

pcmanus:
This is the changes required on 4.0/trunk now that CASSANDRA-11115 has been committed and basically consists at skipping anything thrift related on 4.0. This depends on a https://github.com/pcmanus/ccm/pull/566, but once that's been committed, we should remove/revert the commit named "Point to modified ccm repo".

### DIFF
--- a/bootstrap_test.py
+++ b/bootstrap_test.py
@@ -24,7 +24,6 @@ class BaseBootstrapTest(Tester):
     __test__ = False
 
     allow_log_errors = True
-    cluster_options = ImmutableMapping({'start_rpc': 'true'})
     ignore_log_patterns = (
         # This one occurs when trying to send the migration to a
         # node that hasn't started yet, and when it does, it gets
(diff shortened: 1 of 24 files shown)


bootstrap_test.py
```diff
@@ -16,15 +16,14 @@
...
-from tools.misc import ImmutableMapping, new_node
+from tools.misc import new_node
 from tools.misc import generate_ssl_stores
 
 
 class BaseBootstrapTest(Tester):
     __test__ = False
 
     allow_log_errors = True
-    cluster_options = ImmutableMapping({'start_rpc': 'true'})
```
ptnapoleon: Shouldn't we still want this when running pre-4.0?

ptnapoleon:
+1

thrift_tests.py
```diff
@@ -64,12 +70,23 @@ def setUp(self):
...
+            return
+
         client.transport.close()
         ReusableClusterTester.tearDown(self)
 
     @classmethod
     def post_initialize_cluster(cls):
         cluster = cls.cluster
+
+        # This is called before the @since annotation has had time to take
```
pcmanus: I'm sure there is, wasn't trying for subtlety. You can also let it rot until we don't test against any version supporting thrift and we remove this. After thrift, since has been frozen for a while so it's not like devs should have to look at this file much nowadays.

PR #1409, opened 2016-12-15
Use patient connection to avoid flakiness

stef1927:
I saw a test fail to connect at this line, although I can no longer reproduce it I think it's safer to use `patient_cql_connection`.


ptnapoleon:
+1

PR #1413, opened 2016-12-20
Fix slow and aborted query tests for CASSANDRA-13050

stef1927:
[CASSANDRA-13050](https://issues.apache.org/jira/browse/CASSANDRA-13050) fixed a bug when monitoring queries, as a consequence the delay time that is applied for testing needs to be reduced, since it will now be applied for every CQL row. The total request timeout needs to be increased for the slow query tests, to ensure the query is only reported as slow and not aborted. 

I've also fixed a few more problems:

* `monitoring_check_interval_ms` is actually `monitoring_report_interval_ms`, unsure when it was renamed but the tests were clearly not updated, it's value was decreased since monitoring more frequently ensures the logs are updated more quickly
* some request timeouts were missing
* using logs at INFO level causes flakiness because the no spam logger only logs every 5 minutes, hence if the cluster is not restarted we may have missing logs.

I've also renamed `AbortedQueriesTester` to `AbortedQueryTester` for consistency with `SlowQueryTester`.

### DIFF
--- a/cql_tests.py
+++ b/cql_tests.py
@@ -699,7 +699,7 @@ class MiscellaneousCQLTester(CQLTester):
 
 
 @since('3.2')
-class AbortedQueriesTester(CQLTester):
+class AbortedQueryTester(CQLTester):
     """
     @jira_ticket CASSANDRA-7392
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
These changes look good to me, but are you sure monitoring_report_interval_ms is the appropriate key for all C* branches?

If this is working on all branches, +1

stef1927:
Thanks for the review. 

I've checked `monitoring_report_interval_ms` with a git annotate, and it has been like this since CASSANDRA-7392 in 3.2. I think it must have changed during the code review and we simply forgot to update the tests.

PR #1416, opened 2017-01-03
Add test that DES works in multi-datacenter situations (CASSANDRA-13074)

jkni:
This test adds test coverage for DES in multi-datacenter situations.

It does this by populating scores for the local datacenter and ensuring that read metrics are not incremented on the degraded host when doing local quorum reads.

A [multiplexer run](http://cassci.datastax.com/job/parameterized_dtest_multiplexer/374/) shows this passing 90 times when the patches for CASSANDRA-13074 are applied. This should not be merged until CASSANDRA-13074 has been merged.

### DIFF
--- a/snitch_test.py
+++ b/snitch_test.py
@@ -1,8 +1,15 @@
 import os
 import socket
+import time
 
+from nose.plugins.attrib import attr
+
+from cassandra import ConsistencyLevel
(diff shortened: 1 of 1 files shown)


jkni:
The Cassandra ticket has been merged. This is ready to merge after review.

PR #1419, opened 2017-01-05
remove the allocate_tokens_for_keyspace in disk_balance_test

DikangGu:
It does not make sense to have the allocate_tokens_for_keyspace, because the keyspace is not created at cluster start time yet.

And it will break the test after CASSANDRA-13080.


ptnapoleon:
Looks reasonable to me.

PR #1420, opened 2017-01-05
Let since decorator max version apply to upgrade tests as well

ptnapoleon:
@pcmanus or @smccarthy788 to review


ptnapoleon:
This appears to be the behavior that every upgrade test expected anyway.

smccarthy788:
lgtm +1

PR #1421, opened 2017-01-09
Test paging for distinct queries where the rows and the static column…

blerer:
…s are empty
This test is for CASSANDRA-13017


ptnapoleon:
+1

PR #1423, opened 2017-01-16
Use decommission --force on hintedhandoff_decom_test

pauloricardomg:
This test is broken on trunk after [CASSANDRA-12510](https://issues.apache.org/jira/browse/CASSANDRA-12510), so we need to use the `--force` flag to allow decommission to happen (dropping below RF is not relevant since we just want to make sure hints are being delivered on this test).

### DIFF
--- a/hintedhandoff_test.py
+++ b/hintedhandoff_test.py
@@ -166,8 +166,8 @@ class TestHintedHandoff(Tester):
         insert_c1c2(session, n=100, consistency=ConsistencyLevel.ONE)
         node1.decommission()
         node4.start(wait_for_binary_proto=True)
-        node2.decommission()
-        node3.decommission()
+        node2.nodetool('decommission --force')
+        node3.nodetool('decommission --force')
         time.sleep(5)
         for x in xrange(0, 100):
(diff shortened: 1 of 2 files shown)


ptnapoleon:
Decommission takes a `force=True` kwarg now on the latest ccm. We should cut a release of that, and use it here.

ptnapoleon:
CCM 2.4.10 should have what you need

ptnapoleon:
I'm going to need to run CI on this before merging, as we updated the ccm dep pretty far

ptnapoleon:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-2.2-dtest/1/console
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-trunk-dtest/1/console

ptnapoleon:
Bug in ccm 2.5.0 forced me to update that, re-running CI now:
https://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-2.2-dtest/2/console
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-trunk-dtest/2/console

PR #1425, opened 2017-01-18
CASSANDRA-13112 Populate rows in test_multidatacenter_local_quorum so that read_iteration_delay property takes effect

jkni:
This test was added for CASSANDRA-13074 and multiplexed to make sure it would pass.

Unfortunately, this multiplex run took place before CASSANDRA-13050 was committed, which changed the behavior of the read_iteration_delay property such that it didn't take effect for reads that returned empty results.

We need to populate data to be returned by the selects. A multiplex run [here](http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/380/) shows the test passing on trunk 100 times after this PR. I also tested it on 3.X and 3.11 locally.


ptnapoleon:
+1

PR #1428, opened 2017-01-25
Profiles

ptnapoleon:
@guyboltonking  and @stef1927 to review, please.

http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-profile-dtest/2/console

http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-upgrade/1/console

### DIFF
--- a/dtest.py
+++ b/dtest.py
@@ -468,10 +468,6 @@ class Tester(TestCase):
                             port=port, ssl_options=ssl_opts, connect_timeout=10, allow_beta_protocol_version=True)
         session = cluster.connect(wait_for_all_pools=True)
 
-        # temporarily increase client-side timeout to 1m to determine
-        # if the cluster is simply responding slowly to requests
-        session.default_timeout = 60.0
-
         if keyspace is not None:
             session.set_keyspace(keyspace)
 
--- a/dtest.py
+++ b/dtest.py
@@ -29,6 +29,7 @@ from cassandra import ConsistencyLevel
 from cassandra.auth import PlainTextAuthProvider
 from cassandra.cluster import Cluster as PyCluster
 from cassandra.cluster import NoHostAvailable
+from cassandra.cluster import ExecutionProfile, EXEC_PROFILE_DEFAULT
 from cassandra.policies import RetryPolicy, WhiteListRoundRobinPolicy
 from ccmlib.cluster import Cluster
 from ccmlib.cluster_factory import ClusterFactory
@@ -270,6 +271,12 @@ class Runner(threading.Thread):
             raise self.__error
 
 
+def make_execution_profile(retry_policy=FlakyRetryPolicy(), consistency_level=ConsistencyLevel.ONE, **kwargs):
+    return ExecutionProfile(retry_policy=retry_policy,
+                            consistency_level=consistency_level,
+                            **kwargs)
+
+
 class Tester(TestCase):
 
     maxDiff = None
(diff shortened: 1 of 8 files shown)


ptnapoleon:
~80 upgrade failures that I caused. I'll look into those. The regular dtest run went fine.

ptnapoleon:
Retrying here. I missed setting row_factory in do_upgrade:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-upgrade/2/console

dtest.py
```diff
@@ -434,23 +435,27 @@ def copy_logs(self, cluster, directory=None, name=None):
             if not is_win():
                 os.symlink(basedir, name)
 
+    def make_execution_profile(self, **kwargs):
+        return ExecutionProfile(retry_policy=FlakyRetryPolicy(),
+                                **kwargs)
+
```
stef1927: Where is this method called?

dtest.py
```diff
@@ -434,23 +435,27 @@ def copy_logs(self, cluster, directory=None, name=None):
...
         wlrr = WhiteListRoundRobinPolicy([node_ip])
 
         return self._create_session(node, keyspace, user, password, compression,
-                                    protocol_version, wlrr, port=port, ssl_opts=ssl_opts)
+                                    protocol_version, port=port, ssl_opts=ssl_opts, load_balancing_policy=wlrr, **kwargs)
 
-    def _create_session(self, node, keyspace, user, password, compression, protocol_version, load_balancing_policy=None,
-                        port=None, ssl_opts=None):
+    def _create_session(self, node, keyspace, user, password, compression, protocol_version,
+                        port=None, ssl_opts=None, consistency_level=ConsistencyLevel.ONE, execution_profile=None, **kwargs):
```
stef1927: Do we need to pass in `kwargs`, which as far as I have understood are only used to create an execution profile, given that users can just pass in a different execution profile?

dtest.py
ptnapoleon: Nowhere, but we do want some sort of factory method like this for making an execution profile with dtest's defaults, that the caller can then customize before passing into a connection method.  Since we'll eventually need this, I just added it now. (Though I do see I need to set the default CL here)

dtest.py
ptnapoleon: See the discussion starting here on Slack:
https://datastax.slack.com/archives/cassandra-test-inf/p1485356720000464
For why I ended up doing it this way. Basically it boiled down to not wanting to see 
`cql_connection(execution_profile=get_default_execution_profile(one_custom_kwarg))` all over the codebase, when it could just as easily be `cql_connection(one_custom_kwarg)`. 

It may be (probably definitely is) better to let the user pass an entire dict of profiles, and not just one. @guyboltonking ?

dtest.py
guyboltonking: How about using it to construct the default `EXEC_PROFILE_DEFAULT` in `Tester._create_session`?

dtest.py
guyboltonking: > It may be (probably definitely is) better to let the user pass an entire dict of profiles, and not just one

I agree.

dtest.py
ptnapoleon: 👍

PR #1430, opened 2017-01-27
Fix test_tombstone_failure_v3 error handling for 2.1

spodkowinski:
The `read_failures_test.TestReadFailures.test_tombstone_failure_v3` test is now [failing for months](http://cassci.datastax.com/job/cassandra-2.1_dtest/lastCompletedBuild/testReport/read_failures_test/TestReadFailures/test_tombstone_failure_v3/) for 2.1. The reason for this seems to be that the expected `ReadTimeout` is actually wrapped in `OperationTimedOut` due to the retry semantics at work. One option would be either to disable retries by using FallthroughPolicy, or, as in the PR, to unwrap the error.


ptnapoleon:
+1. I went looking for the jira ticket for this failure, and couldn't find one.

PR #1431, opened 2017-01-27
Update upgrade test post-CASSANDRA-12694

pcmanus:
CASSANDRA-12694 did slightly change the behavior in that case (for the
better) and updated unit tests accordingly (see the end of
https://github.com/ifesdjeen/cassandra/commit/3b7a2f8479eaafa00d55466fe37e43318b8dadcf,
which is "exactly" the same case) but this upgrade test was missed and
currently fails
(http://cassci.datastax.com/job/cassandra-3.0_dtest_upgrade/lastBuild/testReport/junit/upgrade_tests.cql_tests/TestCQLNodes2RF1_Upgrade_current_3_0_x_To_indev_3_0_x/static_columns_cas_test/).


ptnapoleon:
I worry this will break earlier versions, but we'll see. +1

PR #1432, opened 2017-01-31
Fix read_failures_test to handle ReadTimeout, retries, and driver-local timeouts correctly

jkni:
A recent change to the read_failures_test for protocol v3 appeared to fix the test on 2.1 but broke it on 2.2, 3.0, 3.11, and trunk. The 2.1 fix rested on a very delicate balance and needed a rework. This also fixes 2.2+.

On 2.1, when hitting tombstone overwhelming exceptions, we have no way to signal this back to the coordinator. This means that on the v3 protocol under 2.1 we won't get the read timeout back until the cluster configured read timeout. Since the cluster configured read timeout and the client configured read timeout were both 10 seconds, whether the error caught by 2.1 was an OperationTimedOut with one wrapped ReadTimeout or a bare OperationTimedOut depended on vnode count, test machine, and the speed of the wind. We can up the client timeout to 15 seconds and switch to the FallthroughRetryPolicy since the client timeout spans retries and the retries were doing unnecessary extra work anyway. This allows us to handle errors in v3 the same across all server releases. I've tested all of read_failures_test on 2.1, 2.2, 3.0, 3.11, and trunk with these changes.

### DIFF
--- a/read_failures_test.py
+++ b/read_failures_test.py
@@ -1,4 +1,4 @@
-from cassandra import ConsistencyLevel, ReadFailure, ReadTimeout, OperationTimedOut
+from cassandra import ConsistencyLevel, ReadFailure, ReadTimeout
 from cassandra.policies import FallthroughRetryPolicy
 from cassandra.query import SimpleStatement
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
+1

PR #1433, opened 2017-02-01
Always clone from github, instead of apache

ptnapoleon:
@smccarthy788 to review.

Testing here:
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-upgrade/
http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-trunk-upgradeing-dtest/


smccarthy788:
+1 if CI is good.

PR #1434, opened 2017-02-02
Convert avg compaction throughput to mbs before compaing to threshold

Gerrrr:
Solves https://issues.apache.org/jira/browse/CASSANDRA-13170

The reason for [the test](https://github.com/riptano/cassandra-dtest/blob/master/compaction_test.py#L280) to fail is that 5mb threshold is compared against the first matched `avgthroughput` without unit conversion. In the [node_debug.log](https://issues.apache.org/jira/secure/attachment/12850004/node1_debug.log) attached in the ticket, the threshold was compared against `6.175KiB/s`

### DIFF
--- a/compaction_test.py
+++ b/compaction_test.py
@@ -269,15 +269,23 @@ class TestCompaction(Tester):
         avgthroughput = m.named['avgthroughput']
         found_units = m.named['units']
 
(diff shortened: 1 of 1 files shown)


ptnapoleon:
Awesome, thanks!

PR #1436, opened 2017-02-06
Adding and updating tests to work with CASSANDRA-9143

bdeggleston:


tools/decorators.py
```diff
@@ -10,15 +10,18 @@
 
 class since(object):
 
-    def __init__(self, cass_version, max_version=None):
+    def __init__(self, cass_version, max_version=None, max_version_exclusive=None):
```
ptnapoleon: -1. Either use '4' instead of '4.0' as is common, or just make `max_version` exclusive and change all callsites. It's silly to have both a max_version and max_version_exclusive kwarg

bdeggleston: ah I hadn't thought to just use '4', fixed

ptnapoleon:
+1

PR #1438, opened 2017-02-08
Proper fix for upgrade test post-CASSANDRA-12694

pcmanus:
Follow-up to https://github.com/riptano/cassandra-dtest/pull/1431: @ptnapoleon was obviously right, the initial change was not working for old version and we need to condition this, which is what this PR does.

I'll note that I've never really been able to run upgrade tests locally so while it's a simple patch, it's basically untested.


ptnapoleon:
Lgtm.

PR #1440, opened 2017-02-13
Set consistency level on connection to both nodes in do_upgrade

ptnapoleon:
Joel, which test should I run 1000x times to confirm this fixes the issue?


jkni:
I multiplexed a test that displayed failures fairly often [here](http://cassci.datastax.com/view/Parameterized/job/parameterized_dtest_multiplexer/384/). It passed 500 times, and the change looks good to me.

+1

PR #1444, opened 2017-02-16
Use ccm 2.6.0

ptnapoleon:
@mshuler to review
Testing CI here: http://cassci.datastax.com/view/Dev/view/ptnapoleon/job/ptnapoleon-test-upgrade-dtest/1/console

(shouldnt be a serious change, is outside of the path of regular dtest usage)


mshuler:
+1

PR #1445, opened 2017-02-21
New test for CASSANDRA-13185

stef1927:
This is a new test for [CASSANDRA-13185](https://issues.apache.org/jira/browse/CASSANDRA-13185).

### DIFF
--- a/cqlsh_tests/cqlsh_copy_tests.py
+++ b/cqlsh_tests/cqlsh_copy_tests.py
@@ -3253,3 +3253,72 @@ class CqlshCopyTest(Tester):
 
         _test(True)
         _test(False)
+
+    @since('3.0')
+    def test_unusual_dates(self):
+        """
+        Test that we can export and import dates that are outside of the Python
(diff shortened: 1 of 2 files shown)


ptnapoleon:
So, this looks good, but I'm a bit paranoid about the fact that our final assertion is comparing the output of two `SELECT *` calls. If either the python driver, or the select statement have a bug in handling these dates, we won't catch it (and it might mask a COPY issue). I'd feel safer if we compared `args` to `result` or `new_result` at some point.

thobbs:
Agreed, it would be good to have a least a couple of basic sanity checks on the returned dates.  I don't think we have to check every value exactly, just a couple of outliers.

stef1927:
Thanks for the review. I've added a check in https://github.com/riptano/cassandra-dtest/pull/1445/commits/019d31f2b2152e35f448b2b1f13a142614d71782. I plan to squash it before merging.

ptnapoleon:
+1

stef1927:
Thanks, squashed, merging now.

PR #1446, opened 2017-02-21
Test fix for CASSANDRA-10520

blambov:


ptnapoleon:
+1

ifesdjeen:
Merging as C* issue was merged by now. https://issues.apache.org/jira/browse/CASSANDRA-10520

jkni:
I appreciate following up on this, but it looks like this PR was no longer necessary and actually causes the test to fail on 4.0+. This is because of the commit on 10520 that changed the option to default to off. Pushed a revert of this commit.

PR #1447, opened 2017-02-22
Add repair test for CASSANDRA-13153

spodkowinski:
Complementary dtest for [CASSANDRA-13153](https://issues.apache.org/jira/browse/CASSANDRA-13153)

### DIFF
--- a/repair_tests/repair_test.py
+++ b/repair_tests/repair_test.py
@@ -1,5 +1,6 @@
 import threading
 import time
+import re
 from collections import namedtuple
 from threading import Thread
 from unittest import skip, skipIf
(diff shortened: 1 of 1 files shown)


repair_tests/repair_test.py
```diff
@@ -262,6 +263,61 @@ def no_anticompaction_after_subrange_repair_test(self):
         for node in cluster.nodelist():
             self.assertFalse(node.grep_log("Starting anticompaction"))
 
+    def _get_repaired_data(cls, node, keyspace):
```
ptnapoleon: Why is this a method on this class? We never use `cls` anywhere. Better to make it a function, put in an appropriate file in `tools`. If you do want to leave it here, please replace `cls` with `self`.

repair_tests/repair_test.py
```diff
@@ -262,6 +263,61 @@ def no_anticompaction_after_subrange_repair_test(self):
...
+        out = node.run_sstablemetadata(keyspace=keyspace).stdout
+
+        def matches(pattern):
+            return filter(None, [pattern.match(l) for l in out.split('\n')])
+
+        names = [m.group(1) for m in matches(_sstable_name)]
+        repaired_times = [int(m.group(1)) for m in matches(_repaired_at)]
+
+        assert names
+        assert repaired_times
```
ptnapoleon: Don't use naked assertions. See `CONTRIBUTING.md`. Either use the unittest assertions, or the assert tools imported from `nose`.

repair_tests/repair_test.py
```diff
@@ -262,6 +263,61 @@ def no_anticompaction_after_subrange_repair_test(self):
...
+        # stop node2, stress and start full repair to find out how synced ranges affect repairedAt values
+        node2.stop(wait_other_notice=True)
+        node1.stress(stress_options=['write', 'n=10K', 'no-warmup', 'cl=ONE', '-rate', 'threads=50'])
+        node2.start(wait_for_binary_proto=True, wait_other_notice=True)
+        node1.nodetool("repair -full -pr keyspace1 standard1")
+
+        meta = self._get_repaired_data(node1, 'keyspace1')
+        repairedAfterFull = set([ m for m in meta if m.repaired > 0 ])
+        # already repaired sstables must remain untouched
+        assert repaired.intersection(repairedAfterFull) == repaired
```
ptnapoleon: See above comment on naked assertions.

repair_tests/repair_test.py
```diff
@@ -262,6 +263,61 @@ def no_anticompaction_after_subrange_repair_test(self):
...
+        node1, node2, node3 = cluster.nodelist()
+        # we use RF to make sure to cover only a set of sub-ranges when doing -full -pr
+        node1.stress(stress_options=['write', 'n=20K', 'no-warmup', 'cl=ONE', '-schema', 'replication(factor=2)', '-rate', 'threads=50'])
+        # disable compaction to make sure that we won't create any new sstables with repairedAt 0
+        node1.nodetool('disableautocompaction keyspace1 standard1')
+        # Do incremental repair of all ranges. All sstables are expected for have repairedAt set afterwards.
+        node1.nodetool("repair keyspace1 standard1")
+        meta = self._get_repaired_data(node1, 'keyspace1')
+        repaired = set([ m for m in meta if m.repaired > 0 ])
+        assert len(repaired) == len(meta)
```
ptnapoleon: See above comment on naked assertions.

repair_tests/repair_test.py
```diff
@@ -262,6 +263,61 @@ def no_anticompaction_after_subrange_repair_test(self):
...
+        assert repaired_times
+        return [_sstable_data(*a) for a in zip(names, repaired_times)]
+
+    @since('2.2.10', '4')
+    def no_anticompaction_of_already_repaired_test(self):
+        """
+        * Launch three node cluster and stress with RF2
+        * Do incremental repair to have all sstables flagged as repaired
+        * Stop node2, stress, start again and run full -pr repair
+        * Verify that non of the already repaired sstables have been anti-compacted again
```
ptnapoleon: Typo here. Should be "none", not "non"

repair_tests/repair_test.py
```diff
@@ -262,6 +263,61 @@ def no_anticompaction_after_subrange_repair_test(self):
...
+        """
+
+        cluster = self.cluster
+        debug("Starting cluster..")
+        cluster.populate(3).start(wait_for_binary_proto=True)
+        node1, node2, node3 = cluster.nodelist()
+        # we use RF to make sure to cover only a set of sub-ranges when doing -full -pr
+        node1.stress(stress_options=['write', 'n=20K', 'no-warmup', 'cl=ONE', '-schema', 'replication(factor=2)', '-rate', 'threads=50'])
+        # disable compaction to make sure that we won't create any new sstables with repairedAt 0
+        node1.nodetool('disableautocompaction keyspace1 standard1')
```
ptnapoleon: I'm not sure this ordering of steps is totally safe. Shouldn't we:
```
stress write 1 row, to init the schema
disableautocompaction
stress write 20k rows
flush
```
?

spodkowinski: For this test, we just have to make sure that the set of sstables won't change due to auto compaction during and after the first repair, but any compactions before shouldn't really matter.

ptnapoleon: I still think you really ought to flush. 20k rows could easily fit in memory, and then this test does nothing

spodkowinski: Repair will always trigger a flush.

spodkowinski:
The ticket has been committed and the PR can now be merged. Thanks!

PR #1450, opened 2017-03-07
Add dtest for CASSANDRA-13053

iamaleksey:
See title.


ptnapoleon:
+1, very readable

jkni:
Test looks good - if the plan goes forward as on CASSANDRA-13053 to fix 2.2+, this should be gated to run on only 2.2+.

PR #1452, opened 2017-03-13
Fix netstats_test

jkni:
The netstats_test in jmx_test.py recently started failing in CI on all branches. This is because Java 1.8u111 changed the details of a ConnectException thrown from native code. I scanned for other tests failing in similar ways and didn't notice any.

I checked the patched test on 2.1+ both pre- and post-1.8u111.


ptnapoleon:
+1

PR #1453, opened 2017-03-20
set max_version since deprecated repair methods are removed in 4.0

yukim:
[CASSANDRA-11530](https://issues.apache.org/jira/browse/CASSANDRA-11530) removes deprecated methods, so once it is committed, we need to patch `deprecated_repair_test` to have max version.


ptnapoleon:
+1

PR #1454, opened 2017-03-27
Add test to reproduce CASSANDRA-13382

pcmanus:
A simple upgrade test using materialized views to reproduce the bug from CASSANDRA-13382


ptnapoleon:
Looks good, but the linter is upset about `10-i-1` on lines 5406 and 5410. Can you change those to `10 - i - 1` before merge? Thanks

pcmanus:
I wouldn't dream to upset the linter, merged with that fixed. Thanks.

PR #1455, opened 2017-03-28
CASSANDRA-9143 change needs dtest change

yukim:
CASSANDRA-9143 introduced strict check for not allowing incremental subrange repair.
dtest needs to pass `-full` to repair command to invoke subrange repair.

The change works for older version of cassandra as well.


ptnapoleon:
+1

PR #1456, opened 2017-03-29
Added test case for CASSANDRA-13364

stef1927:
A new test case that tests [CASSANDRA-13364](https://issues.apache.org/jira/browse/CASSANDRA-13364).


ptnapoleon:
+1

PR #1458, opened 2017-03-30
Changed streaming keep alive test to use byteman (CASSANDRA-12929)

pauloricardomg:


bootstrap_test.py
```diff
@@ -148,23 +148,28 @@ def simple_bootstrap_test_small_keepalive_period(self):
...
+        node1.import_config_files()
+
         cluster.start(wait_other_notice=True)
 
         # Create more than one sstable larger than 1MB
-        node1.stress(['write', 'n=50K', '-rate', 'threads=8', '-schema',
-                      'compaction(strategy=SizeTieredCompactionStrategy, enabled=false)'])
-        cluster.flush()
-        node1.stress(['write', 'n=50K', '-rate', 'threads=8', '-schema',
+        node1.stress(['write', 'n=1K', '-rate', 'threads=8', '-schema',
```
ptnapoleon: This is a big drop in row count. Is it going to be enough?

bootstrap_test.py
```diff
@@ -148,23 +148,28 @@ def simple_bootstrap_test_small_keepalive_period(self):
...
                       'compaction(strategy=SizeTieredCompactionStrategy, enabled=false)'])
         cluster.flush()
-        self.assertGreater(node1.get_sstables("keyspace1", "standard1"), 1)
+
+        debug("Submitting byteman script to {} to".format(node1.name))
+        # Sleep longer than streaming_socket_timeout_in_ms to make sure the node will not be killed
+        node1.byteman_submit(['./byteman/stream_5s_sleep.btm'])
```
ptnapoleon: I don't see this script. Did you mean to include it?

bootstrap_test.py
pauloricardomg: yeah, with the byteman script which will sleep on streaming this is just required to create the keyspace and 1 sstable so a few rows is sufficient.

bootstrap_test.py
pauloricardomg: oops, forgot to include, nice catch.

pauloricardomg: updated the patch with script

ptnapoleon:
+1

PR #1460, opened 2017-04-03
Restrict data dir for sstabledump test

yukim:
sstabledump test assumes only one SSTable to be written, but in JBOD config multiple SSTables.
This patch strictly set to use only 1 data directory to avoid error.


ifesdjeen:
More information on the issue can be found [here](https://datastax.jira.com/browse/APOLLO-416)

ptnapoleon:
+1

PR #1461, opened 2017-04-03
Fix timeouts at TestAuthUpgrade.upgrade_to_22_test and TestAuthUpgrad…

adelapena:
Both `TestAuthUpgrade.upgrade_to_22_test` and `TestAuthUpgrade.upgrade_to_30_test` use to fail due to timeouts.

These tests started to fail with timeouts when the temporary `session.default_timeout = 60.0` in `dtest.py` was removed, see [here](https://github.com/riptano/cassandra-dtest/commit/059bb8cccb5bd8088572fce73053d5500537a5d1). 

To solve the problem restoring the original behaviour, this PR simply specifies larger timeouts for the failing queries. Also, an optional timeout is added to `assertions.py.assert_all`.

### DIFF
--- a/tools/assertions.py
+++ b/tools/assertions.py
@@ -147,7 +147,7 @@ def assert_none(session, query, cl=None):
     assert list_res == [], "Expected nothing from {}, but got {}".format(query, list_res)
 
 
-def assert_all(session, query, expected, cl=None, ignore_order=False):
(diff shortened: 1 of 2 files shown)


ptnapoleon:
+1

ifesdjeen:
+1

PR #1465, opened 2017-04-26
adding test for repair preview introduced in CASSANDRA-13257

bdeggleston:


repair_tests/preview_repair_test.py
```diff
@@ -0,0 +1,80 @@
+import time
+
+from cassandra import ConsistencyLevel
+from cassandra.query import SimpleStatement
+
+from dtest import Tester
+
```
ptnapoleon: You need two blank lines here, for PEP8 compliance. See the linter's output in the Travis build.

repair_tests/preview_repair_test.py
```diff
@@ -0,0 +1,80 @@
...
+class PreviewRepairTest(Tester):
+
+    def assertNoRepairHistory(self, session):
```
ptnapoleon: Please use snake case, not camel.

repair_tests/preview_repair_test.py
```diff
@@ -0,0 +1,80 @@
...
+        rows = session.execute("select * from system_distributed.repair_history")
+        self.assertEqual(rows.current_rows, [])
+        rows = session.execute("select * from system_distributed.parent_repair_history")
+        self.assertEqual(rows.current_rows, [])
+
+    def preview_test(self):
+        """ Test that preview correctly detects out of sync data """
+        cluster = self.cluster
+        cluster.set_configuration_options(values={'hinted_handoff_enabled': False, 'num_tokens': 1, 'commitlog_sync_period_in_ms': 500})
```
ptnapoleon: If this test always runs with num tokens = 1, it should be annotated with `@no_vnodes()` instead of explicitly setting the config like this.

repair_tests/preview_repair_test.py
```diff
@@ -0,0 +1,80 @@
...
+        node1.stop(gently=False)
+        node3.start(wait_other_notice=True, wait_for_binary_proto=True)
+        session = self.exclusive_cql_connection(node2)
+        for i in range(10):
+            session.execute(stmt, (i + 20, i + 20))
+        node1.start(wait_other_notice=True, wait_for_binary_proto=True)
+
+        # data should not be in sync for full and unrepaired previews
+        result = node1.repair(options=['ks', '--preview'])
+        self.assertNotIn("Previewed data was in sync", result.stdout)
```
ptnapoleon: Is there a positive assertion we can make, as well? Like "Previewed data was not in sync"? Otherwise, we risk getting false negatives when this string changes.

PR #1467, opened 2017-05-01
Add a test for CASSANDRA-13346

juiceblender:
I've optionally made the JMX querying accept a `verbose` flag because the nosetest would print out all stacktraces otherwise...and finally end with an 'OK'. 

Any feedback/comments welcome (Or if you'd rather a patch etc.)

### DIFF
--- a/jmx_test.py
+++ b/jmx_test.py
@@ -13,7 +13,6 @@ from tools.misc import generate_ssl_stores
 
 
 class TestJMX(Tester):
-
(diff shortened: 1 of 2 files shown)


ptnapoleon:
+1. Please ping me when that ticket is merged, and i'll merge this

juiceblender:
@ptnapoleon it has been merged in, thanks!

PR #1468, opened 2017-05-01
adding cluster reconfiguration tests for 9143

bdeggleston:
With CASSANDRA-13257 committed, we can easily validate that all repaired data is in sync, so I've added some tests that verify everything looks ok after cluster reconfigurations.


ptnapoleon:
+1

PR #1471, opened 2017-05-12
Add test for CASSANDRA-13275

ifesdjeen:


ptnapoleon:
Code looks good, but the linter is upset. +1 after that's cleaned up

PR #1472, opened 2017-05-15
CASSANDRA-10726: add dtest to verify it works

xiaolong302:


read_repair_test.py
```diff
@@ -197,6 +197,36 @@ def test_gcable_tombstone_resurrection_on_range_slice_query(self):
...
+        node2.byteman_submit(['./byteman/simulate_write_timeout_node2.btm'])
+
+        session.execute(SimpleStatement("INSERT INTO read_repair_retry.t1 (k, a, b) VALUES (1, 1, 1)",
+                                        consistency_level=ConsistencyLevel.LOCAL_QUORUM))
+        debug("do the local quorum read")
+        # read 99 times to make sure it hit node2 to simulate write timeout in read repair case
+        # details see CASSANDRA-10726
+        # before CASSANDRA-10726 fix, it will hit read exception since it failed to read repair node2
+        # after CASSANDRA-10726 fix, all read will succeed
+        for n in range(1,  100):
```
ptnapoleon: This is what the linter is upset about, there are too many spaces after the `,`

ptnapoleon:
You need to fix the style issues that the linter is warning about, and also if you could match the string interpolation style laid out in CONTRIBUTING.md, that would be much appreciated.

The test itself looks good.

PR #1475, opened 2017-05-25
add test to confirm that hostname validation is working

jasobrown:
we have tests to prove that the TLS hostname verification fails or is ignored, but no positive test to prove that it actually works.


ptnapoleon:
+1

PR #1479, opened 2017-06-13
WIP: New dtest for CASSANDRA-11381 – Node running with join_ring=false and authentication can not serve requests

michaelsembwever:


michaelsembwever:
@jkni ping.

jkni:
Thanks - I like the test coverage here. As a minor nit, the flake8/pep8 stuff mentioned by Travis should be fixed. 

When testing login failures, it looks like the test makes sure an exception thrown is the correct exception, but it doesn't seem to check whether an exception is thrown. We should probably assert an exception is raised.

The other minor nit that would be nice is to cut down the test runtime. The patient cql connection is the main contributor here - what we really want is some sort of "patient" connection that terminates early if we see the exception we expect. As a minimum, we could probably make all connections after the first not patient, since "patient" connections primarily avoid trying to connect to a node too early in its startup as a failure source.

In both the second and third cases above, I'd be fine with scheduling these as follow-up work, since our other auth tests have the same issues as far as I know. What do you think?

ptnapoleon:
Reminder, PRs are no longer being merged here, and presumably this review should be copied over to ASF Jira or a PR to the asf dtest repo.

michaelsembwever:
@ptnapoleon i'll move it over. thanks for the heads up.

michaelsembwever:
Redone for the new repo and latest code in https://github.com/apache/cassandra-dtest/pull/19

PR #1480, opened 2017-06-14
Removed REUSE_CLUSTER from codebase

MichaelHamm:


ptnapoleon:
I'm not sure you got everything. Additionally, let's run some CI on this.

ptnapoleon:
Yes, it looks you can entirely remove `canReuseCluster` and `freshCluster` in dtest.py, and all of their uses across the project (I think just in cqlsh_copy_tests). As a follow-up, we should consider changing cqlsh_copy_tests to use ReusableClusterTester instead.

ptnapoleon:
Your changes look good. How did CI go? Why close the PR?

PR #1485, opened 2017-06-20
Add dtests for compatibility flag introduced in CASSANDRA-13004

ifesdjeen:


ptnapoleon:
I went ahead and just pushed some suggested changes. These tests look correct. Assuming the bug doesn't break our existing upgrade tests (and thus those also need updated), +1. Please squash on merge if you are okay with my commit.

ifesdjeen:
Thank you @ptnapoleon!

PR #1488, opened 2017-07-04
Expand 9673 tests to also run on 3.x

ptnapoleon:
@stef1927 to review. These tests do _not_ pass on 3.11. I am actually not sure why? There are errors reading sstables after the upgrade, but this upgrade path is supported and tested. Stefania, hopefully you know why these were restricted to only 3.0.x originally, which may also explain the failures.


ptnapoleon:
Stefania has confirmed offline that these tests are valid to run on 3.x as well. I'm going to try to figure out why they're failing before merging, in case it's a test issue.

PR #1490, opened 2017-07-06
CASSANDRA-13526: nodetool cleanup on KS with no replicas should remove old data, not silently complete

jasonstack:
[CASSANDRA-13526](https://issues.apache.org/jira/browse/CASSANDRA-13526): allow "nodetool cleanup" to remove local data when there is no replica and node has joined the ring.

### DIFF
--- a/nodetool_test.py
+++ b/nodetool_test.py
@@ -1,9 +1,11 @@
 import os
-
+from cassandra import ConsistencyLevel
+from cassandra.query import SimpleStatement
 from ccmlib.node import ToolError
 
-from dtest import Tester, debug
+from dtest import Tester, debug, create_ks
 from tools.decorators import since
+from tools.assertions import assert_none
 
 
 class TestNodetool(Tester):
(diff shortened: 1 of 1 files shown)


nodetool_test.py
```diff
@@ -93,6 +95,72 @@ def test_nodetool_timeout_commands(self):
...
+        # dc1 data remains
+        statement = SimpleStatement("SELECT * FROM ks.cf", consistency_level=ConsistencyLevel.LOCAL_ONE)
+        self.assertEqual(len(list(session.execute(statement))), 100)
+        if with_index:
+            statement = SimpleStatement("SELECT * FROM ks.cf WHERE value = 'value'", consistency_level=ConsistencyLevel.LOCAL_ONE)
+            self.assertEqual(len(list(session.execute(statement))), 100)
+
+        # alter rf back to query dc2, no data, no index
+        session.execute("ALTER KEYSPACE ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'dc1' : 1, 'dc2' : 1};")
+        assert_none(session_dc2, "SELECT * FROM ks.cf")
```
ptnapoleon: Is there a chance that `cross_dc_read_repair` could occur here, causing this assertion to fail? Do we need to set that to zero at the start of this test?

jasonstack: thanks for pointing out, otherwise it may be another flaky test.. I set "dclocal_read_repair_chance = 0 AND read_repair_chance = 0;"

jasonstack:
pending for JIRA

nodetool_test.py
jeffjirsa: It'd be more explicit to change the replication strategy to be dc2 only, and then there's no possible way to get data from dc1.

jasonstack:
Thanks for reviewing. I will close this MR and move to new repo. https://github.com/apache/cassandra-dtest/pull/1

PR #1493, opened 2017-07-12
Handle index order in describe output on 2.1

jkni:
TL;DR: The index return order on 2.1 is different from 2.2+.

For the version of the Python driver bundled with 2.1, when getting describe output, we iterate over the columns and add any indexes found to an ordered dictionary. This means we return indexes in sort order on the name of the indexed column. On 2.2+, we keep track of indexes in a separate dictionary that isn't ordered. This means we get them in iteration order of the dictionary. In Python 2.7, this order is constant across all runs, which is why we don't need to handle both orders. If Python 3.3, the seed used for dicts was randomized, but in 3.6+, dicts maintain insertion order as an implementation detail. If we support Python versions 3.3-3.5 at any point for dtests, we'll need to handle results being returned in an arbitrary order (or ask the driver to use an ordered dict).

I tested the whole file on 2.1 and 2.2, and test_describe on 3.0+.

### DIFF
--- a/cqlsh_tests/cqlsh_tests.py
+++ b/cqlsh_tests/cqlsh_tests.py
@@ -913,6 +913,9 @@ VALUES (4, blobAsInt(0x), '', blobAsBigint(0x), 0x, blobAsBoolean(0x), blobAsDec
             return ret + "\n" + col_idx_def
 
     def get_users_table_output(self):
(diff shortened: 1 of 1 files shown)


jkni:
Thanks - committed there as `d040629b2a71286105346c3cc637f8d1e16cf0a1`.

